PR #1, opened 2018-10-23
kube-fcos-pipeline: add triggers for builds

dustymabe:
- trigger on buildconfig change
- trigger on imagestream change
- trigger on GitHub webhook
- trigger on Generic webhook (for fedmsg triggered actions)

### DIFF
--- a/kubernetes-jenkins-plugin/README.md
+++ b/kubernetes-jenkins-plugin/README.md
@@ -40,6 +40,30 @@ Create the pipeline (buildconfig with pipeline strategy) and start a build:
 
 ```
 $ oc create -f kube-fcos-pipeline.yaml
+```
+
+Update the "secret" token values in the webooks to be unique
+
+``` 
+$ oc set triggers bc/kubernetes-fcos-pipeline --from-github
+$ oc set triggers bc/kubernetes-fcos-pipeline --from-webhook
+```
+
+Grab the URLs of the webhooks from `oc describe` and set up webhook
+in github.
+
+- `oc describe bc/kube-fcos-pipeline` and grab the `Webhook GitHub` URL
+- From the GitHub web console for the configs repository.
+- Select Add Webhook from Settings → Webhooks & Services.
+- Paste the webook URL output into the Payload URL field.
+- Change the Content Type from GitHub’s default `application/x-www-form-urlencoded` to `application/json`.
+- Click Add webhook.
+
+
+
+Start build using the CLI:
+
+```
 $ oc start-build kube-fcos-pipeline
 ```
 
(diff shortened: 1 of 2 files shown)


kubernetes-jenkins-plugin/README.md
```diff
@@ -40,6 +40,32 @@ Create the pipeline (buildconfig with pipeline strategy) and start a build:
...
+```
+
+Grab the URLs of the webhooks from `oc describe` and set up webhook
+in github.
+
+- `oc describe bc/kube-fcos-pipeline`
+- From the GitHub web console.
+- In your GitHub repository, select Add Webhook from Settings → Webhooks
+  & Services.
+- Paste the URL output (similar to above) into the Payload URL field.
```
arithx: I haven't played with this much, what is the `(similar to above)` actually referencing?

dustymabe: ahh, I copy/pasted the instructions from somewhere else.. will clarify

dustymabe: uploaded an attempt at fixing the problem

kubernetes-jenkins-plugin/kube-fcos-pipeline.yaml
```diff
@@ -4,6 +4,26 @@ apiVersion: "build.openshift.io/v1"
...
+  #   - trigger on buildconfig change
+  #   - trigger on imagestream change
+  #   - trigger on GitHub webhook
+  #   - trigger on Generic webhook (for fedmsg triggered actions)
+  triggers:
+  - type: ConfigChange
+  # use `oc set triggers --from-github` to update token to be unique
+  - type: GitHub
+    github:
+      secret: regenerate
```
jlebon: You can also switch this file to be a template and then define this value to be a parameter which defaults to e.g.:

```
- description: Shared webhook secret.
  name: GITHUB_WEBHOOK_SECRET
  generate: expression
  from: '[A-Z0-9]{32}'
  required: true
```

The advantage of this is that folks can set their webhook secret at `oc create` time so that they don't have to keep changing webhook secrets in their fork settings when testing. E.g. `oc process -p GITHUB_WEBHOOK_SECRET=... | oc create -f -`.

dustymabe: yeah I thought about that. For some reason I've tried to shy away from openshift templates in the past but I guess since we're using a buildconfig here (openshift specific) it wouldn't hurt to make it an openshift template.

I'll put that as a future thing to do since I don't think I'm going to get around to it today. Will handle in a future PR.

PR #2, opened 2018-10-25
Some repo restructuring

jlebon:
No changes to the actual pipeline, just restructuring the repo and updating the README. See individual commit messages for details.


Jenkinsfile
```diff
@@ -0,0 +1,93 @@
+pipeline {
```
dustymabe: can we add a `# vim: syntax=groovy` line here since it got renamed from `pipeline.groovy` in one of your commits  to `Jenkinsfile` in another ?

manifests/bc.yaml
```diff
@@ -0,0 +1,35 @@
...
+  source:
+    type: Git
+    git:
+      # XXX: make these into parameters to make it easy for the local dev case
+      uri: 'https://github.com/dustymabe/fedora-coreos-ci'
+      ref: master
+  strategy:
+    jenkinsPipelineStrategy:
+      type: JenkinsPipeline
+      jenkinsfilePath: Jenkinsfile
```
dustymabe: does it matter that `Jenkinsfile` is actually one directory up from `manifests/bc.yaml` ?

dustymabe: ahh I see it's relative to the uri provided in the source section. 

https://docs.openshift.com/container-platform/3.11/dev_guide/builds/build_strategies.html#pipeline-strategy-options

Jenkinsfile
jlebon: Here's something even nicer: https://github.com/martinda/Jenkinsfile-vim-syntax. It does Groovy but also knows Jenkins-specific stuff.

(I use this through [vim-polyglot](https://github.com/sheerun/vim-polyglot) actually, which I highly recommend!).

dustymabe: that looks nice.

manifests/bc.yaml
jlebon: Yup, correct!

I will mention that there is a small downside to splitting them up. That is, you now have to `git push` to test changes to the pipeline. Overall though, I think it's a price worth paying for how much nicer it is to hack on.

dustymabe:
thanks @jlebon

manifests/bc.yaml
dustymabe: i guess you could `oc edit` and remove the git link and then `oc apply` local changes and the `oc edit` again to add the git link once you've pushed up changes if you really wanted to ? haven't tested that, but could work in theory.

jlebon: Not sure if I follow. Basically after this PR now:
- for changes to the `bc.yaml` file itself you can do `oc replace -f bc.yaml`
- for changes to the Jenkinsfile, you can do `git push && oc start-build ...`

dustymabe: right. yeah I was just describing a flow in which you wanted to test something *without* a `git push`.

PR #4, opened 2018-10-27
Jenkinsfile: make duffy key and rsyncing optional

dustymabe:
This helps us better support the local dev use case.


jlebon:
Haven't tested this, but LGTM!

PR #6, opened 2018-10-30
Only run on nodes with oci_kvm_hook=allowed label

### LINKED ISSUE #5: centos ci: machines without /dev/kvm
In CentOS CI sometimes get a container without `/dev/kvm` in it:

```
stat: cannot stat '/dev/kvm': No such file or directory
error: Unable to find /dev/kvm
```

According to @bstinson we need to look for nodes with label `oci_kvm_hook=allowed` for nodes with `oci-kvm-hook` rpm installed.

dustymabe:
Fixes #5

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -27,6 +27,8 @@ pipeline {
              readOnly: true
            securityContext:
              privileged: false
+        nodeSelector:
+          oci_kvm_hook: allowed
         volumes:
         - name: data
           persistentVolumeClaim:


dustymabe:
@jlebon I'm interested in if this affects the `oc cluster up` case

jlebon:
It does, but I think it's fine to add a requirement on `oci-kvm-hook` for local dev too until this becomes a native Kube thing.

dustymabe:
thanks!

PR #8, opened 2018-11-19
Add a troubleshooting section

dustymabe:
Also clean up HACKING.md a bit and reference the troubleshooting
sections from HACKING.md.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,14 +1,20 @@
 The `manifests/` directory contains OpenShift manifests that will set up a Jenkins pipeline
 using the [Kubernetes Jenkins plugin](https://github.com/jenkinsci/kubernetes-plugin).
 
-Create a Jenkins instance with a persistent volume backing store:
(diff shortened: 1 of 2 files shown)


HACKING.md
```diff
@@ -1,14 +1,20 @@
...
+### Create a Jenkins instance with a persistent volume backing store:
 
 ```
 $ oc new-app --template=jenkins-persistent --param=MEMORY_LIMIT=2Gi --param=VOLUME_CAPACITY=2Gi
 ```
 
-Create an image stream for the coreos-assembler container. It's nice
-to use an image stream so that we can pull from the local container
+It is a good idea to
+[update all plugins](TROUBLESHOOTING.md#issue-for-plugins-not-being-up-to-date)
```
jlebon: Yeah, I want to switch over to s2i for the Jenkins master as well so the image will already have all the plugins and configs we want and be up to date.

PR #9, opened 2018-11-20
use container for coreos-assembler quay org

dustymabe:
This is obviously better than using the one from my namespace.


jlebon:
Nice!

PR #12, opened 2018-11-21
minor fixups; allow reading of all files on artifacts.ci.centos.org

dustymabe:
d0ab9a8  sleep container: remove duff-ssh-key volumemount 
9c2eab9  archive: allow read when artifacts are on webserver
9f41301  archive: Add note about potential failure


Jenkinsfile
```diff
@@ -95,6 +95,8 @@ pipeline {
           RSYNC_PASSWORD=\$(cat \$keyfile)
           export RSYNC_PASSWORD=\${RSYNC_PASSWORD:0:13}
           set -x
+          # Change perms to allow reading on webserver side
+          chmod -R o+r ./builds/
```
jlebon: Probably also worth doing a `chmod a+rX`. See e.g. https://github.com/CentOS/sig-atomic-buildscripts/pull/351. So something like:

```
find builds/ ! -type l -exec chmod a+rX {} +
```

dustymabe: the PR you linked to did `-type d`. Did you mean to say `-type l` ?

jlebon: See https://github.com/CentOS/sig-atomic-buildscripts/pull/355.

dustymabe: fixed

PR #13, opened 2018-11-21
Convert pipeline into template

jlebon:
This makes it possible to pass custom git repos and URLs for convenience
when hacking locally.


dustymabe:
I'm digesting this one. I think I have another idea too

dustymabe:
LGTM If I come up with other suggestions I'll post them in a new PR

PR #16, opened 2018-11-21
Use S2I build for Jenkins master

jlebon:
This gives us multiple advantages:

1. We can override the default Jenkins configuration. Right now, this is
   just used to disable the automatic creation of the "sample" job. But
   e.g. we can customize the Kubernetes plugin configuration or add init
   Groovy scripts etc... For an example of this being used, see [1].
2. We can pin plugins to specific versions.
3. We can pin Jenkins itself to specific versions (though we could do
   that without S2I as well). This patch for example bumps Jenkins to
   the latest LTS release.

[1] https://github.com/projectatomic/paci/tree/master/jenkins/bc/master/configuration


jlebon:
> This patch for example bumps Jenkins to the latest LTS release.

Actually, the before last LTS release (2.138.2 vs 2.138.3). I opened a patch upstream to bump it: https://github.com/openshift/jenkins/pull/744.

This then goes through CI to sanity check that it plays nice with Kubernetes & OpenShift, etc...

dustymabe:
> [1] https://github.com/projectatomic/paci/tree/master/jenkins/bc/master/configuration

might want to make this link refer to the tree at a specific commit in case that file in master changes

jlebon:
> might want to make this link refer to the tree at a specific commit in case that file in master changes

Done!

manifests/bc.yaml
```diff
@@ -22,7 +22,54 @@ parameters:
     name: REPO_REF
     value: master
 objects:
-  - kind: "BuildConfig"
+  # use our own "gated" imagestream for the Jenkins master so we can test new
+  # images before upgrading
+  - apiVersion: v1
+    kind: ImageStream
```
dustymabe: I think I'd feel better about this if we had the jenkins buildconfig in it's own separate yaml file. Any good reason to keep it in here other than reducing the comands to run by 1 ?

manifests/bc.yaml
```diff
@@ -22,7 +22,54 @@ parameters:
...
+    metadata:
+      name: jenkins-s2i
+      namespace: fedora-coreos
+    spec:
+      tags:
+        - name: stable
+          from:
+            kind: DockerImage
+            # 2.138.2
+            name: openshift/jenkins-2-centos7@sha256:fd48b67ed2eca6c4042ac59dd3efd6763681bbc312b5f0c2a3adbfbd71f3c5ce
```
dustymabe: - we lock on a specific commit here. we'll need to upgrade it in the future?
- should we specify the domain since it might move to quay.io in the future? I assume it's `docker.io` today?

manifests/bc.yaml
dustymabe: looks like we are reusing the REPO_URL and REPO_REF below. so that might mean we can't break it out in a separate file

dustymabe:
I don't see any configuration for jenkins plugins today. Did we not need to upgrade any plugins? I hit issues when I didn't update them in the past (not sure why). Did you not hit any?

manifests/bc.yaml
jlebon: Yeah exactly. The reason for keeping them together is so we can share template vars, as well as make set up slightly easier.

> I think I'd feel better about this if we had the jenkins buildconfig in it's own separate yaml file.

Anything in particular that you don't like? I'm fond of the "uni-template" approach as a way to have everything in one place. I think it's a pretty standard approach for OpenShift (see e.g. the Jenkins template we instantiate from for example: https://github.com/openshift/jenkins/blob/76b2cae92c0e5ae4837a83cd5402031d648d1412/openshift/templates/jenkins-persistent.json).

For legibility, note I added some dividers in #19.

manifests/bc.yaml
jlebon: > we lock on a specific commit here. we'll need to upgrade it in the future?

Yeah, basically, because the master is so crucial, it feels like something we should update ourselves for now after sanity checking locally. (Or ideally by a bot later on when we have some CI in place).

> should we specify the domain since it might move to quay.io in the future? I assume it's docker.io today?

Yeah, good point. Will do.

jlebon:
> I  don't see any configuration for jenkins plugins today. Did we not need  to upgrade any plugins? I hit issues when I didn't update them in the  past (not sure why). Did you not hit any?

The plugins are pinned to specific versions that were sanity checked by upstream CI: https://github.com/openshift/jenkins/blob/ba1e4d95a104b948d80cf647ae27d693038f8e6f/2/contrib/openshift/base-plugins.txt. See e.g. example PR that bumps them: https://github.com/openshift/jenkins/pull/732.

jlebon:
(Though of course we can update any one of them that give us trouble, ideally upstream directly if possible but otherwise here directly).

manifests/bc.yaml
jlebon: Fixed in latest rebased! :arrow_down:

PR #17, opened 2018-11-21
Jenkinsfile: use dumb-init

jlebon:
Otherwise, I've got a bunch of defunct qemu-kvm processes and the final
`qemu-kvm` from `virt-install` just hangs at
`reboot: Restarting system`. Not sure why prod isn't hitting this.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -17,8 +17,8 @@ pipeline {
          - name: coreos-assembler
            image: docker-registry.default.svc:5000/fedora-coreos/coreos-assembler:master
            imagePullPolicy: Always
-           command: ['/bin/bash']
-           args: ['-c', 'sleep infinity']
+           command: ['/usr/bin/dumb-init']
+           args: ['sleep', 'infinity']
            volumeMounts:
            - name: data
              mountPath: /srv/


jlebon:
For more information about `dumb-init`, see https://github.com/Yelp/dumb-init. Another popular one I've used in the past is: https://github.com/krallin/tini.

jlebon:
> Another popular one I've used in the past is: https://github.com/krallin/tini.

Oh wow, didn't realize this was powering `--init` in the latest docker versions:

> NOTE: If you are using Docker 1.13 or greater, Tini is included in Docker itself. This includes all versions of Docker CE. To enable Tini, just pass the --init flag to docker run.

rfairley:
I'm reading about the Jenkinsfile syntax for the `containers:` section here https://github.com/jenkinsci/kubernetes-plugin#pod-and-container-template-configuration, not sure whether something else happens when specifying `command:` in the Jenkinsfile. Was assuming giving a `command:` in the Jenkinsfile is like giving arguments to `docker run` https://docs.docker.com/engine/reference/builder/#entrypoint. Will be reading a little more to understand the Jenkinsfile better!

jlebon:
Yeah it's kinda confusing, but there's two terminologies at play here. What Docker calls `ENTRYPOINT` and `COMMAND`, Kubernetes calls `command` and `args`. So in fact `command` here overrides the entrypoint baked in the image. See https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#notes.

rfairley:
Ahh, that clears things up. Thanks!

In that case, the changes LGTM.

jlebon:
Thanks for reviewing!

PR #18, opened 2018-11-22
Add simple-httpd helper app for browsing PV

jlebon:
This is useful for the local dev case. See `HACKING.md`.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -75,10 +75,29 @@ in github.
 - Change the Content Type from GitHub’s default `application/x-www-form-urlencoded` to `application/json`.
 - Click Add webhook.
 
-Start build using the CLI:
+### Start build using the CLI
 
 ```
 $ oc start-build kubernetes-fcos-pipeline
 ```
 
 Use the web interface to view logs from builds.
+
+### [OPTIONAL] Set up simple-httpd
+
+When hacking locally, it might be useful to look at the contents of the
+PV to see the builds since one isn't rsync'ing to an artifact server.
+One alternative to creating a "sleeper" pod with the PV mounted is to
+expose a simple httpd server:
+
+```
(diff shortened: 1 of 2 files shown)


manifests/simple-httpd.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: v1
+metadata:
+  name: simple-httpd
+  labels:
+    app: simple-httpd
+kind: Pod
+spec:
+  containers:
+   - name: httpd
+     image: registry.centos.org/centos/centos:7
```
dustymabe: any reason to use centos here and not fedora?

jlebon: Just that it's what the other OpenShift images should be using as their base image, so just trying to share that base layer. (Though OTOH, coreos-assembler also pulls in the f28 base layer, so... I don't mind switching if you'd prefer).

HACKING.md
```diff
@@ -75,10 +75,27 @@ in github.
...
+When hacking locally, it might be useful to look at the contents of the
+PV to see the builds since one isn't rsync'ing to an artifact server.
+One alternative to creating a "sleeper" pod with the PV mounted is to
+expose a simple httpd server:
+
+```
+$ oc create -f manifests/simple-httpd.yaml
+```
+
+You'll then be able to browse the contents of the PV at:
```
dustymabe: might be worth telling them they can run `oc get route simple-httpd` in order to get the URL instead of having to try to come up with it on their own.

manifests/simple-httpd.yaml
dustymabe: since this is the local dev case I guess it doesn't matter too much. either way works

HACKING.md
jlebon: Yeah, makes sense. Fixed!

Issue #3, opened 2018-10-25
Make it easy to build and test locally


jlebon:
We should strive to make the pipeline easy to hack on locally. Current issues that I've hit so far:

1. for some reason, in the `oc cluster up` case, I have to do:

```diff
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,28 +12,30 @@ pipeline {
       spec:
         containers:
          - name: jnlp
-           image: docker-registry.default.svc:5000/fedora-coreos/jenkins-slave-base-centos7:latest
+           image: 172.30.1.1:5000/myproject/jenkins-slave-base-centos7:latest
          - name: coreos-assembler
-           image: docker-registry.default.svc:5000/fedora-coreos/dustymabe-coreos-assembler:latest
+           image: 172.30.1.1:5000/myproject/dustymabe-coreos-assembler:latest
```

to get the images to pull successfully. I've definitely hit this in the past too when hacking on [PACI](https://github.com/projectatomic/paci), though I don't remember right now how I've worked around it.

2. mounting the Duffy key & pushing artifacts: we don't want to do that locally, because we don't want to overwrite the artifacts in the first place. We need something like the dev mode in the RHCOS pipeline.

dustymabe:
>     1. mounting the Duffy key & pushing artifacts

this is handled in #4

PR #19, opened 2018-11-22
also instantiate the slave and coreos-assembler in the template

jlebon:
This makes the set up instructions much easier and allows us finer
grained control over the resources in the future if we'd like. One
notable change here is `lookupPolicy.local: true`, which allows us to
directly reference the imagestream from the pod. This conveniently also
works around the registry DNS issues mentioned in #3.


jlebon:
Rebased!

jlebon:
This one should unlock a lot of the remaining PRs.

HACKING.md
```diff
@@ -31,7 +31,10 @@ $ oc new-app --file=manifests/bc.yaml \
 Right now this template creates two buildconfigs:
```
dustymabe: there are 5 things in the list below. maybe update this comment

HACKING.md
```diff
@@ -47,34 +50,13 @@ It is a good idea to
 and 
 [set the kubernetes plugin to use DNS for service names](TROUBLESHOOTING.md#issue-for-jenkins-dns-names).
```
dustymabe: are the statements above needed any longer?

manifests/bc.yaml
```diff
@@ -73,6 +73,65 @@ objects:
...
+  # keep a local copy of the Jenkins slave image so we're not constantly pulling
+  # it each time from docker.io
+  - apiVersion: v1
+    kind: ImageStream
+    metadata:
+      name: jenkins-slave-base-centos7
+      namespace: fedora-coreos
+    spec:
+      lookupPolicy:
+        # this allows e.g. the pipeline to directly reference the imagestream
```
dustymabe: nice - didn't know about this feature

dustymabe:
> This conveniently also
> works around the registry DNS issues mentioned in #3.

nice. does that mean we could close #3 after this is merged?

HACKING.md
```diff
@@ -31,7 +31,10 @@ $ oc new-app --file=manifests/bc.yaml \
 Right now this template creates two buildconfigs:
 
 1. The Jenkins master imagestream
-2. The Jenkins pipeline build
+1. The Jenkins slave imagestream
+1. The coreos-assembler imagestream
+2. The PVC
+3. The Jenkins pipeline build
```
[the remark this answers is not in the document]
dustymabe: you did this already with 1/2/3/4/5

HACKING.md
jlebon: I deleted the one about updating the plugins but kept the DNS one. I haven't hit it myself, though I don't think switching to s2i would somehow fix that. Eventually, we could likely tweak the s2i to include this as part of the built master image.

jlebon:
> nice. does that mean we could close #3 after this is merged?

#21 will close that one. :)

PR #20, opened 2018-11-22
pipeline: switch to scripted mode

jlebon:
Jenkins declarative vs scripted modes are a bit confusing. In the end, I
always find the declarative mode too limiting and verbose. It is not
uncommon for plugins to only offer some features through the more
powerful scripted mode rather than declarative.

Switching over allows us to make a cleaner Jenkinsfile without the yaml
definition of the containers baked in, thus reducing further the number
of nested languages we have to deal with.

We also avoid having to always nest commands inside `steps {}` and
`container() {}`. This considerably reduces the size of the file.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,112 +1,57 @@
-pipeline {
-  agent {
-    kubernetes {
-      cloud 'openshift'
-      label 'coreos-assembler'
-      defaultContainer 'jnlp'
-      yaml """
-      apiVersion: v1
-      metadata:
-          name: coreos-assembler
-      kind: Pod
-      spec:
-        containers:
-         - name: jnlp
-           image: jenkins-slave-base-centos7:latest
-           args: ['\$(JENKINS_SECRET)', '\$(JENKINS_NAME)']
-         - name: coreos-assembler
-           image: coreos-assembler:master
-           imagePullPolicy: Always
-           command: ['/usr/bin/dumb-init']
-           args: ['sleep', 'infinity']
-           volumeMounts:
-           - name: data
-             mountPath: /srv/
-           - name: duffy-key
-             mountPath: /var/run/secrets/kubernetes.io/duffy-key
-             readOnly: true
-           securityContext:
-             privileged: false
-        nodeSelector:
-          oci_kvm_hook: allowed
-        volumes:
-        - name: data
-          persistentVolumeClaim:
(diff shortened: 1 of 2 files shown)


jlebon:
Requires #19.

dustymabe:
> In the end, I
> always find the declarative mode too limiting and verbose.

As a new user doing this for the first time I found "limiting and verbose" a feature. Trying to get the scripted mode to work was impossible (too much magic) and I couldn't get the right combination until I switched to declarative mode, which I prefer since it includes the entire pipeline in the one jenkins file.

Are there benefits of scripted mode we want to use right now other than minimizing the file size ?

dustymabe:
> Are there benefits of scripted mode we want to use right now other than minimizing the file size ?

after reviewing a few of the other PRs I'll try to answer my own question. Do we not get "function definitions and other "groovy goodness" with the declarative style from above ?

jlebon:
> As a new user doing this for the first time I found "limiting and verbose" a feature. Trying to get the scripted mode to work was impossible (too much magic) and I couldn't get the right combination until I switched to declarative mode, which I prefer since it includes the entire pipeline in the one jenkins file.

Heh, that's funny. I actually find the declarative mode to be too magic. For example, it automatically does what `checkout scm` does for you.

This might be helpful here: https://jenkins.io/doc/book/pipeline/syntax/#compare. Basically, the declarative pipeline is its own DSL. The scripted pipeline is just Groovy. Things that I'm not sure how to do (or if they can be done) in declarative pipeline:

- factoring out a utils Groovy script and then loading it in (done in #24)
- exiting earlier (done in #25)
- decorating the build description (done in #25)

> Do we not get "function definitions"

That at least looks like it was added in later on: https://stackoverflow.com/questions/47628248.

Edit: ahh OK, it looks like there's a `script()` step one can use to execute in scripted mode from declarative mode: https://jenkins.io/doc/pipeline/steps/pipeline-model-definition/#script-run-arbitrary-pipeline-script

Though... I'm worried that will make hacking on it more complex than it needs to be switching between the two. Note also that the RHCOS pipeline is currently more complex than ours and is in scripted mode. Sharing would probably be easier if we're also in scripted mode.

jlebon:
Rebased!

dustymabe:
LGTM. I'll warm up to scripted mode.

PR #21, opened 2018-11-22
HACKING.md: add instructions for local development

jlebon:
Guide users through the `oc cluster up` workflow and getting the
pipeline set up.

Closes: #3

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,10 +1,94 @@
-The `manifests/` directory contains OpenShift manifests that will set up a Jenkins pipeline
-using the [Kubernetes Jenkins plugin](https://github.com/jenkinsci/kubernetes-plugin).
+The `manifests/pipeline.yaml` file contains an OpenShift template that
+will set up a Jenkins pipeline using the
+[Kubernetes Jenkins plugin](https://github.com/jenkinsci/kubernetes-plugin).
 
-### Create a Jenkins instance with a persistent volume backing store:
+These instructions will cover both local developer workflows, as well as
+interacting with the production pipeline in CentOS CI. A section header
+may indicate that the section only applies to one of the two workflows.
+
+### [LOCAL] Set up an OpenShift cluster
+
+First, make sure to install `oci-kvm-hook` on your host system (not in a
+pet container). This is required to ensure that the pipeline has access
+to `/dev/kvm`:
 
 ```
-$ oc new-app --template=jenkins-persistent --param=NAMESPACE=fedora-coreos --param=MEMORY_LIMIT=2Gi --param=VOLUME_CAPACITY=2Gi
+rpm-ostree install oci-kvm-hook # if on OSTree-based system
+dnf install -y oci-kvm-hook # if on traditional
+```
+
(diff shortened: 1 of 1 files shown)


jlebon:
Requires: #19

HACKING.md
```diff
@@ -1,10 +1,88 @@
...
+may indicate that the section only applies to one of the two workflows.
 
+### [LOCAL] Set up an OpenShift cluster
+
+First, make sure to install `oci-kvm-hook` on your host system (not in a
+pet container). This is required to ensure that the pipeline has access
+to `/dev/kvm`:
+
+```
+rpm-ostree install oci-kvm-hook # if on Silverblue
```
dustymabe: maybe we should say "ostree based system" instead of silverblue. I might run on Atomic Host today or FCOS tomorrow.

HACKING.md
```diff
@@ -1,10 +1,88 @@
...
+dnf install -y oci-kvm-hook # if on traditional
+```
+
+We will use `oc cluster up` to set up a local cluster for testing. To do
+this, simply obtain the OpenShift binary from
+[the latest release](https://github.com/openshift/origin/releases/latest)
```
dustymabe: any reason to not just tell them to install `origin-clients` rpm? and combine with the step above?

We could also mention if they don't want to use an rpm they can grab from github.

HACKING.md
```diff
@@ -1,10 +1,88 @@
...
+
+```
+oc cluster up --version v3.6.1 \
+    --host-config-dir=/srv/oc/config \
+    --host-data-dir=/srv/oc/data \
+    --host-pv-dir=/srv/oc/pv
+```
+
+Then when doing `oc cluster up` in the future, you can use
+`--use-existing-config`.
```
dustymabe: can you provide a full example of the command to run with `--use-existing-config` ? Is it just `oc cluster up --use-existing-config` ?

HACKING.md
```diff
@@ -1,10 +1,88 @@
...
+
+Once the cluster is up, we need to mark our only node (`localhost`) as
+oci-kvm-hook enabled. To do this:
+
+```
+oc login -u system:admin
+oc edit node localhost
```
dustymabe: I tried in like 5 minutes to come up with an `oc patch` equivalent but failed

HACKING.md
jlebon: Because then we'd need to make sure the instructions keep up with the interface for whatever version is in the repos right now. Today, that's 3.11, which looks very different from v3.9 (i.e. different switches). I also couldn't get 3.11 to work when trying to deploy v3.6.1 (I don't remember the exact error, but in general it's not clear if deploying any version of origin with any version of `origin-clients` more recent is definitely supported). If I or someone else gets it working with 3.11, then I'm totally up for recommending that and suggesting `origin-clients-3.11` as an alternative source.

HACKING.md
jlebon: Ahh that's a good idea. Let me play with that.

jlebon:
Rebased and comments addressed!

dustymabe:
nice. with the updated additions it reads really well.

PR #22, opened 2018-11-22
pipeline: add helper shwrap() function

jlebon:
Add a helper to execute shell commands in a consistent environment with
the traditional `set -xeuo pipefail` and always starting in the `/srv`
dir. I initially played with `workingDir: /srv`, but the Kubernetes
plugin tries to be smart and interprets it as a mountpoint to mount into
the container. Which we don't want here since we already have an `/srv`
mountpoint.

I also removed the `| tee` since I don't think it's helping. I think the
issue there is that the `rpm-ostree` running inside supermin thinks it's
got a tty, so we'd have to hack this directly in the init script.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -4,33 +4,36 @@ node {
     pod = readFile(file: "manifests/pod.yaml")
 }
 
+def shwrap(cmds) {
+  sh """
+    set -xeuo pipefail
+    cd /srv
+    ${cmds}
+  """
+}
+
 podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultContainer: 'jnlp') {
     node('coreos-assembler') { container('coreos-assembler') {
         stage('Init') {
-            sh """
-            cd /srv/
+            shwrap("""
             if [ ! -d src/config ]; then
                 coreos-assembler init https://github.com/coreos/fedora-coreos-config
             fi
-            """
+            """)
         }
         stage('Fetch') {
-            sh """
-            cd /srv/src/config
-            git pull
-            cd /srv
-            coreos-assembler fetch | tee
-            """
+            shwrap("""
+            git -C src/config pull
+            coreos-assembler fetch
+            """)
         }
         stage('Build') {
-            sh """
-            cd /srv/
-            coreos-assembler build | tee
-            """
+            shwrap("""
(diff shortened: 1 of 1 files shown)


jlebon:
Requires: #20

dustymabe:
> I also removed the `| tee` since I don't think it's helping. I think the
> issue there is that the `rpm-ostree` running inside supermin thinks it's
> got a tty

I just ran coreos-assembler (after executing `shell` to get into the container) redirecting all output to a file (i.e. no tty):

```
[coreos-assembler]$ /usr/bin/coreos-assembler build &> /tmp/file.txt
```

I see this in the output when I open the file in `vi`:

```
⠁ Importing rpm-md   0% [░░░░░░░░░░░░░░░░░░░░] (0s) ^M
^[[1A^M^[[2K^[[1B^[[1A⠙ Importing rpm-md  25% [██████░░░░░░░░░░░░░░] (3s) ^M
^[[1A^M^[[2K^[[1B^[[1A⠒ Importing rpm-md  50% [███████████░░░░░░░░░] (2s) ^M
^[[1A^M^[[2K^[[1B^[[1A⠂ Importing rpm-md  75% [████████████████░░░░] (1s) ^M
^[[1A^M^[[2K^[[1B^[[1AImporting rpm-md... done^M
```

should I open an upstream issue?

dustymabe:
> I initially played with `workingDir: /srv`, but the Kubernetes
> plugin tries to be smart and interprets it as a mountpoint to mount into
> the container. Which we don't want here since we already have an `/srv`
> mountpoint.

yeah I hit that too when I was first playing around with this :(

jlebon:
> should I open an upstream issue?

Sure, yeah. It's minor, though it shouldn't be too hard to fix.

jlebon:
Rebased!

dustymabe:
> > should I open an upstream issue?
> 
> Sure, yeah. It's minor, though it shouldn't be too hard to fix.

will do here in a minute

dustymabe:
This LGTM

dustymabe:
> > > should I open an upstream issue?
> > 
> > 
> > Sure, yeah. It's minor, though it shouldn't be too hard to fix.
> 
> will do here in a minute

opened upstream: https://github.com/projectatomic/rpm-ostree/issues/1689

PR #23, opened 2018-11-23
sleeper pod: also use dumb-init and add optional: true

jlebon:
Also use `dumb-init` here as good practice and also make use of the
shortened imagestream name so that it works equally well locally and
in prod.

And finally, also make use of `optional: true` for the duffy key so it
can be used locally.


jlebon:
Requires: #19 

(I realized all these PRs are slowly building up a DAG of dependencies that might make it intimidating to review. The base ones are definitely #16 and #19. That should unlock most of the remaining ones).

dustymabe:
changes here LGTM. can you rebase ?

jlebon:
Rebased!

PR #24, opened 2018-11-23
pipeline: always sync in state from prod first

jlebon:
The basic issue this patch is trying to fix is our overdependence on the
PVC; right now, if the PVC gets wiped, the next build will reset the
versioning and delete all the previous builds in prod.

So an easy way to fix this is simply to make sure that the PVC is in the
same state as the prod version. This should be a no-op normally, except
if the PVC got wiped.

Eventually, we should think about higher-level "release" management such
as how many builds to keep and whether to push smoketested images to
e.g. S3 or something (and instead treat artifacts.ci.centos.org/ as a
"buildmaster" pipeline), etc...

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,11 +5,40 @@ node {
 }
 
 def shwrap(cmds) {
-  sh """
-    set -xeuo pipefail
-    cd /srv
-    ${cmds}
-  """
+    sh """
+        set -xeuo pipefail
+        cd /srv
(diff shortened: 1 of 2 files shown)


jlebon:
Requires #22.

dustymabe:
mostly LGTM. will wait for other PRs to merge + rebase before doing final review.

jlebon:
Rebased!

PR #27, opened 2018-11-28
pipeline: adjust for repo move and rename

jlebon:
The only thing that really changes is the default `REPO_URL` parameter.


jlebon:
Good point, rebased and fixed!

rfairley:
:+1: good to merge from me

PR #28, opened 2018-11-28
pipeline: minor logic fixes with fileExists

jlebon:
Three fixes:

- First, `test -f builds/latest` will always fail because either it
  doesn't exist, or it does but it's a symlink, whereas `test -f` checks
  that it's a file.
- Second, `shwrap_rc()` returns the exit code. But in shell semantics, 0
  means successful, so the if-statement actually had an opposite logic.
  Of course taken together those two issues meant two negatives became a
  positive and it *seemed* like it worked when testing locally... So
  evil.
- Third, when checking for the SSH key, we were still using
  `fileExists()` which is broken inside Kubernetes containers.

So let's create a dedicated `path_exists()` which fixes all of this and
lowers the chances of getting this wrong again in the future. Note we
use `test -e` here to more closely match `fileExists()` (which really
doesn't actually care if it's a file or a directory or a link).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,7 +28,7 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
         }
 
         def prevBuildID = null
-        if (utils.shwrap_rc("test -f builds/latest")) {
+        if (utils.path_exists("builds/latest")) {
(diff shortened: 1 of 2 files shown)


sinnykumari:
I don't have merge permission. @dustymabe or @jlebon can you please merge it?

jlebon:
Thanks for reviewing!

PR #31, opened 2018-11-28
[WIP]: Jenkinsfile: add Testing stage

arithx:
Adds a testing stage which runs all kola tests for the latest build.
Currently does not archive any testing artifacts.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -61,6 +61,28 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             currentBuild.description = "⚡ ${newBuildID}"
         }
 
+        stage('Test') {
+            // clone & build kola
+            utils.shwrap("""
+            git clone https://github.com/arithx/mantle
+            cd mantle
+            git checkout fcos_ci
+            ./build kola kolet
+            """)
+            
+            // determine the image & run kola
+            utils.shwrap("""
+            latest_build=\$(readlink builds/latest)
+            qcow=\$(ls builds/"\${latest_build}"/*-"\${latest_build}"-qemu.qcow2)
+            mantle/bin/kola -p unprivileged-qemu --qemu-image "\${qemu}" -b fcos run | tee
+            """)
+
+            // clean-up the mantle folder
+            utils.shwrap("""
+            rm -rf mantle/
+            """)
+        }
+
         stage('Prune') {
             utils.shwrap("""
             coreos-assembler prune --keep=10


arithx:
@dustymabe can you sync up with me on how to test this? I've ran the commands themselves locally inside a `coreos-assembler` container but want to validate it against the actual pipeline.

dustymabe:
> @dustymabe can you sync up with me on how to test this? I've ran the commands themselves locally inside a `coreos-assembler` container but want to validate it against the actual pipeline.

@arithx one way might be to run the environment locally as described in https://github.com/coreos/fedora-coreos-pipeline/blob/a8bc0efdf1c6eba0bd75a03d434c74f473536f52/HACKING.md and see if it works there. We can also set up a container in centos ci and run it and see if it works. we should also look to get you access to that cluster.

Jenkinsfile
```diff
@@ -47,6 +47,14 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             currentBuild.description = "⚡ ${newBuildID}"
         }
 
+        stage('Test') {
+            utils.shwrap("""
+            latest_build=$(readlink builds/latest)
```
jlebon: Need to escape all the `$` in this hunk.

jlebon:
Yeah, HACKING.md should be pretty good now! It does involve some steps to get `oc cluster up` working, but that's ideally a one time thing :)

jlebon:
(Also, thanks for this patch! :+1: )

dustymabe:
marking as WIP

jlebon:
This is now blocked on a workaround for https://github.com/coreos/mantle/issues/956.

arithx:
Updated to use the [fcos_ci](https://github.com/arithx/mantle/tree/fcos_ci) branch of mantle.

Untested.

jlebon:
I guess if one wanted to test this today, they'd use the `qemu_2_electric_boogaloo` branch right?

Jenkinsfile
```diff
@@ -61,6 +61,28 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
...
+            cd mantle
+            git checkout fcos_ci
+            ./build kola kolet
+            """)
+            
+            // determine the image & run kola
+            utils.shwrap("""
+            latest_build=\$(readlink builds/latest)
+            qcow=\$(ls builds/"\${latest_build}"/*-"\${latest_build}"-qemu.qcow2)
+            mantle/bin/kola -p unprivileged-qemu --qemu-image "\${qemu}" -b fcos run | tee
```
cgwalters: IMO something like this should be [coreos-assembler kola](https://github.com/coreos/coreos-assembler/pull/85).

Jenkinsfile
```diff
@@ -61,6 +61,28 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             currentBuild.description = "⚡ ${newBuildID}"
         }
 
+        stage('Test') {
+            // clone & build kola
+            utils.shwrap("""
+            git clone https://github.com/arithx/mantle
```
cgwalters: Yeah...we need to make this more convenient in cosa.

jlebon:
So, https://github.com/coreos/coreos-assembler/pull/85 was merged now. Rebasing on top of that would be a great opportunity for someone who wants to get familiar with the pipeline and hacking on it!

At least as a first step, just running `qemu-unpriv` as a sanity check, even if we don't act on it. And then as a follow-up we can do the "skip remaining artifacts and archive qcow2" bits.

PR #32, opened 2018-11-28
[WIP]: Fix tag for output of jenkins master bc

dustymabe:
The deployment is looking for `jenkins:latest`, not `jenkins:2`.

### DIFF
--- a/manifests/pipeline.yaml
+++ b/manifests/pipeline.yaml
@@ -68,7 +68,7 @@ objects:
       output:
         to:
           kind: ImageStreamTag
-          name: jenkins:2
+          name: jenkins:latest
           namespace: fedora-coreos
       successfulBuildsHistoryLimit: 2
       failedBuildsHistoryLimit: 2


jlebon:
Hmm, `:2` should definitely work. This is used by the Jenkins template, which defaults to looking for `jenkins:2`. I matched the name so we didn't have to override another template var. See: https://github.com/openshift/jenkins/blob/ba1e4d95a104b948d80cf647ae27d693038f8e6f/openshift/templates/jenkins-persistent.json#L319

dustymabe:
looks like a mismatch in templates from what CentOS CI has and upstream.. In CentOS CI:

```
$ oc get template jenkins-persistent -n openshift -o yaml
...
- description: Name of the ImageStreamTag to be used for the Jenkins image.                                                                                                                                       
  displayName: Jenkins ImageStreamTag                                                                                                                                                                             
  name: JENKINS_IMAGE_STREAM_TAG                                                                                                                                                                                  
  value: jenkins:latest 
...
```

jlebon:
Hmm, indeed. That's odd. In that case, I don't mind changing it to `jenkins:latest` to match CentOS CI, but then we also need to update HACKING.md to make sure folks override `JENKINS_IMAGE_STREAM_TAG` locally.

Or maybe... to keep the local dev flow simpler, we *keep* it as `jenkins:2` and update HACKING.md to make sure we override the variable if working with prod?

jlebon:
(Basically, given that the local workflow will be exercised much more often than the prod one, I'd rather we prioritized that one.)

dustymabe:
> to keep the local dev flow simpler, we _keep_ it as `jenkins:2` and update HACKING.md to make sure we override the variable if working with prod?

yeah I think that makes sense.. we may even be able to write something that grabs the "default" from `oc get template jenkins-persistent -n openshift -o json` and sets that in the object using `oc patch`

jlebon:
@dustymabe Want to update this?

dustymabe:
> @dustymabe Want to update this?

yeah. i'll pick it back up. been digging out from the notification backlog in my inbox. will do shortly.

dustymabe:
marking as WIP

jlebon:
This is now part of #53.

Both CentOS CI and a local `oc cluster up` do in fact still use `jenkins:latest`, so there's nothing else to change in fact.

AFAICT, v3.6 templates live here: https://github.com/openshift/openshift-ansible/blob/release-3.6/roles/openshift_examples/files/examples/v3.6/quickstart-templates/jenkins-persistent-template.json, which does have `jenkins:latest`: https://github.com/openshift/openshift-ansible/blob/17241ee23ddc1e299c065ed773e22c1b7299ef7d/roles/openshift_examples/files/examples/v3.6/quickstart-templates/jenkins-persistent-template.json#L302.

I updated the commit message to clarify the confusion here:

> The deployment is looking for `jenkins:latest`, not `jenkins:2`. This
was changed in later releases of OpenShift to `jenkins:2`, but since
we're working with CentOS CI here, we need to match v3.6.

PR #33, opened 2018-12-04
pipeline: run every 30 minutes

jlebon:
Right now, the pipeline is only ever triggered manually. Let's fire up
the engine here and start churning out builds at a regular interval. We
can reconsider turning this off down the line if we have sufficient
"push triggers" to safely disable the timer.

This makes use of the new devel flag that was added in the previous
patch.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,6 +12,11 @@ node {
     }
 }
 
+properties([
+    disableConcurrentBuilds(),
+    pipelineTriggers(devel ? [] : [cron("H/30 * * * *")])
+])
+
 podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultContainer: 'jnlp') {
     node('coreos-assembler') { container('coreos-assembler') {
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,8 +1,15 @@
-def pod, utils
+def pod, utils, devel
 node {
     checkout scm
     pod = readFile(file: "manifests/pod.yaml")
     utils = load("utils.groovy")
+
+    // just autodetect if we're in prod or not
+    devel = (env.JENKINS_URL != 'https://jenkins-fedora-coreos.apps.ci.centos.org/')
+
+    if (devel) {
+        echo "Running in devel mode on ${env.JENKINS_URL}."
+    }
 }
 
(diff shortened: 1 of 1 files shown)


dustymabe:
funny enough.. I had this sitting in my working tree since last thursday:

```patch
diff --git a/Jenkinsfile b/Jenkinsfile
index babea53..0cfd9d9 100644
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,6 +5,13 @@ node {
     utils = load("utils.groovy")
 }
 
+properties(
+    [
+        # run the pipeline every 30 minutes
+        pipelineTriggers([cron('*/30 * * * *')]),
+    ]
+)
+
 podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultContainer: 'jnlp') {
     node('coreos-assembler') { container('coreos-assembler') {
```

but I haven't had a chance to test it (been setting up a new laptop) so haven't pushed it up yet. 
Thanks for doing that.

Jenkinsfile
```diff
@@ -1,10 +1,22 @@
...
+    devel = (env.JENKINS_URL != 'https://jenkins-fedora-coreos.apps.ci.centos.org/')
+
+    if (devel) {
+        echo "Running in devel mode on ${env.JENKINS_URL}."
+    }
 }
 
+properties([
+    disableConcurrentBuilds(),
+    pipelineTriggers(devel ? [] : [cron("H/30 * * * *")])
```
dustymabe: The H/30 makes it run every 30 minutes but at indeterminate times, right? Is that what we want. Sometimes I like knowing I can expect a build to kick off at the 0 or 30.

dustymabe:
one nit comment, otherwise LGTM.

Jenkinsfile
jlebon: Yeah, the idea is that you spread the load from all the other jobs across the hour instead of having a bunch of jobs all running at 0 and 30. OTOH... we own the Jenkins master, and there's only one job... But then again, we don't own the cluster and there might be *many* Jenkins masters, so I feel like that argument is still somewhat relevant.

> Sometimes I like knowing I can expect a build to kick off at the 0 or 30.

Have some tea and read emails, or really... just click the "Build Now" button :)

Issue #14, opened 2018-11-21
compress FCOS artifacts to optimize downloads


dustymabe:
Right now the qcow is 1.7G and the connection to artifacts.ci.centos.org is slow. We probably need to take a look in the image to see if we can trim some fat, but we probably also need to compress this beast so we can download it faster.

jlebon:
I guess one high level question is whether this should fall under coreos-assembler or not. I.e. should coreos-assembler output already be in a "palatable" state for direct consumption over httpd? It'd be totally trivial to `gzip` in coreos-assembler.

dustymabe:
right. I honestly wasn't sure. In the local dev case (`build` and then `run`) it wouldn't make sense. In the build pipeline case it totally makes sense.

jlebon:
So I think a larger issue here is that we need to figure out how to not rsync the whole directory in and out for each new build. Once we fix that, then we can compress artifacts etc... without affecting our ability to restart from the last version even on a fresh PVC. (Well, compressing the artifacts themselves won't affect that, since `coreos-assembler` just needs the `meta.json` from the last build, but more generally, we should fix that so that all we actually need to pull down is the `meta.json` and the last OSTree commit, similar to what RHCOS does.

dustymabe:
> So I think a larger issue here is that we need to figure out how to not rsync the whole directory in and out for each new build.

I wasn't really worried about that since all the transfer is "local" to centos CI infra, right? 

>  Once we fix that, then we can compress artifacts etc... without affecting our ability to restart from the last version even on a fresh PVC. (Well, compressing the artifacts themselves won't affect that, since `coreos-assembler` just needs the `meta.json` from the last build, but more generally, we should fix that so that all we actually need to pull down is the `meta.json` and the last OSTree commit, similar to what RHCOS does).

Agree. we should only need to 'rsync-in' the previous build info + OSTree commit info (the last one so that we can have ostree commit history)

jlebon:
> I wasn't really worried about that since all the transfer is "local" to centos CI infra, right?

Yeah, it's more about the conceptual concern of needing the whole state of the world, just so we can push out a new build. Seems like a pattern we should avoid. Anyway, I think I'm getting ahead of myself. We can definitely punt on that part for the time being.

PR #34, opened 2018-12-06
pipeline: compress artifacts

jlebon:
Make use of the new `compress` command in coreos-assembler to compress
the artifacts before syncing out.

Closes: #14


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/246
I built the above and pushed it to https://quay.io/repository/jlebon/coreos-assembler for testing locally.

PR #35, opened 2018-12-06
pipeline: prune less aggressively

jlebon:
Let's not keep only the last 3 builds here. That default is really to
appeal for the local dev case. I'm working towards adding a
`prune --keep-younger-than` switch to make it time based instead but for
now let's keep say... 15 builds so there's some semblance of permanence.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -48,7 +48,7 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
 
         stage('Build') {
             utils.shwrap("""
-            coreos-assembler build
+            coreos-assembler build --skip-prune
             """)
         }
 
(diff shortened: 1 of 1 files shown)


dustymabe:
so we only have a 50G PV - I think 15 might be cutting us closer to capacity than I'd like. Maybe 10 would be better ?

dustymabe:
> I think 15 might be cutting us closer to capacity than I'd like

well, before the compression anyway.

dustymabe:
just checked and we are currently using `11G` 
```
sh-4.4$ du -sh /srv/                                                                                                                                                                      
11G     /srv/
```

jlebon:
Sure. Updated! :arrow_up: 

Once we fix https://github.com/coreos/fedora-coreos-pipeline/issues/14#issuecomment-444492638 anyway, we can bump it even higher.

dustymabe:
> Once we fix [#14 (comment)](https://github.com/coreos/fedora-coreos-pipeline/issues/14#issuecomment-444492638) anyway, we can bump it even higher.

agree. Once we get there we'll only need a trivial amount of local cache.

PR #37, opened 2018-12-10
pipeline: fix `prune` invocation

jlebon:
It's `--keep`, not `--keep-last-n`. (Though the underlying script that
does the work uses `--keep-last-n`.)


jlebon:
Yeah, good point!

dustymabe:
LGTM. Thanks @jlebon

PR #40, opened 2018-12-13
pipeline: re-create cache.qcow2 after nuking it

jlebon:
When testing #39, I had only tested that the qcow2 gets nuked. The
issue is that the *next* build will error out since it expects the qcow2
to already be created.

Simply switch to always running `coreos-assembler init --force` since
it's idempotent anyway so it'll only recreate the qcow2 if it was nuked
in the previous run.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -71,6 +71,8 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             utils.shwrap("""
             if [ \$(du cache/cache.qcow2 | cut -f1) -gt \$((1024*1024*8)) ]; then
                 rm -vf cache/cache.qcow2
+                qemu-img create -f qcow2 cache/cache.qcow2 10G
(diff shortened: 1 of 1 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/258

dustymabe:
i wonder.. would it not be better to add the init --force in the section where we delete the qcow rather than running it every time.

jlebon:
Yeah, though we do still need it if the PVC is wiped. So this just dedupes things a bit. :)

dustymabe:
assuming everything that we need is idempotent enough (i.e. nothing gets overwritten, just things that doen't exist get created) then I'm +1

jlebon:
OK, I just switched this to directly recreate the qcow2 instead once we nuke it. `init --force` is idempotent *right now*, but it does seem odd to expect `--force` to be idempotent. I guess we could rework `init` so that:

```
$ coreos-assembler init http://...
```

is idempotent and only requires `--force` if needs to change something.

jlebon:
OK split that out as https://github.com/coreos/coreos-assembler/issues/259.

PR #42, opened 2018-12-14
Extend Fedora CoreOS Pipeline for other distributions.

darkmuggle:
pipeline.yaml: extended template to allow for more variables.

pod-w-certs.yaml: 
   - added pod definition to trust an untrusted CA. 
   - added support for defining the resource contstraints


HACKING.md
```diff
@@ -168,6 +168,11 @@ oc new-app --file=manifests/pipeline.yaml \
     --param=REPO_REF=my-feature-branch
 ```
 
+Optionally, the following parameters are supported:
+- `DISTRO`: build a different distribution, e.g `rhcos`
+- `NAMESPACE`: use a different namespace other than `fedora-coreos`
+- `VOLUME_SIZE`: define a different volume size for Jenkins, defaults to 50Gi
```
dustymabe: should we mention `TAGS` here too?

manifests/pipeline.yaml
```diff
@@ -1,53 +1,67 @@
...
     name: REPO_URL
     value: https://github.com/coreos/fedora-coreos-pipeline
   - description: Git branch/tag reference for Jenkinsfile
     name: REPO_REF
     value: master
+  - description: Distribution name
+    name: DISTRO
+    value: fedora
+  - description: Tags
+    name: tags
```
dustymabe: is there a reason to leave this parameter lowercase, vs uppercase like the others?

manifests/pipeline.yaml
```diff
@@ -1,53 +1,67 @@
...
-      Jenkins pipeline for Fedora CoreOS.
+      Jenkins pipeline for CoreOS distributions.
     iconClass: icon-jenkins
     openshift.io/display-name: Fedora CoreOS Pipeline
     openshift.io/documentation-url: https://github.com/coreos/fedora-coreos-pipeline
     openshift.io/support-url: https://github.com/coreos/fedora-coreos-pipeline
-    openshift.io/provider-display-name: Fedora CoreOS
-    tags: fcos,jenkins,fedora
-  name: fedora-coreos
+    openshift.io/provider-display-name: fedora CoreOS
```
dustymabe: should the `fedora` here be `${DISTRO}`?

darkmuggle: That's _what_ I thought too, however, when its uploaded, it becomes "${DISTRO} CoreOS"; the template does not populate it.

manifests/pipeline.yaml
darkmuggle: Nope, fixed and repushed. I also caught that the default value needed to be corrected.

HACKING.md
darkmuggle: Fixed and re-pushed.

dustymabe: i refreshed my browser. I still don't see `TAGS` mentioned here. Do you?

manifests/pipeline.yaml
jlebon: Hmm, in that case let's just just use `CoreOS`?

darkmuggle:
I just pushed a changed that makes `pod-w-certs.yaml` `pod.yaml`.

manifests/pipeline.yaml
```diff
@@ -1,53 +1,67 @@
...
+    tags: ${TAGS}
```
jlebon: All this stuff BTW is just to make the template look pretty in the console template browser. We don't really have to support that though since one would normally set up the pipeline from the CLI, not the console. For `tags` here, I'd say it's not worth the overhead of a separate template parameter. Let's just extend it to e.g. `fcos,rhcos,jenkins,fedora`?

manifests/pipeline.yaml
```diff
@@ -1,53 +1,67 @@
...
     value: master
+  - description: Distribution name
+    name: DISTRO
+    value: fedora
+  - description: Tags to label the pipeline with
+    name: TAGS
+    value: fedora,jenkins,fcos
+  - description: Name of the namespace to use
+    name: NAMESPACE
+    value: fedora-coreos
```
jlebon: Here's what I'd suggest here: drop the `NAMESPACE` parameter completely and drop all the `namespace: fedora-coreos` keys throughout this file. By default, all these objects are created in the current project one is in. So we should be able to abstract the actual namespace name out of this template.

manifests/pipeline.yaml
```diff
@@ -1,53 +1,67 @@
...
+    name: VOLUME_SIZE
+    value: 50Gi
 objects:
 
   ### JENKINS MASTER ###
 
   # use our own "gated" imagestream for the Jenkins master so we can test new
   # images before upgrading
-  - apiVersion: v1
+  - apiVersion: "image.openshift.io/v1"
```
jlebon: Hmm, why these `apiVersion` changes? Did you check that this still works against v3.6?

manifests/pod.yaml
```diff
@@ -1,34 +1,121 @@
-# This pod template is used in the Jenkins pipeline by the Kubernetes plugin
+# This pod template provides the same functionality as pod.yaml, but includes a
+# a configMap to adding a CA-Certifiate to trust (i.e. private Git Repo).
```
jlebon: Did you mean to make this a different file? Changing from a pod definition to a template will break the pipeline unless we adjust the `Jenkinsfile` too, right?

manifests/pod.yaml
```diff
@@ -1,34 +1,121 @@
...
 
 apiVersion: v1
+kind: Template
```
dustymabe: can this pod.yaml file actually be an openshift template itself? It's consumed by the kubernetes plugin in the Jenkinsfile, so I'm not sure:

https://github.com/coreos/fedora-coreos-pipeline/blob/fb43e759bd73b7f2c6aade2d66d501687fb42cab/Jenkinsfile#L4

manifests/pod.yaml
darkmuggle: That was originally a separate file. I've fixed that.

manifests/pod.yaml
```diff
@@ -1,34 +1,120 @@
...
+
+objects:
+  - apiVersion: v1
+    kind: ConfigMap
+    metadata:
+      name: trust-ca-cert
+      namespace: ${NAMESPACE}
+    data:
+      cert: |
+        ${CERT_DATA}
```
dustymabe: Assuming that we can't use an openshift template here, we should be able to move this ConfigMap definition to pipeline.yaml, right?

darkmuggle: I would agree with that.

manifests/pod.yaml
```diff
@@ -1,34 +1,120 @@
...
+        secret:
+          secretName: duffy.key
+          optional: true
+      - name: trust-ca
+        configMap:
+          defaultMode: 420
+          items:
+           - key: cert
+             path: trust-ca.pem
+          name: trust-ca-cert
```
dustymabe: can we make this optional like we do for the duffy key so that it works in the case where we don't have any certs to add:

https://github.com/coreos/fedora-coreos-pipeline/blob/fb43e759bd73b7f2c6aade2d66d501687fb42cab/manifests/pod.yaml#L31-L34

HACKING.md
darkmuggle: Per another comment, TAGS are dropped in favor of hard-coded values.

manifests/pipeline.yaml
darkmuggle: Done

manifests/pipeline.yaml
darkmuggle: Agreed, and dropped. The initial thinking was to change as little as possible and that's why it was made a parameter.

manifests/pipeline.yaml
darkmuggle: Using a newer version of `oc`, it would fail without having the scoped version. I haven not tried it against 3.6. In my test, 3.11 would fail without scoping it.

dustymabe:
is this ready for another round of review?

darkmuggle:
Actually, I'm going to withdrawal the PR and resubmit. I'm hitting an issue where in order to make this work for my use-case, it will introduce a ton of complexity; one of the things I like about this pipeline is its simple.

dustymabe:
:+1:

PR #43, opened 2019-02-06
Build metal-bios and metal-uefi

cgwalters:
Let's get these images out there for people to try.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -48,7 +48,7 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
 
         stage('Build') {
             utils.shwrap("""
-            coreos-assembler build --skip-prune
+            coreos-assembler build --skip-prune qemu metal-bios metal-uefi
             """)
         }
 


dustymabe:
I don't think we can do this until we've compressed these since the raw files are ~ 2G (8G virtual, 2G apparent size). Our PV is 50G right now and we are keeping 10 builds around so this would stretch us.

cgwalters:
`coreos-assembler compress` should do it right?

dustymabe:
> `coreos-assembler compress` should do it right?

you are right.. weird. I forgot we added that to this pipeline already.. we did that in #34 
just verified locally it worked:

```
    "images": {
        "qemu": {
            "path": "fedora-coreos-29.5-qemu.qcow2.gz",
            "sha256": "03585926b4d1668091b8b4a566b1d17bab2b855ab10263e6fb9275ac1789f793",
            "size": "1957036032",
            "uncompressed-sha256": "e2ea4c13473f49c2a9c3e8192c1ed518f6a224463d0dcc738e4818fbd115367d"
        },
        "metal-uefi": {
            "path": "fedora-coreos-29.5-metal-uefi.raw.gz",
            "sha256": "76aa4c721669033737c73a83630a936674bf7aaf727df802129fe6751529a9c9",
            "size": "8589934592",
            "uncompressed-sha256": "f7b9051f12f79972aeaa7ea5278cf8e04c86142ac83a77e8f554db9f069e568f"
        }
    },

```

cgwalters:
This came from https://github.com/coreos/coreos-assembler/pull/305

PR #46, opened 2019-02-22
pvc: use coreos-assembler-claim2 in centos CI

dustymabe:
Brian created us a new PVC and the claim has a new name because
we needed the other one around while we copied it over.

### DIFF
--- a/manifests/pipeline.yaml
+++ b/manifests/pipeline.yaml
@@ -123,7 +123,7 @@ objects:
   - apiVersion: v1
     kind: PersistentVolumeClaim
     metadata:
-      name: coreos-assembler-claim
+      name: coreos-assembler-claim2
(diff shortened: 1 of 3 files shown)


jlebon:
How about we rename the old PVC to `-old` and the new one to the current name? Feels cleaner than updating the manifest itself.

dustymabe:
> How about we rename the old PVC to `-old` and the new one to the current name? Feels cleaner than updating the manifest itself.

I don't think you can rename them:

```
$ oc edit pvc coreos-assembler-claim
A copy of your changes has been stored to "/tmp/oc-edit-3i5y1.yaml"
error: At least one of apiVersion, kind and name was changed

```

jlebon:
Yeah, that's too bad. OK, let's do this then!

PR #48, opened 2019-02-22
pipeline: make PVC_SIZE a template parameter

jlebon:
Allow developers to easily customize it if they don't have 100Gi to
spare.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -176,6 +176,10 @@ This template creates:
 4. the `PersistentVolumeClaim` in which we'll cache and compose, and
 5. the Jenkins pipeline build.
 
+The default size of the PVC is 100Gi. There is a `PVC_SIZE`
+parameter one can use to make this smaller if you do not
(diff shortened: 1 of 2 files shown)


HACKING.md
```diff
@@ -176,6 +176,17 @@ This template creates:
...
+The default size of the PVC is 100Gi. There is a `PVC_SIZE`
+parameter one can use to make this smaller if you do not
+have enough space. E.g.:
+
+```
+oc new-app --file=manifests/pipeline.yaml \
+    --param=REPO_URL=https://github.com/jlebon/fedora-coreos-pipeline \
+    --param=REPO_REF=my-feature-branch \
+    --param=PVC_SIZE=30Gi
+```
```
dustymabe: could we just add the PVC_SIZE param in the above example that already exist?

jlebon: I wanted to keep it after saying that it creates the PVC just higher up. I reduced it to a one-liner instead of repeating it again.

PR #49, opened 2019-02-22
HACKING.md: match v3.6.1

jlebon:
There are minor inconsistencies otherwise between the CLI and the
server. E.g. `new-app` will fail to find templates on v3.9. Or editing
some aspects of imagestreams will fail if not with v3.6.1.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -18,10 +18,10 @@ dnf install -y oci-kvm-hook # if on traditional
 ```
 
 We will use `oc cluster up` to set up a local cluster for testing. To do
-this, simply obtain the OpenShift binary from
-[the latest release](https://github.com/openshift/origin/releases/latest)
-(though note that I'm currently still using the binaries from v3.9, so
-pick that release if you hit issues).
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -18,10 +18,8 @@ dnf install -y oci-kvm-hook # if on traditional
 ```
 
 We will use `oc cluster up` to set up a local cluster for testing. To do
-this, simply obtain the OpenShift binary from
-[the latest release](https://github.com/openshift/origin/releases/latest)
-(though note that I'm currently still using the binaries from v3.9, so
-pick that release if you hit issues).
+this, simply obtain the OpenShift v3.6.1 binary from
+[here](https://github.com/openshift/origin/releases/tag/v3.6.1).
```
dustymabe: maybe we should mention why we're choosing 3.6.1 (or do we do that somewhere else).

jlebon: We do mention it down, but this is the first time, so I moved the explanation here instead.

PR #50, opened 2019-02-22
pipeline: add openstack to output artifacts

dustymabe:


dustymabe:
community member asked for this. might as well start making them now that we have larger PV

PR #54, opened 2019-04-29
Bump Jenkins to v2.150.2

jlebon:
Tested working locally.


jlebon:
This is not the latest LTS, but it's the latest image available on Quay right now. I've inquired about the latest in https://github.com/openshift/jenkins/pull/833#issuecomment-487682965.

jlebon:
Thanks!

PR #56, opened 2019-05-06
pipeline: adapt for split metal building

jlebon:
Building metal images is now a separate step since:
https://github.com/coreos/coreos-assembler/pull/510


cgwalters:
LGTM!

dustymabe:
👍

Issue #57, opened 2019-05-24
Support developer pipelines


jlebon:
So, being able to easily stand up your own cluster is great, though it limits our ability to interact with things that use secrets, like the artifact server and S3. Just like RHCOS, it'd be nice to support developer pipelines that stand alongside the prod one.

The "standup a local cluster" path is still useful when hacking on some part of the OpenShift manifests themselves though, and keeping that path "hot" ensures we don't rely on live edits of the running instance. (Related: the OpenShift CI cluster has this cool thing that reconfigures the live cluster when changes to [openshift/release](https://github.com/openshift/release) are merged in. Would be nice to have that functionality here too.)

jlebon:
Likely going to implement this in the process of adding S3 support.

jlebon:
Fixed in #62.

PR #59, opened 2019-05-28
pipeline: upload builds to S3

jlebon:
This is a first step towards switching to S3. We still rsync to the
artifact server, but we also upload new builds to S3. Once we confirm
this works nicely, we can switch over to using `buildprep` and
completely wean off the artifact server.

---

Requires: https://github.com/coreos/coreos-assembler/pull/527

I tested this from my local cluster and it works, though I'm also working on #57 to make it easier to test changes like these.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,6 +30,9 @@ properties([
     ])
 ])
 
+// see bucket layout in https://github.com/coreos/fedora-coreos-tracker/issues/189
+def s3_builddir = "fcos-builds/prod/streams/${params.STREAM}/builds"
+
 podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultContainer: 'jnlp') {
     node('coreos-assembler') { container('coreos-assembler') {
 
@@ -46,6 +49,15 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
         }
 
         stage('Fetch') {
+            /*
+            // XXX: uncomment once we have a build there
+            if (prod && utils.path_exists("/.aws")) {
+                utils.shwrap("""
+                coreos-assembler buildprep s3://${s3_builddir}
+                """)
+            }
+            */
+
             if (prod) {
                 // make sure our cached version matches prod exactly before continuing
                 utils.rsync_in("builds", "builds")
@@ -101,6 +113,7 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
         }
 
         stage('Prune') {
+            // XXX: stop pruning like this when we fully drop artifact server
             utils.shwrap("""
             coreos-assembler prune --keep=8
             """)
@@ -125,6 +138,7 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
 
             // Change perms to allow reading on webserver side.
             // Don't touch symlinks (https://github.com/CentOS/sig-atomic-buildscripts/pull/355)
+            // XXX: can drop this when dropping artifact server
             utils.shwrap("""
             find builds/ ! -type l -exec chmod a+rX {} +
             """)
@@ -134,6 +148,13 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             // https://stackoverflow.com/questions/1636889
             if (prod) {
                 utils.rsync_out("builds", "builds")
+            }
+
+            if (prod && utils.path_exists("/.aws")) {
+              // XXX: just upload as public-read for now
+              utils.shwrap("""
+              coreos-assembler buildupload s3 --acl=public-read ${s3_builddir}
+              """)
             }
         }
     }}
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,22 +1,24 @@
-def pod, utils, devel
+def pod, utils, prod
 node {
     checkout scm
     pod = readFile(file: "manifests/pod.yaml")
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -120,6 +138,12 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             // https://stackoverflow.com/questions/1636889
             if (!devel) {
                 utils.rsync_out("builds", "builds")
+                if (utils.path_exists("/.aws")) {
```
dustymabe: any preference on just using ENV vars here instead of a config file there?

jlebon: Heh, I explicitly chose to move away from env vars actually to raise the barrier a bit more on how easily credentials could leak e.g. into logs.

cgwalters: Side note, for AWS it's best to use https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html if you can.

dustymabe: > Heh, I explicitly chose to move away from env vars actually to raise the barrier a bit more on how easily credentials could leak e.g. into logs.

yeah. i've just always done it the other way so I have a slight preference. We can stick with this.

jlebon:
OK, this is updated and tested now!

You can use #62 to test this:

```
$ ./devel-up --pipeline https://github.com/jlebon/fedora-coreos-pipeline@pr/devel-pipeline
Parameters:
  DEVEL_PREFIX=jlebon-
  PIPELINE_REPO_URL=https://github.com/jlebon/fedora-coreos-pipeline
  PIPELINE_REPO_REF=pr/devel-pipeline
imagestream "jlebon-coreos-assembler" configured
buildconfig "jlebon-fedora-coreos-pipeline" configured

You may start your dev pipeline with:
  oc start-build jlebon-fedora-coreos-pipeline
$ oc start-build jlebon-fedora-coreos-pipeline
...
```

Wait for build to finish. Then e.g.:

```
$ aws s3 ls fcos-builds/devel/streams/jlebon/30.237/
2019-05-31 15:03:44      31874 commitmeta.json
2019-05-31 15:03:43        486 coreos-assembler-config-git.json
2019-05-31 15:03:42      11601 coreos-assembler-config.tar.gz
2019-05-31 15:03:09   57889466 fedora-coreos-30.237-installer-initramfs.img
2019-05-31 15:03:08    8863944 fedora-coreos-30.237-installer-kernel
2019-05-31 15:03:02   76756992 fedora-coreos-30.237-installer.iso
2019-05-31 15:01:49  613107738 fedora-coreos-30.237-metal-bios.raw.gz
2019-05-31 15:01:58  612779524 fedora-coreos-30.237-metal-uefi.raw.gz
2019-05-31 15:03:13  612640074 fedora-coreos-30.237-openstack.qcow2.gz
2019-05-31 15:01:38  612645539 fedora-coreos-30.237-qemu.qcow2.gz
2019-05-31 15:03:22  626155520 fedora-coreos-30.237-vmware.ova
2019-05-31 15:03:43      68387 install.log
2019-05-31 15:03:54      10255 meta.json
2019-05-31 15:03:44      14604 ostree-commit-object
2019-05-31 15:03:45  557957120 ostree-commit.tar
```

One obvious thing here is that we're re-using the same workdir as prod, though that's not an urgent issue right now since (1) Jenkins won't run them concurrently and (2) we always rsync in the prod data from the artifact server.

Still, once we move away from the artifact server (which we should be able to do as soon as this code runs once), I'll clean that up.

jlebon:
> Still, once we move away from the artifact server (which we should be able to do as soon as this code runs once), I'll clean that up.

This is fixed in #62 now.

Jenkinsfile
jlebon: > Side note, for AWS it's best to use https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html if you can.

Hmm neat, hadn't seen this before. I don't want to address this right now in this patch, but I don't want to lose track of it either, so filed https://github.com/coreos/fedora-coreos-pipeline/issues/63.

Jenkinsfile
```diff
@@ -28,6 +30,8 @@ properties([
     ])
 ])
 
+def s3_builddir = "fcos-builds/prod/streams/${params.STREAM}"
```
dustymabe: where is this path coming from? did we define our s3 bucket layout anywhere?

dustymabe:
gave this a quick review - had [one comment](https://github.com/coreos/fedora-coreos-pipeline/pull/59#discussion_r290061239). will try out with devel pipelines next.

Jenkinsfile
jlebon: That's https://github.com/coreos/fedora-coreos-tracker/issues/189, right?

jlebon: I added a comment with the link. We can definitely still move builds around after this merges if discussions in that ticket lead to a different layout.

dustymabe: > That's [coreos/fedora-coreos-tracker#189](https://github.com/coreos/fedora-coreos-tracker/issues/189), right?

right. I'm actually OK with the layout (as you mentioned being defined over in that ticket. I think we should make the bucket name a parameter. We may be using `fcos-builds` but someone without access to that bucket may want to try to setup a pipeline and we should make it possible.

jlebon: I added a patch to #62 instead to fix this because this intersects with the devel path and it would've been messy to understand the changes if they were split across here and there.

dustymabe:
Assuming this PR is updated with all the content from #62  this LGTM

dustymabe:
LGTM - i'll try to enhance the docs in a future PR

PR #61, opened 2019-05-31
Prep patches for S3 & devel

jlebon:
These are split out of #59 and should be straightforward to review. The first two have no impact on the current pipeline. The last one slightly modifies how initialization is performed.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -16,7 +16,16 @@ node {
 
 properties([
     disableConcurrentBuilds(),
-    pipelineTriggers(devel ? [] : [cron("H/30 * * * *")])
+    pipelineTriggers(devel ? [] : [cron("H/30 * * * *")]),
+    parameters([
+      choice(name: 'STREAM',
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -16,17 +16,30 @@ node {
...
-            if [ ! -d src/config ]; then
-                coreos-assembler init https://github.com/coreos/fedora-coreos-config
-            fi
+            # just always restart from scratch in case it's a devel pipeline
+            # and it changed source url or ref; this info also makes it into
+            # the build metadata through cosa reading the origin remote
+            rm -rf src/config
+
+            # in the future, the stream will dictate the branch in the prod path
+            coreos-assembler init --force https://github.com/coreos/fedora-coreos-config
```
dustymabe: I wonder if the URL should be templatized

dustymabe:
LGTM

Jenkinsfile
jlebon: Yup, that's coming up in a next patch. :)

PR #62, opened 2019-05-31
Add support for developer pipelines

jlebon:
So, a lot happening in this patch, but essentially the idea is:

1. add some more parameters to the template to support "devel" resources
2. add annotations to the pipeline buildconfig to allow us to pass data
   to the pipeline code; update the pipeline to pick this up
3. allow uploading to S3 in devel mode using the provided devel prefix
4. add a new `devel-up` utility to make it easier and safer to create
   and update devel pipelines; no way to delete them yet, will do that
   in a follow up

The reason for the `devel-up` utility is that I wanted more "semantic"
control over the way devel resources are created than through directly
using parameters, especially around ensuring that they absolutely do not
modify prod resources. But also, it's much more convenient to use than
overriding each param one by one.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -6,6 +6,21 @@ These instructions will cover both local developer workflows, as well as
 interacting with the production pipeline in CentOS CI. A section header
 may indicate that the section only applies to one of the two workflows.
 
+Another option if one simply wants to test a different COSA image or
+FCOS config or pipeline code (but otherwise maintain the manifest the
+same) is to create a developer pipeline alongside the production one. To
+do this, you *must* use the `devel-up` script. For example:
+
+```
+./devel-up --update \
+        --pipeline https://github.com/jlebon/fedora-coreos-pipeline \
+        --config https://github.com/jlebon/fedora-coreos-config@feature \
+        --cosa-img quay.io/jlebon/coreos-assembler:random-tag
+```
+
+Now carrying on with the instructions for setting up the environment
+from scratch...
+
 ### [LOCAL] Set up an OpenShift cluster
 
 First, make sure to install `oci-kvm-hook` on your host system (not in a
@@ -151,18 +166,23 @@ the Jenkins pipeline).
 
 ### Create the pipeline from the template
 
-If working on the production pipeline, or you're not
-planning any modifications, you may simply do:
+If working on the production pipeline, you may simply do:
 
 ```
 oc new-app --file=manifests/pipeline.yaml
 ```
 
-If working on your own repo, you will want to override the
+If working on a local cluster you will want to define the `DEVEL_PREFIX`
+parameter to be e.g. `$USER-`. The pipeline will simply not run without
+a `DEVEL_PREFIX` set if it is not running in prod mode.
+
+There are additional useful parameters for the devel case. For example,
+to use a custom pipeline repo, you'll want to override the
 `PIPELINE_REPO_URL` and `PIPELINE_REPO_REF` parameters:
 
 ```
 oc new-app --file=manifests/pipeline.yaml \
+    --param=DEVEL_PREFIX=$(whoami)- \
     --param=PIPELINE_REPO_URL=https://github.com/jlebon/fedora-coreos-pipeline \
     --param=PIPELINE_REPO_REF=my-feature-branch
 ```
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,4 +1,4 @@
-def pod, utils, prod
+def pod, utils, prod, devel_prefix, src_config_url, src_config_ref
 node {
     checkout scm
     pod = readFile(file: "manifests/pod.yaml")
@@ -13,6 +13,18 @@ node {
         echo "Running in prod mode."
     } else {
         echo "Running in devel mode on ${env.JENKINS_URL}."
+    }
+
+    devel_prefix = utils.get_pipeline_annotation('devel-prefix')
+    src_config_url = utils.get_pipeline_annotation('source-config-url')
+    src_config_ref = utils.get_pipeline_annotation('source-config-ref')
+
+    // sanity check that a valid prefix is provided if in devel mode and drop
+    // the trailing '-' in the devel prefix
+    if (!prod) {
+      assert devel_prefix.length() > 0 : "Missing devel prefix"
+      assert devel_prefix.endsWith("-") : "Missing trailing dash in devel prefix"
+      devel_prefix = devel_prefix[0..-2]
     }
 }
 
@@ -34,10 +46,31 @@ properties([
 ])
 
 // see bucket layout in https://github.com/coreos/fedora-coreos-tracker/issues/189
-def s3_builddir = "fcos-builds/prod/streams/${params.STREAM}/builds"
+def s3_builddir
+if (prod) {
+  s3_builddir = "fcos-builds/prod/streams/${params.STREAM}/builds"
+  pod = pod.replace("COREOS_ASSEMBLER_IMAGE", "coreos-assembler:master")
+} else {
+  // One prefix = one pipeline = one stream; the devel-up script is geared
+  // towards testing a specific combination of (cosa, pipeline, fcos config),
+  // not a full duplication of all the prod streams. One can always instantiate
+  // a second prefix to test a separate combination if more than 1 concurrent
+  // devel pipeline is needed.
+  s3_builddir = "fcos-builds/devel/streams/${devel_prefix}/builds"
+  pod = pod.replace("COREOS_ASSEMBLER_IMAGE", "${devel_prefix}-coreos-assembler:master")
+}
 
 podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultContainer: 'jnlp') {
     node('coreos-assembler') { container('coreos-assembler') {
+
+        // Only use the PVC for prod caching. For devel pipelines, we just
+        // always refetch from scratch: we don't want to allocate cached data
+        // for pipelines which may only run once.
+        if (prod) {
+          utils.workdir = "/srv"
+        } else {
+          utils.workdir = env.WORKSPACE
+        }
 
         stage('Init') {
             utils.shwrap("""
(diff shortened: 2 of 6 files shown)


jlebon:
(The first three commits are from #59, but it's probably easier to *test* #59, using this patch, as described in https://github.com/coreos/fedora-coreos-pipeline/pull/59#issuecomment-497827768).

jlebon:
Now also with the ability to delete the pipeline:

```
$ ./devel-up --update \
    --pipeline https://github.com/jlebon/fedora-coreos-pipeline@devel-pipeline
Parameters:
  DEVEL_PREFIX=jlebon-
  PIPELINE_REPO_URL=https://github.com/jlebon/fedora-coreos-pipeline
  PIPELINE_REPO_REF=devel-pipeline

Updating:
  imagestream "jlebon-coreos-assembler" created
  buildconfig "jlebon-fedora-coreos-pipeline" created

You may start your devel pipeline with:
  oc start-build jlebon-fedora-coreos-pipeline

$ ./devel-up --delete
Parameters:
  DEVEL_PREFIX=jlebon-

Deleting:
  imagestream "jlebon-coreos-assembler" deleted
  buildconfig "jlebon-fedora-coreos-pipeline" deleted
```

manifests/pipeline.yaml
```diff
@@ -117,7 +130,7 @@ objects:
             kind: DockerImage
             # for now we just follow :master, but we may start freezing if
             # things become too unstable
```
dustymabe: does this comment need to be moved to the definition of `COREOS_ASSEMBLER_IMAGE` ?

jlebon: Thanks, done!

HACKING.md
```diff
@@ -6,6 +6,21 @@ These instructions will cover both local developer workflows, as well as
...
+FCOS config or pipeline code (but otherwise maintain the manifest the
+same) is to create a developer pipeline alongside the production one. To
+do this, you *must* use the `devel-up` script. For example:
+
+```
+./devel-up \
+        --pipeline https://github.com/jlebon/fedora-coreos-pipeline \
+        --config https://github.com/jlebon/fedora-coreos-config@feature \
+        --cosa-img quay.io/jlebon/coreos-assembler:random-tag
+```
```
dustymabe: `devel-up: error: one of the arguments --update --delete is required`

devel-up
```diff
@@ -0,0 +1,149 @@
...
+    action = parser.add_mutually_exclusive_group(required=True)
+    action.add_argument("--update", action='store_true',
+                        help="Create or update devel pipeline")
+    action.add_argument("--delete", action='store_true',
+                        help="Delete devel pipeline")
+    parser.add_argument("--start", action='store_true',
+                        help="Start build after updating it")
+    parser.add_argument("--prefix", help="Devel prefix to use for resources",
+                        default=get_username())
+    parser.add_argument("--pipeline", metavar='<URL>[@REF]',
```
dustymabe: is `@REF` a convention? I know at least `oc new-build` uses `#`:

```
  # Create a build config from a remote repository using its beta2 branch
  oc new-build https://github.com/openshift/ruby-hello-world#beta2
```

dustymabe:
ok ran it against a fresh cluster:

```
$ ./devel-up --update --pipeline https://github.com/dustymabe/fedora-coreos-pipeline@PR62 --config https://github.com/dustymabe/fedora-coreos-config@dusty                             
Parameters:
  DEVEL_PREFIX=dustymabe-
  PIPELINE_REPO_URL=https://github.com/dustymabe/fedora-coreos-pipeline
  PIPELINE_REPO_REF=PR62
  PIPELINE_FCOS_CONFIG_URL=https://github.com/dustymabe/fedora-coreos-config
  PIPELINE_FCOS_CONFIG_REF=dusty

Updating:
  imagestream.image.openshift.io/dustymabe-coreos-assembler created
  buildconfig.build.openshift.io/dustymabe-fedora-coreos-pipeline created

You may start your devel pipeline with:
  oc start-build dustymabe-fedora-coreos-pipeline

```

which looks good but I the pods to run the build never start because `coreos-assembler-claim2` isn't satisfiled. For some reason we aren't making a claim because it's not a prefixed item (we only seem to be operating on prefixed items). What do we do about that?

devel-up
jlebon: Inspired by ci-operator: https://github.com/openshift/ci-operator/#usage. I like that `@` actually means "at" which is appropriate here. I'm open to switching it though.

jlebon:
That said, re. the PVC specifically, we don't actually need it for devel pipelines, so we could make it optional. Though I think that'd require splitting it out of `pipeline.yaml` and having a "prod-only-resources" bit on the side, which is awkward.

HACKING.md
jlebon: Thanks, fixed!

devel-up
dustymabe: ok - yeah I didn't know where it came from. I also don't know if there is a case where we'll want to delineate between branch or commit hash. In that case we could have used `#` for branch and `@` for commit hash. It's cool to leave it like it is.

dustymabe:
> Right, you would only use `devel-up` on a namespace which is already set up.

i'm still working through it all but I think it would be extremely useful to be able to target this at a new project (i.e. blank slate) and have it work. Let me revisit this opinion once I've worked out all the kinks.

jlebon:
Just going to copy something I wrote on IRC too:

> so one thing we could do here is add a `--all` switch to `devel-up` so that it defines all the resources when targeting a cluster that's not cci. then the `HACKING.md` docs become: `new-app` once for the Jenkins instance, then directly use `devel-up --update --all`

jlebon:
Re. the buildconfig annotation approach, I like it though I wouldn't be opposed to making them proper Jenkins parameters either. What I was trying to avoid with this was requiring devel setups having to manually be triggered through the "Build with Parameters" UI instead of just through `oc start-build`. I'm inclined to just run with it for now, and see if it causes us too much trouble down the line.

Jenkinsfile
```diff
@@ -1,22 +1,37 @@
-def pod, utils, devel
+def pod, utils, prod, devel_prefix, src_config_url, src_config_ref
```
dustymabe: add `s3_bucket` ?

Jenkinsfile
```diff
@@ -39,12 +86,19 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
...
             # in the future, the stream will dictate the branch in the prod path
-            coreos-assembler init --force https://github.com/coreos/fedora-coreos-config
+            coreos-assembler init --force --branch ${src_config_ref} ${src_config_url}
             """)
         }
 
         stage('Fetch') {
-            if (!devel) {
+            // XXX: drop `!prod && ` once we've uploaded prod builds there
+            if (!prod && s3_builddir)
```
dustymabe: missing `{` ?

Jenkinsfile
```diff
@@ -39,12 +94,19 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
...
         }
 
         stage('Fetch') {
-            if (!devel) {
+            // XXX: drop `!prod && ` once we've uploaded prod builds there
+            if (!prod && s3_builddir) {
+                utils.shwrap("""
+                coreos-assembler buildprep s3://${s3_builddir}
+                """)
+            }
```
dustymabe: i don't want to address this in this PR (we're very close to merging) but would like to brainstorm on how we make this step non-fatal in cases where we are starting with a fresh s3 bucket (i.e. someone set up a new s3 bucket and just brought this up in openshift for the first time).

jlebon: Hmm yeah, `buildprep` has:

```python
    builds = fetcher.fetch_json('builds.json')
    if len(builds) == 0:
        print("Remote has no builds!")
        return
```

though that still assumes that `builds.json` is present. Just need to tweak that.

IOW, I think this is something we should fix in cosa, not necessarily here.

dustymabe: > IOW, I think this is something we should fix in cosa, not necessarily here.

we'd probably have to handle different errors differently, though? in one case builds.json doesn't exist and it's the first build and we want to go on. In the other case builds.json does exist but there was an error retrieving it (network issue, permissions issue, etc).

dustymabe:
one new issue I see:

If originally call `oc new-app --file=manifests/pipeline.yaml` with `--param=S3_BUCKET=dusty-fcos` and then subsequently call `devel-up --update` the annotation on the buildconfig gets set back to `fcos-builds`. This isn't what we want IMHO.

Jenkinsfile
jlebon: Yeah, I think we should be able to key off of the response code in both cases. Will look into this.

dustymabe:
here is a pkgdiff from two consecutive runs (i added the `htop` rpm):


```
    "pkgdiff": [
        [
            "htop",
            0,
            {
                "NewPackage": [
                    "htop",
                    "2.2.0-4.fc30",
                    "x86_64"
                ]
            }
        ],
        [
            "fedora-coreos-overlay",
            2,
            {
                "PreviousPackage": [
                    "fedora-coreos-overlay",
                    "1559760183-098f5d8",
                    "noarch"
                ],
                "NewPackage": [
                    "fedora-coreos-overlay",
                    "1559763067-098f5d8",
                    "noarch"
                ]
            }
        ]
    ],
```

one interesting thing is to see that the `fedora-coreos-overlay` is thought to have changed. We are no longer operating on a local cache, right? How do we keep that as not being a "change"?

dustymabe:
> one interesting thing is to see that the `fedora-coreos-overlay` is thought to have changed. We are no longer operating on a local cache, right? How do we keep that as not being a "change"?

opened https://github.com/coreos/coreos-assembler/issues/544 for this.

Jenkinsfile
dustymabe: opened https://github.com/coreos/coreos-assembler/issues/543 for this

jlebon:
OK, now rebased! Prod pipeline still happy.

PR #64, opened 2019-06-04
manifests/pipeline: remove namespace params from template

dustymabe:
This allows for someone running in a different namespace/project
than `fedora-coreos`.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -140,7 +140,6 @@ like `fedora-coreos-devel` if you'd like.
 
 ```
 oc new-project fedora-coreos
-oc delete project myproject
 ```
 
 And you're all set!
--- a/HACKING.md
+++ b/HACKING.md
@@ -134,7 +134,9 @@ oc login -u developer https://127.0.0.1:8443
 (Any password will work). The IP address to log in here may differ
 according to the output from `oc cluster up`.
 
-We'll want to match the project name used in CentOS CI:
+Ideally you will match the project name used for prod in CentOS CI
+(`fedora-coreos`), but feel free to use a different project name
+like `fedora-coreos-devel` if you'd like.
 
 ```
 oc new-project fedora-coreos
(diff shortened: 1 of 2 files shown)


dustymabe:
@jlebon - is there a reason to leave these in the template?

jlebon:
I think it *used* to be required because of the imagestream, but isn't anymore.

Maybe update the `HACKING.md` too, though mention that the prod one must still be called `fedora-coreos`? (The `prod` var keys off of this in the Jenkinsfile).

dustymabe:
ok - added a note to HACKING.md - let me know if there is anywhere else where we should clarify!

HACKING.md
```diff
@@ -134,7 +134,9 @@ oc login -u developer https://127.0.0.1:8443
 (Any password will work). The IP address to log in here may differ
 according to the output from `oc cluster up`.
 
-We'll want to match the project name used in CentOS CI:
+Ideally you will match the project name used for prod in CentOS CI
+(`fedora-coreos`), but if feel free to use a different project name
```
jlebon: Minor: s/if // ?

dustymabe: +1

dustymabe:
pushed up a fix and also added one minor commit so might be worth one more look

HACKING.md
```diff
@@ -140,7 +140,6 @@ like `fedora-coreos-devel` if you'd like.
 
 ```
 oc new-project fedora-coreos
-oc delete project myproject
```
jlebon: The default `oc cluster up` behaviour starts you up with a project called `myproject`, right?
Oh or are you saying there's no need to go out of our way to delete it?

dustymabe: yeah mostly you don't need to go out of your way to delete it. It's also not really necessary and may even be destructive if you target an existing cluster rather than going the `oc cluster up` route.

jlebon:
Fair enough. LGTM!

PR #66, opened 2019-06-06
pipeline: add FORCE parameter to force rebuild

jlebon:
When testing pipeline and/or cosa changes, one sometimes wants to run
through the whole thing even if nothing changed in the input.


jlebon:
Need this to test the prod pipeline changes after #59 before also merging #62.

(Tested this patch itself using #62.)

dustymabe:
LGTM

PR #67, opened 2019-06-06
pipeline: fully move to S3

jlebon:
Now that we have a prod build in S3, we can fully switch over to it.
Most of the changes are straightforward. One note of interest: the
"pruning" stage is now really just about pruning the local cache. For
now, we're not pruning from the bucket at all, pending a more defined
policy & mechanism.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -102,16 +102,10 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
         }
 
         stage('Fetch') {
-            // XXX: drop `!prod && ` once we've uploaded prod builds there
-            if (!prod && s3_builddir) {
+            if (s3_builddir) {
                 utils.shwrap("""
                 coreos-assembler buildprep s3://${s3_builddir}
                 """)
-            }
-
-            if (prod) {
-                // make sure our cached version matches prod exactly before continuing
-                utils.rsync_in("builds", "builds")
             }
 
             utils.shwrap("""
@@ -164,10 +158,9 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             """)
         }
 
-        stage('Prune') {
-            // XXX: stop pruning like this when we fully drop artifact server
+        stage('Prune Cache') {
             utils.shwrap("""
-            coreos-assembler prune --keep=8
+            coreos-assembler prune --keep=1
             """)
 
             // If the cache img is larger than e.g. 8G, then nuke it. Otherwise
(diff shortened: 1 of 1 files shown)


jlebon:
(Still testing this.)

jlebon:
OK, dropping WIP, now tested! :heavy_check_mark:

Jenkinsfile
```diff
@@ -102,18 +102,12 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
         }
 
         stage('Fetch') {
-            // XXX: drop `!prod && ` once we've uploaded prod builds there
-            if (!prod && s3_builddir) {
```
dustymabe: so we ran this once already? or do we need https://github.com/coreos/coreos-assembler/pull/545 first?

dustymabe:
this mostly LGTM. only concern I have is if we (or anyone else) want to set up another pipeline and not touch s3. Would keeping the rsync functionality around be useful ?

Jenkinsfile
jlebon: We did `buildupload` once already, so we have something to `buildprep` from. https://github.com/coreos/coreos-assembler/pull/545 is for the case where there's nothing to `buildprep` from (which is mostly relevant for devel pipelines).

jlebon:
OK, I've reworked this for now so that we still do upload the latest build to the artifact server. From comment:

```groovy
// XXX: For now, we keep uploading the latest build to the artifact
// server to make it easier for folks to access since we don't have
// a stream metadata frontend/website set up yet. The key part here
// is that it is *not* the canonical storage for builds.
```

jlebon:
> Would keeping the rsync functionality around be useful ?

So as per the previous comment, I've kept it for now for another reason. Though once we're ready to move on for good, I think I'd rather we drop it completely until someone actually shows up with that use case. `git` will be there, remembering everything.

Note for the no-S3 devel pipeline, we do store builds in the PVC. So one should be able to just bring up e.g. simple-httpd to access it over HTTP if they'd like.

dustymabe:
> OK, I've reworked this for now so that we still do upload the latest build to the artifact server. From comment:
> 
> ```groovy
> // XXX: For now, we keep uploading the latest build to the artifact
> // server to make it easier for folks to access since we don't have
> // a stream metadata frontend/website set up yet. The key part here
> // is that it is *not* the canonical storage for builds.
> ```

hmm i'd prefer to just open up the bucket and make it brows-able IMHO



> So as per the previous comment, I've kept it for now for another reason. Though once we're ready to move on for good, I think I'd rather we drop it completely until someone actually shows up with that use case. `git` will be there, remembering everything.

I'm OK with letting `git` store it 👍

jlebon:
> hmm i'd prefer to just open up the bucket and make it brows-able IMHO

I thought consensus was to *not* have it browsable, right? Because we don't want users mucking around in there in the first place. See e.g. https://github.com/coreos/fedora-coreos-tracker/issues/169#issuecomment-497522046.

dustymabe:
> > hmm i'd prefer to just open up the bucket and make it brows-able IMHO
> 
> I thought consensus was to _not_ have it browsable, right? Because we don't want users mucking around in there in the first place. See e.g. [coreos/fedora-coreos-tracker#169 (comment)](https://github.com/coreos/fedora-coreos-tracker/issues/169#issuecomment-497522046).

yeah should have clarified *make it brows-able until we get a release browswer*

either way I guess we can still rsync the content out and just point people there for now.

jlebon:
AFAIK, S3 doesn't have a native "web browser" (like e.g. the Apache listing). Making the bucket "browsable" would mean indexing it with something like https://github.com/projectatomic/papr/blob/master/papr/utils/indexer.py which drops a bunch of `index.html` files everywhere. :grimacing:

PR #70, opened 2019-06-11
Update docs so we can use jenkins:2

dustymabe:
Update the docs so that we can be more compatible with latest templates.
We update the input ImageStreamTag for our jenkins deployment to match
latest openshift (`jenkins:2`)

related: https://github.com/coreos/fedora-coreos-pipeline/pull/32

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -165,6 +165,18 @@ reason we create the app first is that otherwise OpenShift will
 automatically instantiate Jenkins with default parameters when creating
 the Jenkins pipeline).
 
+Next, let's update the input ImageStreamTag for our Jenkins deployment
+to match latest OpenShift (`jenkins:2`). Some older versions of the
+template in OpenShift uses `jenkins:latest`. This will no longer be
+needed once we are running on a newer version of OpenShift than 3.6 in
+CentOS CI. See [#32](https://github.com/coreos/fedora-coreos-pipeline/pull/32)
+and [#70](https://github.com/coreos/fedora-coreos-pipeline/pull/70)
+for context on why this may be needed):
+
+```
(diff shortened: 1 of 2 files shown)


dustymabe:
i'm going to cascade these PRs so I can test them all together - now requires #64

jlebon:
To confirm, nothing actually breaks today as a result of using `:latest`, right? This is solely to make transitioning to a new OCP easier in the future?

HACKING.md
```diff
@@ -164,6 +166,18 @@ reason we create the app first is that otherwise OpenShift will
 automatically instantiate Jenkins with default parameters when creating
 the Jenkins pipeline).
 
+Next, let's update the input ImageStreamTag for our jenkins deployment
+to match latest openshift (`jenkins:2`). Some older versions of the
+template in openshift uses `jenkins:latest`. This will no longer be
+needed once we are running on a newer version of OpenShift than 3.6 in
+CentOS CI. See [#32](https://github.com/coreos/fedora-coreos-pipeline/pull/32)
+and [#70](https://github.com/coreos/fedora-coreos-pipeline/pull/70)
+for context on why this may be needed):
```
jlebon: This is super minor/OCD, though can we consistently capitalize Jenkins and OpenShift? :)

dustymabe: done - let me know if you spot any I missed

dustymabe:
> To confirm, nothing actually breaks today as a result of using `:latest`, right? This is solely to make transitioning to a new OCP easier in the future?

Every time i deploy in newer openshift (pretty much anything other than centos ci as that's the only remaining openshift 3.6 cluster I know of) I have to `oc patch bc/fedora-coreos-jenkins -p '{"spec":{"output":{"to":{"name":"jenkins:2"}}}}'` so that my buildconfig will output to the right tag so my jenkins deployment will start. I'd prefer to not have to do that any longer.

The alternative to this approach is that we add that `oc patch bc` step to the notes that says "if you are on newer openshift, do this"

The reason I settled on the current approach is because you should be able to run the command I've added to the docs regardless of what version you are on and it should be a no-op if you are on a version that is new enough. So slightly less confusing.

PR #71, opened 2019-06-11
HACKING: add note about creating aws credentials secret

dustymabe:
We recently started uploading artifacts to s3. Let's add a note to the
docs about creating a secret in openshift with that information.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -150,6 +150,28 @@ If you're planning to test changes, it would be best to fork
 this repo so that you do your work there. The workflow
 requires a remote repo to which to push changes.
 
+### Creating AWS credentials config
+
+If you are in production where we upload builds to S3 OR you want to
+test uploading to S3 as part of pipeline development you need to
+create a credentials config as a secret within OpenShift.
+
+First create a file with your secret content:
+
+```
+cat <<'EOF' > /path/to/file
+[default]
+aws_access_key_id=keyid
+aws_secret_access_key=key
+EOF
+```
+
+Then create the secret in OpenShift:
+
+```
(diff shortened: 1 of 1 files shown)


dustymabe:
just review - i'll merge after #70

HACKING.md
```diff
@@ -150,6 +150,28 @@ If you're planning to test changes, it would be best to fork
 this repo so that you do your work there. The workflow
 requires a remote repo to which to push changes.
 
+### Creating AWS credentials config
+
+If you are in production where we upload builds to s3 OR you want to
+test uploading to s3 as part of pipeline development you need to
+create a credentials config as a secret within openshift.
```
jlebon: Minor: OpenShift and S3

HACKING.md
```diff
@@ -150,6 +150,28 @@ If you're planning to test changes, it would be best to fork
...
+
+First create a file named `config` (the file must be named `config`) with
```
jlebon: > (the file must be named `config`)

Should be able to just change the invocation below to `--from-file=config=/path/to/config` to make this more foolproof. (Though when I created it I used the older `oc secrets new` interface).

HACKING.md
```diff
@@ -150,6 +150,28 @@ If you're planning to test changes, it would be best to fork
...
+your secret content:
+
+```
+cat <<'EOF' > /path/to/config
+[default]
+aws_access_key_id=keyid
+aws_secret_access_key=key
```
jlebon: Missing `EOF`

dustymabe:
I think I took care of everything. Thanks for the tip about `--from-file=config=/path`

PR #72, opened 2019-06-11
HACKING: clear up notes around development pipelines

dustymabe:
Make it clear that developer pipelines are to be run on existing
clusters/deployments. Also add more context and notes about defaults
and existing behavior.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,25 +1,18 @@
-The `manifests/pipeline.yaml` file contains an OpenShift template that
+
+## Setting up a local or production cluster from scratch
+
+The instructions in this section will cover setting up a local (developer)
+cluster as well as a production cluster pipeline in CentOS CI. If you
+have access to an existing deployment and just want to set up a parallel
+"developer pipeline" skip ahead to the [Running A Developer Pipeline Alongside an Existing Deployment](#running-a-developer-pipeline-alongside-an-existing-deployment)
+section.
+
+As for setting up a new pipeline, one key piece of information is that
+the `manifests/pipeline.yaml` file contains an OpenShift template that
 will set up a Jenkins pipeline using the
 [Kubernetes Jenkins plugin](https://github.com/jenkinsci/kubernetes-plugin).
-
-These instructions will cover both local developer workflows, as well as
-interacting with the production pipeline in CentOS CI. A section header
+The instructions below will help you deploy the pipeline. A section header
 may indicate that the section only applies to one of the two workflows.
-
-Another option if one simply wants to test a different COSA image or
-FCOS config or pipeline code (but otherwise maintain the manifest the
-same) is to create a developer pipeline alongside the production one. To
-do this, you *must* use the `devel-up` script. For example:
-
-```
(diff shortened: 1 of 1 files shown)


dustymabe:
still testing this out to make sure there isn't any small tweaks needed before merging

dustymabe:
> still testing this out to make sure there aren't any small tweaks needed before merging

ok I think all is good! lifting WIP

HACKING.md
```diff
@@ -299,3 +310,38 @@ http://simple-httpd-fedora-coreos.$CLUSTER_URL
 ```
 
 (Or simply check the output of `oc get route simple-httpd`).
+
+
+
```
jlebon: Minor: lots of extra blank spaces here

HACKING.md
```diff
@@ -299,3 +310,38 @@ http://simple-httpd-fedora-coreos.$CLUSTER_URL
...
+## Running A Developer Pipeline Alongside an Existing Deployment
+
+If you already know of an existing deployed pipeline you can add some
+namespaced (i.e. DEVEL_PREFIX) resources into the existing project
+and run your own developer pipeline. This is useful if you simply want
+to test a different COSA image or FCOS config or pipeline code, but
+otherwise maintain the manifest the same.
+do this, you *must* use the `devel-up` script. For example:
```
jlebon: Non sequitur

HACKING.md
```diff
@@ -1,25 +1,18 @@
...
-may indicate that the section only applies to one of the two workflows.
+## Setting up a local or production cluster from scratch
 
-Another option if one simply wants to test a different COSA image or
-FCOS config or pipeline code (but otherwise maintain the manifest the
-same) is to create a developer pipeline alongside the production one. To
-do this, you *must* use the `devel-up` script. For example:
+The instructions in this section will cover setting up a local (developer)
+cluster as well as a production cluster pipeline in CentOS CI. If you
+have an existing deployment and just want to set up a parallel
```
jlebon: Maybe instead of `have` --> `have access to` ?

HACKING.md
dustymabe: Yeah. didn't know if the extra spaces would help delineate the next heading as a new **major** section or not. Can you view this in rich format and let me know the best way to do that?

HACKING.md
dustymabe: yeah - copy/paste mistake I think - deleted in next upload.

HACKING.md
dustymabe: either way I deleted the extra space.

jlebon: You can see it here: https://github.com/coreos/fedora-coreos-pipeline/blob/b2c9ec407ecb56b274c6b9377447c69c6e482080/HACKING.md.

AFAIK, markdown doesn't care about the number of spaces. Just using the appropriate number of `#` signs should be good. :+1:

dustymabe:
new update ^^

jlebon:
LGTM, thanks for documenting the new workflow!

PR #74, opened 2019-06-13
pipeline: switch default branch to testing-devel

jlebon:
This will shortly be the default branch in the repo.

### DIFF
--- a/devel-up
+++ b/devel-up
@@ -145,8 +145,21 @@ def params_from_git_refspec(refspec, param_prefix):
 
 def parse_git_refspec(refspec):
     if '@' not in refspec:
-        return (refspec, 'master')
+        return (refspec, get_default_branch(refspec))
     return tuple(refspec.split('@'))
+
+
+def get_default_branch(repo):
+    output = subprocess.check_output(['git', 'ls-remote', '--symref',
+                                      repo, 'HEAD'],
+                                     encoding='utf-8')
+    for line in output.splitlines():
+        if line.startswith('ref: '):
+            ref, symref = line[len('ref: '):].split()
+            if symref != "HEAD":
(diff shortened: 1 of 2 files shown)


jlebon:
Hmm actually, maybe it's cleaner to just make it blank, and handle that in the Groovy code so `git clone` does the right thing.

dustymabe:
LGTM

Maybe we should leave a master branch with a single file README.md that explains the setup of the repo.

dustymabe:
> Maybe we should leave a master branch with a single file README.md that explains the setup of the repo.

that is for the fedora-coreos-config repo

jlebon:
> Hmm actually, maybe it's cleaner to just make it blank, and handle that in the Groovy code so git clone does the right thing.

Hmm nevermind, seems like BuildConfigs default to `master` rather the real default branch. So let's roll with what we have here!

> Maybe we should leave a master branch with a single file README.md that explains the setup of the repo.

Hmm, kinda like in the RHEL dist-git? Wouldn't having a `master` branch at all send mixed signals though? Seems clearer to just not have it at all so anything trying to clone/checkout `master` just fails. But yeah, definitely we should explain the setup in the `README.md` on the default branch.

dustymabe:
> Hmm, kinda like in the RHEL dist-git? Wouldn't having a `master` branch at all send mixed signals though? Seems clearer to just not have it at all so anything trying to clone/checkout `master` just fails. But yeah, definitely we should explain the setup in the `README.md` on the default branch.

ok - i'm cool either way

jlebon:
> But yeah, definitely we should explain the setup in the README.md on the default branch.

https://github.com/coreos/fedora-coreos-config/pull/102

PR #76, opened 2019-06-17
pipeline: add prod, devel, and mechanical pipelines

jlebon:
This adds pipeline buildconfigs for all the current streams (testing,
testing-devel, and bodhi-updates). Rather than having a separate
buildconfig for each stream, we just group them by prod, devel, and
mechanical since the triggering within a particular group is likely
going to be the same.

One note here is that I removed the GitHub webhook from the OpenShift
side, and instead made it part of the Jenkinsfile itself. This is
because we need webhooks to Jenkins anyway to trigger off of
`fedora-coreos-config`, so we might as well use the same mechanism for
`fedora-coreos-pipeline` as well.

Another note: I initially had the `prod` pipeline be solely manually
triggered, though changed that later on to be triggered on pushes to
`fedora-coreos-config`: I think it'd be cool if the only manual step
required for a release was simply to push a commit to `testing`. Anyway,
we can refine that later if needed (and all the other triggers as well).

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -171,7 +171,8 @@ oc create secret generic fcos-builds-bot-aws-config --from-file=config=/path/to/
 oc new-app --template=jenkins-persistent \
     --param=NAMESPACE=fedora-coreos \
     --param=MEMORY_LIMIT=2Gi \
-    --param=VOLUME_CAPACITY=2Gi
+    --param=VOLUME_CAPACITY=2Gi \
+    --param=JENKINS_IMAGE_STREAM_TAG=jenkins:2
 ```
 
 Notice the `NAMESPACE` parameter. This makes the Jenkins master use the
@@ -180,17 +181,13 @@ reason we create the app first is that otherwise OpenShift will
 automatically instantiate Jenkins with default parameters when creating
 the Jenkins pipeline).
 
-Next, let's update the input ImageStreamTag for our Jenkins deployment
-to match latest OpenShift (`jenkins:2`). Some older versions of the
-template in OpenShift uses `jenkins:latest`. This will no longer be
-needed once we are running on a newer version of OpenShift than 3.6 in
+The `jenkins:2` parameter is to match the tag name in the latest
+OpenShift. Some older versions of the template in OpenShift uses
+`jenkins:latest`. This will no longer be needed once we are running on a
+newer version of OpenShift than 3.6 in
 CentOS CI. See [#32](https://github.com/coreos/fedora-coreos-pipeline/pull/32)
 and [#70](https://github.com/coreos/fedora-coreos-pipeline/pull/70)
-for context on why this may be needed):
-
-```
-oc patch dc/jenkins -p '{"spec":{"triggers":[{"imageChangeParams":{"automatic":true,"containerNames":["jenkins"],"from":{"kind":"ImageStreamTag","name":"jenkins:2"}},"type":"ImageChange"},{"type":"ConfigChange"}]}}'
-```
+for more context).
 
 ### Create the pipeline from the template
 
--- a/HACKING.md
+++ b/HACKING.md
@@ -203,11 +203,11 @@ oc new-app --file=manifests/pipeline.yaml
 If working on a local cluster you will want to override the defaults
 for some parameters:
 
-- `DEVEL_PREFIX` - **REQUIRED**
+- `DEVELOPER_PREFIX` - **REQUIRED**
     - The prefix to prepend to created resources.
     - This is recommended to be your username followed by a dash.
       i.e. `jlebon-`. The pipeline will simply not run without a
-      `DEVEL_PREFIX` set if it is not running in prod mode.
+      `DEVELOPER_PREFIX` set if it is not running in prod mode.
 - `PIPELINE_REPO_URL`
     - Git source URI for pipeline Jenkinsfile
 - `PIPELINE_REPO_REF`
@@ -226,7 +226,7 @@ for some parameters:
 
 ```
 oc new-app --file=manifests/pipeline.yaml \
-    --param=DEVEL_PREFIX=$(whoami)- \
+    --param=DEVELOPER_PREFIX=$(whoami)- \
     --param=PIPELINE_REPO_URL=https://github.com/jlebon/fedora-coreos-pipeline \
     --param=PIPELINE_REPO_REF=my-feature-branch \
     --param=S3_BUCKET=jlebon-fcos \
@@ -313,7 +313,7 @@ http://simple-httpd-fedora-coreos.$CLUSTER_URL
 ## Running A Developer Pipeline Alongside an Existing Deployment
 
 If you already know of an existing deployed pipeline you can add some
-namespaced (i.e. DEVEL_PREFIX) resources into the existing project
+namespaced (i.e. DEVELOPER_PREFIX) resources into the existing project
 and run your own developer pipeline. This is useful if you simply want
 to test a different COSA image or FCOS config or pipeline code, but
 otherwise maintain the manifest the same.
@@ -333,7 +333,7 @@ yourself using something like:
         --bucket jlebon-fcos
 ```
 
-You can provide `--prefix` to devel up but if you don't it will
+You can provide `--prefix` to `devel-up` but if you don't it will
 default to your local username.
 
 Any value you don't pass to devel-up will be reset to its default
--- a/HACKING.md
+++ b/HACKING.md
@@ -273,24 +273,23 @@ oc start-build fedora-coreos-pipeline
 
 Use the web interface to view logs from builds.
 
-### [PROD] Update the "secret" token values in the webhooks to be unique
+### [PROD] Update the "secret" token value in the webhook to be unique
 
 ```
-oc set triggers bc/fedora-coreos-pipeline --from-github
-oc set triggers bc/fedora-coreos-pipeline --from-webhook
+oc set triggers bc/fedora-coreos-pipeline-mechanical --from-webhook
 ```
 
 ### [PROD] Set up webhooks/automation
 
-Grab the URLs of the webhooks from `oc describe` and set up webhook
-in github.
-
-- `oc describe bc/fedora-coreos-pipeline` and grab the `Webhook GitHub` URL
-- From the GitHub web console for the configs repository.
-- Select Add Webhook from Settings → Webhooks & Services.
-- Paste the webook URL output into the Payload URL field.
+- From the GitHub Settings tab for `fedora-coreos-config`, go to the
+  "Webhooks" panel
+- Click "Add webhook"
+- In the address, type `$JENKINS_URL/github-webhook/`. So e.g.:
+  https://jenkins-fedora-coreos-devel.apps.ci.centos.org/github-webhook/
 - Change the Content Type from GitHub’s default `application/x-www-form-urlencoded` to `application/json`.
-- Click Add webhook.
+- Click "Add webhook"
+
+Repeat these steps for the `fedora-coreos-pipeline` repo.
 
 ### [OPTIONAL] Set up simple-httpd
 
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,18 +1,21 @@
+## Setting up the pipeline
 
-## Setting up a local or production cluster from scratch
+There are multiple ways to set up the pipeline. There is a single
+official prod mode, which can only be instantiated in CentOS CI under
+the `fedora-coreos` namespace. However, one can set up everything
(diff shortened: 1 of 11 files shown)


jlebon:
OK, dropped WIP on this! It's looking pretty good now I think and works well. I'm sure there's some knobs in there e.g. around triggering that'll need more adjustments. But we can do those as follow-ups as we gain more experience.

jlebon:
Requires #78.

dustymabe:
I have a nit on the terminology we are using. In this PR we are naming/referring to things as `prod`, `devel` and `mechanical`. We also have a concept in this repo of *developer (could be confused with devel) pipelines*.

I suggest we slightly change our terminology and naming such that it is now `stable`, `devel` and `mechanical`. All of these are classified as `prod`. Anything running in centos CI in the fedora-coreos namespace should be `prod`. We should not allow anything else to be run in that namespace. Anything running anywhere else is a developer pipeline essentially.

So we have the *prod* `stable` streams. *prod* `devel` streams, and *prod* `mechanical` streams. You could individually set up a *devel* `stable` stream, *devel* `devel` stream, etc.

jlebon:
Yeah, I've thought about the devel overloading happening too while working on this.

Hmm, how about we just spell it out, i.e. `development`, instead of just `devel` to clear this up? So, `prod`, `development`, and `mechanical`. This actually matches the terminology used in https://github.com/coreos/fedora-coreos-tracker/blob/master/stream-tooling.md as well.

jlebon:
> Anything running in centos CI in the fedora-coreos namespace should be prod. We should not allow anything else to be run in that namespace. Anything running anywhere else is a developer pipeline essentially.

Agreed. This is enforced today by explicitly checking for the job name itself in the pipeline. And just plain error out if it doesn't match.

dustymabe:
> Yeah, I've thought about the devel overloading happening too while working on this.
> 
> Hmm, how about we just spell it out, i.e. `development`, instead of just `devel` to clear this up? So, `prod`, `development`, and `mechanical`. 

Yeah I think that may help. 

> This actually matches the terminology used in https://github.com/coreos/fedora-coreos-tracker/blob/master/stream-tooling.md as well.

Yeah looking at that document makes me rethink this a bit. Thanks for linking that. 


I think the problem here is we need a clear term to delineate between official fcos streams and a stream that is a side build for a developer.. Maybe we can call official streams official (i.e. change our `prod` var in Jenkinsfiles to `official` , and we could call developer pipelines `hacker` pipelines.

jlebon:
> Maybe we can call official streams official (i.e. change our `prod` var in Jenkinsfiles to `official`, and we could call developer pipelines `hacker` pipelines.

Yup sure, WFM.

jlebon:
Rebased this on top of #80.

I added a commit which updates the terminology we use in the various places. See the commit message for details. Re-adding WIP on this since I haven't tested it yet.

Jenkinsfile.mechanical
```diff
@@ -0,0 +1,33 @@
...
+    ])
+])
+
+node {
+    change = checkout(
+        [$class: 'GitSCM',
+         userRemoteConfigs: [
+            [url: 'https://github.com/coreos/fedora-coreos-config']
+         ],
+         branches: streams.as_branches(streams.prod)
```
dustymabe: streams.production

Jenkinsfile
```diff
@@ -100,6 +106,10 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
                 utils.shwrap("""
                 coreos-assembler buildprep s3://${s3_builddir}
                 """)
+            } else if (!prod && utils.path_exists(developer_builddir)) {
```
dustymabe: `!official`

Jenkinsfile
```diff
@@ -176,14 +190,14 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
...
               // In devel mode without an S3 server, just archive into the PVC
               // itself. Otherwise there'd be no other way to retrieve the
               // artifacts. But note we only keep one build at a time.
               utils.shwrap("""
-              rm -rf /srv/devel/${devel_prefix}/build
-              mkdir -p /srv/devel/${devel_prefix}/build
-              cp -a \$(realpath builds/latest) /srv/devel/${devel_prefix}/build
+              rm -rf ${developer_builddir}
+              mkdir -p ${developer_builddir}
+              cp -aT builds ${developer_builddir}
```
dustymabe: one more `if (prod)` at the bottom of `stage('Archive')`

devel-up
```diff
@@ -92,7 +92,7 @@ def get_username():
 
 
 def process_template(args):
-    params = [f'DEVEL_PREFIX={args.prefix}']
+    params = [f'DEVELOPER_PREFIX={args.prefix}']
```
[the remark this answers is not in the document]
jlebon: Done!

Jenkinsfile.mechanical
jlebon: Heh, yeah caught those while testing this morning.

dustymabe:
after a few fixes I hit this from within my jenkins job:

```
[workspace] Running shell script

+ oc start-build fedora-coreos-pipeline -e STREAM=bodhi-updates

Error from server (Forbidden): buildconfigs.build.openshift.io "fedora-coreos-pipeline" is forbidden: buildconfigs.build.openshift.io "fedora-coreos-pipeline" not found

script returned exit code 1
```

Jenkinsfile.mechanical
```diff
@@ -0,0 +1,33 @@
...
+    checkout scm
+    utils = load("utils.groovy")
+    streams = load("streams.groovy")
+}
+
+properties([
+    pipelineTriggers(streams.get_push_trigger() + [
+        // also run every 30m: we could increase the interval more if we start
+        // triggering on the relevant fedmsgs.
+        cron("H/30 * * * *")
```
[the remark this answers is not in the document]
jlebon: Developer pipelines don't have any cron jobs right now. Note the only buildconfig that actually gets created by `devel-up` right now is still just `${developer_prefix}fedora-coreos-pipeline`. And that one was stripped of all triggers.

jlebon:
> after a few fixes I hit this from within my jenkins job:

Right yeah, basically right now, the "wrapper" pipelines don't support developer mode (i.e. it doesn't have a `${developer_prefix}` in its name). To test all this, I just created them all without a prefix in the devel namespace. The main pipeline still failed as expected since it wasn't in the right namespace, but for testing the triggering itself that's fine (which is the only thing that really needs testing in the wrapper pipelines).

I guess we could make those devel aware as well? Though then I think we'd still want `./devel-up` to default to creating just the pipeline.

jlebon:
Dropping WIP on this. Tested working well!

Jenkinsfile.mechanical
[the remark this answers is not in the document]
dustymabe: I think we're glossing over the fact that if I want to run a developer pipeline in a brand new deployment (new project in openshift in any openshift instance) you can't start with `devel-up`. You start with `oc new-app --file=manifests/pipeline.yaml --param=DEVELOPER_PREFIX=$(whoami)-`. In this case you get **everything**, which was OK in the past because we wanted everything (i.e. we used it all for our developer pipelines usecase), but now it's not the same.

dustymabe:
> I guess we could make those devel aware as well? 

I think so?

> Though then I think we'd still want `./devel-up` to default to creating just the pipeline.

Some more thoughts in https://github.com/coreos/fedora-coreos-pipeline/pull/76#discussion_r295844134

Jenkinsfile.mechanical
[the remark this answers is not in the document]
jlebon: Yeah, one base issue here is that OpenShift templates don't have conditionals (I'm not sure if Helm or something else would be better here). What we can do is split `pipeline.yaml` into 3:
- `jenkins.yaml` --> S2I + worker base + PVC (and we can split that one out as well when we're ready to make it optional)
- `pipeline.yaml` --> coreos-assembler IS + fedora-coreos-pipeline BC
- `streams.yaml` --> fedora-coreos-pipeline-{production,developer,mechanical} BCs

So then, the workflow would be `oc new-app --file=manifests/jenkins.yaml` and then straight to `./devel-up`, which instantiates `pipeline.yaml`. We could then also have e.g. a `--streams` switch if a developer wants to instantiate streams BCs as well.

jlebon:
OK, updated this now! `./devel-up` has now evolved into `./deploy`! See commit message for details.

jlebon:
>> I guess we could make those devel aware as well?
>
> I think so?

jlebon:
>> I guess we could make those devel aware as well?
>
> I think so?

One note about this: I still didn't make the stream pipelines prefixable. Let's do that as a follow-up? I don't think it's something we need the ability to do right away (and of course, one can always do `./deploy --all` in their separate cluster if they want to test them).

manifests/pipeline.yaml
```diff
@@ -52,6 +52,8 @@ objects:
     kind: ImageStream
     metadata:
       name: jenkins-s2i
+      annotations:
+        coreos.com/deploy-default: true
```
dustymabe: these annotations are causing me some grief on first startup:

```
Updating:                                                                                
Error from server (BadRequest): error when creating "STDIN": ImageStream in version "v1" cannot be handled as a ImageStream: v1.ImageStream.ObjectMeta: v1.ObjectMeta.Annotations: ReadString: expects " or n, but found t, error found
 in #10 byte of ...|default":true},"labe|..., bigger context ...|ata":{"annotations":{"coreos.com/deploy-default":true},"labels":{"app":"fedora-coreos","template":"f|...
Traceback (most recent call last):           
  File "./deploy", line 235, in <module>      
    sys.exit(main())                         
  File "./deploy", line 38, in main                                                        
    update_resources(args, resources)        
  File "./deploy", line 164, in update_resources                                                                                                                                                                                           check=True, stdout=subprocess.PIPE)      
  File "/usr/lib64/python3.7/subprocess.py", line 487, in run
    output=stdout, stderr=stderr)            
subprocess.CalledProcessError: Command '['oc', 'create', '--filename', '-']' returned non-zero exit status 1.

```

If I delete all the newly added `coreos.com/deploy-default: true` annotations and all `--all` to my `deploy` command then it succeeds.

jlebon: Hmm weird, I didn't hit this on my local server. But yeah, looks like it's complaining about string vs bool.

dustymabe:
re: `HACKING.md: simply Jenkins template instantiation`

Thanks for that commit. I didn't know it was a parameter otherwise I would have used it before :)

One nit: `simplify Jenkins template instantiation`

jlebon:
Fixes! :arrow_up: 

Tested in a fresh namespace.

HACKING.md
```diff
@@ -180,70 +184,58 @@ reason we create the app first is that otherwise OpenShift will
...
-    --param=PVC_SIZE=20Gi
+./deploy --update \
+    --pipeline https://github.com/jlebon/fedora-coreos-pipeline     \
+    --config https://github.com/jlebon/fedora-coreos-config@feature \
+    --cosa-img quay.io/jlebon/coreos-assembler:random-tag           \
+    --bucket jlebon-fcos
 ```
 
-This template creates:
+See `./deploy --help` for more information.
```
dustymabe: I kind of liked having all of the parameters explained out here like we were doing before when we listed out all of the parameters of the template. Are you opposed to adding that stanza back but just convert it to reference arguments to deploy rather than template parameters?

dustymabe:
> One note about this: I still didn't make the stream pipelines prefixable. Let's do that as a follow-up?

sounds good

Mostly LGTM - one comment: https://github.com/coreos/fedora-coreos-pipeline/pull/76#discussion_r296285156

HACKING.md
jlebon: Done! :arrow_down:

HACKING.md
```diff
@@ -180,70 +184,73 @@ reason we create the app first is that otherwise OpenShift will
...
-- `S3_BUCKET`
-    - AWS S3 bucket in which to store builds (or blank for none)
+- `--prefix`
+    - The prefix to prepend to created developer-specific resources. By
+      default, this will be your username, but you can provide a
+      specific prefix as well.
+- `--pipeline`
+    - Git source URI for pipeline Jenkinsfile.
+- `--config`
+    - Git source URI for FCOS config.
```
dustymabe: since before we had the `_REF` parameters and now we don't  you might want to mention the `@ref` formatting for these two

jlebon: Fixed!

jlebon:
Awesome, thanks for all the reviews! Let's get this thing set up for real now! :)

PR #77, opened 2019-06-17
Prep patches for multi-streams

jlebon:
Split out/reworked from #76.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,8 +1,9 @@
-def pod, utils, prod, prod_jenkins, devel_prefix, src_config_url, src_config_ref, s3_bucket
+def utils, streams, prod, prod_jenkins, devel_prefix, src_config_url, src_config_ref, s3_bucket
 node {
     checkout scm
-    pod = readFile(file: "manifests/pod.yaml")
     utils = load("utils.groovy")
+    streams = load("streams.groovy")
+    pod = readFile(file: "manifests/pod.yaml")
 
     // just autodetect if we're in prod or not
     prod_jenkins = (env.JENKINS_URL == 'https://jenkins-fedora-coreos.apps.ci.centos.org/')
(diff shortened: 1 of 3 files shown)


dustymabe:
mostly LGTM - still need to process it. Will try a run through with #78

Jenkinsfile
```diff
@@ -34,10 +35,7 @@ properties([
     pipelineTriggers(prod ? [cron("H/30 * * * *")] : []),
     parameters([
       choice(name: 'STREAM',
-             // XXX: Just pretend we're the testing stream for now... in
-             // reality, we're closer to what "bodhi-updates" will be. Though the
-             // testing stream is the main stream.
-             choices: ['testing' /*, 'stable', 'testing-devel', 'bodhi-updates', etc... */ ],
+             choices: (streams.prod + streams.devel + streams.mechanical),
```
dustymabe: if I just do `oc start-build dustymabe-fedora-coreos-pipeline` which choice is used?

jlebon: Unless you overrode it through `./devel-up`, you'd get `testing` (the first one). We can change that so it defaults to `testing-devel`. But really, if you care which specific one gets built (i.e. you're testing something sensitive to the FCOS config itself), you should use `./devel-up --config`.

dustymabe: yeah let's switch it so that `testing-devel` is first (i.e. default behavior is the default branch in fcos-configs repo). I think that is a sane default we can live with.

jlebon: Done! :arrow_down:

dustymabe:
one final thing that is a bit subtle when working with all of this is that sometimes we refer to things as streams and sometimes we override those things with ref names. The key here is that the stream name and the ref name for the branches in the fcos configs repo correspond 1:1. This is something that should be obvious to me (and it mostly is), but I still find myself tripping up on it.

PR #78, opened 2019-06-17
pipeline: stop using /srv as workdir

jlebon:
I think we should stop using `/srv` as a workdir entirely and just
always build in the workspace. The core issue here is that (1) we want
to be able to have concurrent builds, and (2) a cosa workdir can't be
easily shared today. This also simplifies the devel vs prod logic quite
a bit since it had some funky conditionals around this.

So then, how can developers without S3 creds actually *access* built
artifacts? We simply archive them as part of the build. This is in line
also with #31, where we'll probably be archiving things anyway.

Finally, how *can* we use the PVC as cache in a safe way shareable
across all the streams? I see two options offhand:
1. as a local RPM mirror: add flags and logic to `cosa fetch` to read &
   write RPMs in `/srv`, hold a lock to regen metadata
2. as a pkgcache repo: similarly to the above, but also doing the
   import, so it's just a pkgcache repo; this would probably require
   teaching rpm-ostree about this, or `cosa fetch` could just blindly
   import every ref

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -158,8 +158,6 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             utils.shwrap("""
             if [ \$(du cache/cache.qcow2 | cut -f1) -gt \$((1024*1024*8)) ]; then
                 rm -vf \$(realpath cache/cache.qcow2)
-                qemu-img create -f qcow2 cache/cache.qcow2 10G
-                LIBGUESTFS_BACKEND=direct virt-format --filesystem=xfs -a cache/cache.qcow2
             fi
             """)
         }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -55,15 +55,6 @@ if (prod) {
 podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultContainer: 'jnlp') {
     node('coreos-assembler') { container('coreos-assembler') {
 
-        // Only use the PVC for prod caching. For devel pipelines, we just
-        // always refetch from scratch: we don't want to allocate cached data
-        // for pipelines which may only run once.
-        if (prod) {
-          utils.workdir = "/srv"
-        } else {
-          utils.workdir = env.WORKSPACE
-        }
-
         // this is defined IFF we *should* and we *can* upload to S3
         def s3_builddir
 
@@ -81,13 +72,6 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
           }
         }
 
-        // Special case for devel pipelines not running in our project and not
-        // uploading to S3; in that case, the only way to make the builds
-        // accessible at all is to have them in the PVC.
-        if (!prod && !prod_jenkins && !s3_builddir) {
-            utils.workdir = "/srv"
-        }
-
         stage('Init') {
 
             def ref = params.STREAM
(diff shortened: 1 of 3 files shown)


jlebon:
Requires #77.

Jenkinsfile
```diff
@@ -187,6 +159,11 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
               utils.shwrap("""
               coreos-assembler buildupload s3 --acl=public-read ${s3_builddir}
               """)
+            } else if (!prod) {
+              // In devel mode without an S3 server, just archive in Jenkins
+              // itself. Otherwise there'd be no other way to retrieve the
+              // artifacts.
+              archiveArtifacts('builds/latest/*')
```
dustymabe: so this will just show up in the jenkins web interface ? Would we need to make the size of the PVC much larger for jenkins in this case ?

jlebon: > so this will just show up in the jenkins web interface ?

Yup, exactly!

> Would we need to make the size of the PVC much larger for jenkins in this case ?

Yeah, at least for the Jenkins in the devel namespace? (Though also in the prod case in preparations for #31). I can adjust `HACKING.md` to reflect this.

I guess another approach is to store builds in the PVC only in the devel case (but still build in the workspace as done in this patch). E.g. something like `/srv/builds/$devel_prefix`?

dustymabe: > I guess another approach is to store builds in the PVC only in the devel case (but still build in the workspace as done in this patch). E.g. something like `/srv/builds/$devel_prefix`?

I think I like that better, but i'm not sure. It's a lot easier for me to think about freeing up space in the PVC for coreos-assembler-claim2 than it is to conceptualize freeing up space in the jenkins PVC. Then again if we can maybe we should just get rid of the `coreos-assembler-claim2` for the dev case and only have one storage requirement.

jlebon:
OK, updated this now to use the PVC as cache!

Haven't made it optional... let's do that as a follow-up? (Though on that subject, we should probably encourage creating it in the local dev case, unless you have a really good connection.)

jlebon:
Once this is merged, I can cleanup the workdir-related dirs in the prod and devel PVCs for space.

dustymabe:
> OK, updated this now to use the PVC as cache!

sweet - will check it out

> 
> Haven't made it optional... let's do that as a follow-up? (Though on that subject, we should probably encourage creating it in the local dev case, unless you have a really good connection.)

yeah, encouraged, but optional

In case you aren't doing local dev but actually using a different openshift cluster you don't control then sometimes PVCs can actually be a blocker. So for example, in the case of it being optional, I could run an entire pipeline with a jenkins ephemeral template and no PVC. This would just be POC/testing of course as you'd lose data eventually.

utils.groovy
```diff
@@ -1,25 +1,20 @@
-workdir = env.WORKSPACE
```
dustymabe: so now that we aren't setting this any longer we just always default to `env.WORKSPACE` anyway?

jlebon: Yup, exactly!

Jenkinsfile
```diff
@@ -81,28 +72,26 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
...
-            # just always restart from scratch in case it's a devel pipeline
-            # and it changed source url or ref; this info also makes it into
-            # the build metadata through cosa reading the origin remote
-            rm -rf src/config
+            // for now, just use the PVC to keep cache.qcow2 in a stream-specific dir
+            def cache_img
+            if (prod) {
+                cache_img = "/srv/prod/${params.STREAM}/cache.qcow2"
+            } else {
+                cache_img = "/srv/devel/${devel_prefix}/cache.qcow2"
```
dustymabe: wondering if we can make this `/srv/devel/${devel_prefix}/${params.STREAM}/cache/cache.qcow2`

and then below store the builds in `/srv/devel/${devel_prefix}/${params.STREAM}/builds/`

jlebon: Hmm, though in devel mode, the "stream" doesn't mean anything since you provide the ref to check out. See also https://github.com/coreos/fedora-coreos-pipeline/blob/e50e6780972c3c29b6c3fb529136fc81972da174/Jenkinsfile#L75. Does that make sense?

dustymabe: yeah - it was just a gut reaction when I started digging in the PVC. seemed like the stream (ref) should be in the path

dustymabe:
this works - we can do other things in follow up PRs - @jlebon I'll let you push the button for merge since it might need some coordination with jenkies in prod

jlebon:
Thanks for reviewing!

dustymabe:
so I've run a few builds now. It always does a build (never detects it doesn't need to) and always makes a build version of `30`.

dustymabe:
The good news is that it didn't re-download packages every time because it was properly using the cache 🤣

jlebon:
Ahh yeah, we can tweak things so it `buildprep`s from the PVC as well for devel builds.

jlebon:
#80

PR #79, opened 2019-06-18
HACKING.md: drop references to NAMESPACE parameter

jlebon:
This is no longer a parameter since #64.


dustymabe:
> This is no longer a parameter since #64.

Is that true? `NAMESPACE` is a param in the `jenkins-persistent` template within openshift

```
$ oc get template/jenkins-persistent -o yaml --namespace openshift | grep NAMESPACE
          namespace: ${NAMESPACE}
  name: NAMESPACE

```

jlebon:
Ohh whoops, mixed up my templates. Yes, you're correct!

PR #81, opened 2019-06-21
deploy: fix referenced before assignment bug

dustymabe:
otherwise in `--official` mode you see:

```
Traceback (most recent call last):
  File "./deploy", line 235, in <module>
    sys.exit(main())
  File "./deploy", line 35, in main
    resources = process_template(args)
  File "./deploy", line 134, in process_template
    params += params_from_git_refspec(args.pipeline, 'PIPELINE_REPO')
UnboundLocalError: local variable 'params' referenced before assignment
```

### DIFF
--- a/deploy
+++ b/deploy
@@ -128,8 +128,9 @@ def get_username():
 
 
 def process_template(args):
+    params = []
     if not args.official:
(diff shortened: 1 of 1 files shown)


jlebon:
Heh, literally was about to push this.
Thanks!

PR #83, opened 2019-06-21
streams: use `--wait` so we wait and report errors if downstream fails

jlebon:
I actually wanted to have "blocking" semantics from the beginning, but
abandoned on that when switching away from the `build` step. And I
didn't think to try `oc start-build --wait`. There's `--follow` too, but
it doesn't work with Jenkins pipelines.

We can probably improve on this in the future by capturing the build
name, and printing a link to the Jenkins console for that build.

### DIFF
--- a/streams.groovy
+++ b/streams.groovy
@@ -36,9 +36,10 @@ def triggered_by_push() {
 
 // Starts a stream build.
 def build_stream(stream) {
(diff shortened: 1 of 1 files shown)


dustymabe:
my build hasn't completed yet but seems to be working for me..

PR #84, opened 2019-06-24
pipeline: stop rsync'ing to artifact server completely

jlebon:
Now that we have S3 uploads set up, let's stop the "compat" rsync'ing to
the artifact server. Instead, I propose we leave a single file there
which points people to https://builds.coreos.fedoraproject.org/browser.

Once we're officially in preview mode, we can evaluate whether we should
make the above less accessible, or perhaps simply dropping the prod
streams from the dropdown.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -317,9 +317,9 @@ Repeat these steps for the `fedora-coreos-pipeline` repo.
 ### [OPTIONAL] Set up simple-httpd
 
 When hacking locally, it might be useful to look at the contents of the
-PV to see the builds since one isn't rsync'ing to an artifact server.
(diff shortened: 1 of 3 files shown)


dustymabe:
LGTM - should we send an email about this change?

jlebon:
> LGTM - should we send an email about this change?

Yeah, I can do that.

jlebon:
http://artifacts.ci.centos.org/fedora-coreos/prod/builds/latest/README.txt

jlebon:
https://lists.fedoraproject.org/archives/list/coreos@lists.fedoraproject.org/thread/EKHDKT6U3BCSKRFBVCBAZNHEAM52XMBB/

PR #86, opened 2019-06-28
HACKING.md: add note on running oc rootless

rfairley:
Add a note on running oc rootless while Docker by default requires
root, which tripped me up at first coming back to this.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -75,6 +75,10 @@ And now, bring up a v3.6.1 cluster (again, to match CentOS CI):
 oc cluster up --version v3.6.1
 ```
 
+Note that `oc cluster up/down` will require running as root, to
+communicate with the docker daemon. However, the other steps in
+this guide can (and should) be run as non-root.
+
 To have persistent configs and data, I would recommend specifying the
 dirs to use.
 


dustymabe:
adding the user to the docker group gives them root on the system so not sure if that's the best advice.

jlebon:
Note you only need to run `oc cluster up` as root. Everything else can (and should) be done as non-root. Not super opposed to this, though yeah it's generally not recommended (see https://www.projectatomic.io/blog/2015/08/why-we-dont-let-non-root-users-run-docker-in-centos-fedora-or-rhel/).

Maybe let's just document that `oc cluster up/down` should be run as root?

jlebon:
This is the bash alias I use BTW:

```bash
ocup() {
    local v=$1; shift
    local existing=

    if [ -z "$v" ]; then
        echo "Missing version!"
        return 1
    fi

    # just concentrate everything in /srv/oc to make it easier to wipe off
    local d=/srv/ocup/oc-$v
    if [ -d $d ]; then
        existing="--use-existing-config"
    else
        sudo mkdir -p $d/{config,data,pv}
    fi

    sudo oc cluster up \
        --host-config-dir=$d/config \
        --host-data-dir=$d/data \
        --host-pv-dir=$d/pv $existing \
        --version "$v" "$@"

    # transfer creds over from root dir so we can log in as system:admin
    if [ $UID -ne 0 ] && [ -n "$HOME" ]; then
        rm -rf $HOME/.kube
        sudo cp -r /root/.kube $HOME
        sudo chown -R $UID:$UID $HOME/.kube
    fi

    oc login -u developer
}
```

And then I do e.g. `ocup v3.6.1`,

Note the `sudo cp -r /root/.kube $HOME` as a consequence of standing up the cluster as root. (Also note this will wipe any other cached k8s creds).

rfairley:
@dustymabe, @jlebon Agreed - makes sense to document running as root just for the commands that need it. Updated!

Thank you for linking the bash alias @jlebon, trying it on my system! (F30 Silverblue)

rfairley:
Just did a minor last-minute push to adjust wording.

PR #87, opened 2019-07-09
pipeline: switch to xz compression

jlebon:
As discussed in https://github.com/coreos/coreos-assembler/issues/445.


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/596

jlebon:
OK, this one should be ready to go!

zonggen:
:+1:

PR #89, opened 2019-07-10
Jenkinsfile: add step for building release metadata

arithx:
Adds a step which downloads the `fedora-coreos-releng-automation` repo
and runs the `coreos-meta-translator` against the current build to
produce `release.json` files.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -171,7 +171,7 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             // which will generate a release.json from the meta.json files
             utils.shwrap("""
             git clone https://github.com/coreos/fedora-coreos-releng-automation /var/tmp/fcos-releng
-            /var/tmp/fcos-releng/coreos-meta-translator/trans.py --workdir .
+            /var/tmp/fcos-releng/coreos-meta-translator/trans.py --workdir . || true
             """)
         }
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -166,6 +166,15 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             }
         }
 
+        stage('Build Release Metadata') {
+            // Run the coreos-meta-translator against the most recent build,
+            // which will generate a release.json from the meta.json files
+            utils.shwrap("""
+            git clone https://github.com/coreos/fedora-coreos-releng-automation /var/tmp/fcos-releng
+            /var/tmp/fcos-releng/coreos-meta-translator/trans.py --workdir .
+            """)
+        }
+
         stage('Prune Cache') {
             // If the cache img is larger than e.g. 8G, then nuke it. Otherwise
             // it'll just keep growing and we'll hit ENOSPC. Use realpath since


arithx:
Untested (looking into adding a `plume` command in `cosa` and wrapping that in a job to do both at once)

Jenkinsfile
```diff
@@ -166,6 +166,16 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             }
         }
 
+        stage('Build Release Metadata') {
+            // Run the coreos-meta-translator against the most recent build,
+            // which will generate a release.json from the meta.json files
+            utils.shwrap("""
+            git clone https://github.com/coreos/fedora-coreos-releng-automation
```
jlebon: Minor: let's put this somewhere under e.g. `/var/tmp/` so as not to pollute the workdir? Doesn't matter now, though keeping a clear mental model of `$PWD = cosa workdir` is helpful.

jlebon:
> Untested (looking into adding a plume command in cosa and wrapping that in a job to do both at once)

You mean the new cosa command would run `coreos-meta-translator`?

Did you mean to hold this until it's part of cosa, or getting this in provisionally? I'd vote for the latter.

arithx:
> You mean the new cosa command would run coreos-meta-translator?

No adding the building of `plume` in the cosa container as well as a wrapper for it.

I was planning on adding that piece (and spinning a PR to use the new `plume` cosa command in the `fedora-coreos-pipeline` so I could test both this PR as well as the one for plume at the same time).

For this I'm fine with having it outside of cosa for now.

jlebon:
OK gotcha. Let's hold it for now then.

arithx:
Blocked by https://github.com/coreos/coreos-assembler/pull/580 (the meta translator dies with the current `cosa` builds.json layout)

jlebon:
https://github.com/coreos/coreos-assembler/pull/580 has merged, but our bucket layout is still on the old one. Chatted with @bgilbert and once https://github.com/coreos/fedora-coreos-config/pull/112 is merged which has the new versioning scheme, we'll just `aws s3 mv s3://...builds.json{,.old}` instead of migrating the old builds. We will later delete these builds.

IOW, this is now soft blocked on https://github.com/coreos/fedora-coreos-config/pull/112.

If you're targeting a new bucket though, now that https://github.com/coreos/coreos-assembler/pull/580 is merged, you should be able to use the new layout in a developer pipeline (once quay finishes rebuilding the container).

jlebon:
OK https://github.com/coreos/fedora-coreos-config/pull/112 is merged now. I completed the move for `bodhi-updates`. Triggered a new build to sanity check the layout works as expected before switching `testing-devel` and `testing` over as well.

arithx:
This works for me in the local cluster.

jlebon:
OK, new layout working fine, though I'm still working on adapting the build browser. I'd like to merge this to test it on `bodhi-updates`, but it'll fail on `testing-devel` until we do the move there too (but I'd like to finish adapting the browser first). Hacky, but for the short-term, can you add a `|| true` for now? We'll revert it later.

arithx:
@jlebon added a `HACK` commit referencing this comment with the `|| true` for ease of revert later.

jlebon:
Awesome, thanks!

jlebon:
OK, browser support added, and `testing` and `testing-devel` moved over to new layout.
Reverting the `|| true` in https://github.com/coreos/fedora-coreos-pipeline/pull/96.

PR #90, opened 2019-07-11
Jenkinsfile: add Publish step

arithx:
Adds a publish step for official builds that runs plume against the
version/stream. This will modify all object ACLs for the given build to
be public-read and create or modify the releases.json metadata index.

Requires coreos/coreos-assembler#606


arithx:
`plume` points directly at the prod s3 bucket. Provided the Jenkins steps themselves doesn't have any glaring errors plume itself has already been tested.

PR #93, opened 2019-07-11
production: disable all triggers for now

jlebon:
In the short-term, we'll be triggering production (i.e. `testing`)
builds manually. Other streams are unaffected (`testing-devel` and
`bodhi-updates`).

This is so that we can manually drive versioning for the time being
until we have something automated in place.

### DIFF
--- a/Jenkinsfile.production
+++ b/Jenkinsfile.production
@@ -6,7 +6,8 @@ node {
 }
 
 properties([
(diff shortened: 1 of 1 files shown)


jlebon:
/cc @bgilbert 

This is more of a formality. I've already disabled the job in Jenkins.

PR #94, opened 2019-07-12
pipeline: add VERSION job parameter

jlebon:
Add a temporary `VERSION` parameter that will allow us to manually input
the version we want for `testing` builds. This will be used in the
short-term for releasing FCOS preview.


jlebon:
Requires https://github.com/coreos/coreos-assembler/pull/611.

jlebon:
Thanks, added comment and link!

PR #96, opened 2019-07-12
Jenkinsfile: drop `|| true` hack

jlebon:
We shouldn't need this anymore. All streams have now migrated to the new
layout.


arithx:
Should this just be a revert commit of 7eb04908c451bb0c7be071401d8c05172ba179ea?

jlebon:
OK, updated! :up:

PR #100, opened 2019-07-12
Jenkinsfile: directly reference aws creds location

arithx:
We already run `test -e` against this hard-coded path, switch the plume
call to using it directly as well.


jlebon:
So I think what's happening here is that `AWS_CONFIG_FILE` isn't in the environment of the Groovy interpreter. Which I guess could makes sense since that's running in Jenkins itself, and not literally inside the container, on which the env var is defined through the OpenShift manifest. My guess is that switching to `\${AWS_CONFIG_FILE}` so bash does the substitution might also have fixed it.

Anyway, good with this too!

PR #102, opened 2019-07-15
Jenkinsfile: start uploading AMIs

jlebon:
There are a few undesirable things in here, but it will do for now. E.g.
we're not replicating across multiple regions, nor are we uploading
separate vmdks.

There's a lot of stuff around this that the RHCOS team has internally.
We should look to upstreaming it into cosa (or replacing it with
existing functionality in mantle and extending as necessary), so that it
can be shared with FCOS.

For now, we're just patching things together so we at least have
`us-east-1` for preview.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -183,6 +183,8 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
                     // also publish vmdks, we could make this more efficient by
                     // uploading first, and then pointing ore at our uploaded vmdk
                     utils.shwrap("""
+                    # https://github.com/coreos/mantle/issues/1023
+                    export AWS_SDK_LOAD_CONFIG=1
                     coreos-assembler buildextend-aws ${suffix} \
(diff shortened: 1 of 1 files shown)


jlebon:
Still testing this; working out final permissions issues with releng.

cgwalters:
> For now, we're just patching things together so we at least have `us-east-1` for preview.

You say that like there are other AWS regions?

arithx:
> > For now, we're just patching things together so we at least have `us-east-1` for preview.
> 
> You say that like there are other AWS regions?

*cries in us-west-2*

jlebon:
OK, tested working now! :heavy_check_mark:
You can see the uploaded AMI in the `jlebon` developer stream in the browser.

Some of these hacks we can clean up once https://github.com/coreos/mantle/pull/1028 and https://github.com/coreos/mantle/pull/1027 make it into cosa.

jlebon:
We have a `testing` AMI now, though it doesn't quite work yet. See https://github.com/coreos/coreos-assembler/pull/616.

PR #103, opened 2019-07-16
Jenkinsfile: add CHECKSUMS file in uploads

jlebon:
There's no way to verify right now either that the artifacts were
corrupted during uploads or tampered with. The right solution for this
is signing (see [1]), but until then, let's at least upload a CHECKSUMS
file for the former, and display it in the Jenkins logs as a temporary
hack for the latter (I mean, the logs are rotated out eventually, so
it's definitely not ideal).

[1] https://github.com/coreos/coreos-assembler/issues/268


jlebon:
(Not to mention this also duplicates the compressed image artifact checksums already in `meta.json`.)

PR #107, opened 2019-07-19
Jenkinsfile: use latest release on stream as parent

jlebon:
We want to maintain OSTree history between our releases. To do this, we
fetch the latest release from the release index, and pass it to cosa
through the `--parent` switch.

For more information, see:
https://github.com/coreos/coreos-assembler/issues/159#issuecomment-512594688

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -85,6 +85,7 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
         }
 
         def developer_builddir = "/srv/devel/${developer_prefix}/build"
+        def basearch = utils.shwrap_capture("coreos-assembler basearch")
 
         stage('Init') {
 
@@ -260,7 +261,7 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
                 if (!params.MINIMAL) {
                     // And make AMIs launchable by all; XXX: should probably integrate this into
                     // `plume release --distro fcos` along with copying into other regions.
-                    def hvm = utils.shwrap_capture("jq -r '.amis[0].hvm' builds/${newBuildID}/x86_64/meta.json")
+                    def hvm = utils.shwrap_capture("jq -r '.amis[0].hvm' builds/${newBuildID}/${basearch}/meta.json")
                     utils.shwrap("""
                     aws ec2 --region us-east-1 modify-image-attribute \
                         --image-id ${hvm} \
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -115,6 +115,8 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
                 utils.shwrap("""
                 coreos-assembler buildprep s3://${s3_stream_dir}/builds
                 """)
+                // also fetch releases.json to get the latest build, but don't error if it doesn't exist
+                utils.aws_s3_cp_allow_noent("s3://${s3_stream_dir}/releases.json", "tmp/releases.json")
             } else if (!official && utils.path_exists(developer_builddir)) {
                 utils.shwrap("""
                 coreos-assembler buildprep ${developer_builddir}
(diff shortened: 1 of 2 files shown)


jlebon:
Note this is required before we can do a new release.

Jenkinsfile
```diff
@@ -110,9 +111,11 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
         }
 
         stage('Fetch') {
-            if (s3_builddir) {
+            if (s3_stream_dir) {
                 utils.shwrap("""
-                coreos-assembler buildprep s3://${s3_builddir}
+                coreos-assembler buildprep s3://${s3_stream_dir}/builds
+                # also fetch releases.json to get the latest build, but don't error if it doesn't exist
+                aws s3 cp s3://${s3_stream_dir}/releases.json tmp/releases.json || true
```
cgwalters: Stuff like this is dangerous, really want to only ignore `ENOENT`.

jlebon: Agreed (`buildprep` does this); I just cut a corner here since there's no easy way to do this from the CLI. "Dangerous" might be a bit too strong a word though; regardless of whether this fails or not, we will always need to sanity check the parent as part of our release process (manually for now, but encoded into our release tools too in the future).

I guess since we're in the cosa container, we *could* `sh "python3 -c ' ... '"` so we can use the boto3 API directly...

jlebon: Fixed! :arrow_down:

utils.groovy
```diff
@@ -35,4 +35,36 @@ def path_exists(path) {
     return shwrap_rc("test -e ${path}") == 0
 }
 
+def get_basearch() {
+    // behold! python in shell in groovy!
```
cgwalters: :laughing:

Jenkinsfile
```diff
@@ -130,11 +133,22 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             prevBuildID = utils.shwrap_capture("readlink builds/latest")
         }
 
+        def parent_version = ""
+        def parent_commit = ""
         stage('Build') {
+            def parent_arg = ""
+            if (utils.path_exists("tmp/releases.json")) {
+                parent_commit = utils.shwrap_capture("""
+                    jq -r '.releases[-1].commits[] | select(.architecture == "${basearch}") | .checksum' tmp/releases.json
```
cgwalters: This seems OK, but we could also parse the JSON directly in Groovy which seems less awkward.

Jenkinsfile
```diff
@@ -146,6 +159,20 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
...
         } else {
             currentBuild.description = "[${params.STREAM}] ⚡ ${newBuildID}"
+
+            // and insert the parent info into meta.json so we can display it in
+            // the release browser and for sanity checking
+            if (parent_commit && parent_version) {
+                utils.shwrap("""
+                echo '{
+                    "fedora-coreos.parent-version": "${parent_version}",
+                    "fedora-coreos.parent-commit": "${parent_commit}"
```
cgwalters: Similarly here, seems way less ugly to do in Groovy.

cgwalters: i.e.:

```
def metaPath = "builds/${newBuildID}/${basearch}/meta.json"
meta = readJSON file: metaPath
meta["fedora-coreos.parent-version"] = parent_version
meta["fedora-coreos.parent-commit"] = parent_commit
writeJSON file: metaPath, json: meta
```

or so (not tested)

arithx: Alternatively you could update it all in `jq` at once like so:

```
jq '. + {"fedora-coreos.parent-commit": "${parent_commit}", "fedora-coreos.parent-version": "${parent_version}"}' < $meta > tmp/meta.json.new
mv tmp/meta.json.new $meta

jlebon: OK yeah, that's definitely much cleaner. Done! :arrow_down:

jlebon:
OK, and now using https://github.com/coreos/coreos-assembler/pull/659!

PR #108, opened 2019-07-24
Jenkinsfile: rely on plume to modify image attributes

arithx:
`plume release --distro fcos` now has the ability to modify image
attributes including launch permission; switch to that over manually
setting them.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -278,17 +278,6 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
                     --channel ${params.STREAM} \
                     --bucket ${s3_bucket}
                 """)
-
-                if (!params.MINIMAL) {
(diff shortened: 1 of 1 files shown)


arithx:
Requires: https://github.com/coreos/coreos-assembler/pull/663

jlebon:
Awesome, thanks!

Jenkinsfile
```diff
@@ -278,17 +278,6 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
                     --channel ${params.STREAM} \
```
jlebon: Super minor: let's update that comment above to add e.g. "and makes the AMI public" ?

arithx:
Superseded by https://github.com/coreos/fedora-coreos-pipeline/pull/110

Jenkinsfile
arithx: Done in #110

PR #114, opened 2019-07-26
Jenkinsfile: add kola run before building other stages

arithx:
Initial work, currently does not store the artifacts and is only running against the QEMU image.

My current plan is to add those features (artifact storage & running against different platforms) in later PRs.

Supersedes #31

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -171,6 +171,22 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             }
         }
 
+        stage('QEMU Kola Run') {
+            utils.shwrap("""
+            coreos-assembler kola run || :
+            tar -cf - tmp/kola/ | xz -c9 > _kola_temp.tar.xz
+            """)
+            archiveArtifacts "_kola_temp.tar.xz"
+        }
+
+        // archive the image if the tests failed
+        def report = readJSON file: "tmp/kola/reports/report.json"
+        if (report["result"] != "PASS") {
+            archiveArtifacts "builds/latest/**/*.qcow2"
+            currentBuild.result = 'FAILURE'
+            return
+        }
+
         if (!params.MINIMAL) {
             stage('Build Metal') {
                 utils.shwrap("""


Jenkinsfile
```diff
@@ -152,6 +152,12 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             """)
         }
 
+        stage('Test QEMU Image') {
```
jlebon: I think we want this after the next block below, i.e. only if it's actually a new build.

jlebon: Also, super bikeshed: WDYT about having `Kola` in the stage name? E.g. `QEMU Kola Run`?

arithx: Ah yes, for whatever reason I thought that the new build check had already been done. Moving it below the next block.

arithx:
Update pushed addressing comments ⬆️

jlebon:
> currently does not store the artifacts

Hmm, though right now we're not even storing the logs right? So it'd be really tricky to debug issues. Maybe at least let's do:

```
archiveArtifacts 'tmp/kola/*'
```

Though... once we do that, it shouldn't be *too* hard to check if it exited `rc != 0` and also `archiveArtifacts 'builds/**/*.qcow2'` if so. :)

Just trying to avoid being in a situation where a kola run blocks the pipeline and we have no easy way to investigate it locally ourselves.

Jenkinsfile
```diff
@@ -171,6 +171,19 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
...
+        stage('QEMU Kola Run') {
+            utils.shwrap("""
+            coreos-assembler kola run
+            tar -cf - tmp/kola/ | xz -c9 > _kola_temp.tar.xz
+            """)
+            archiveArtifacts: "_kola.tar.xz"
+            // archive the image if the tests failed
+            def kolaResult = utils.shwrap_capture("jq -r '.result' < tmp/kola/reports/report.json")
+            if (kolaResult == "FAIL") {
+                archiveArtifacts: "builds/${newBuildID}/fedora-coreos-${newBuildID}-qemu.qcow2"
```
jlebon: I think we also want to return early in this case? E.g. something like:

```
currentBuild.result = 'FAILURE'
return
```

Though one gotcha IIRC is that the `return` can't be in a `stage`; it has to be at the "top-level" script.

Jenkinsfile
```diff
@@ -171,6 +171,19 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             }
         }
 
+        stage('QEMU Kola Run') {
+            utils.shwrap("""
+            coreos-assembler kola run
+            tar -cf - tmp/kola/ | xz -c9 > _kola_temp.tar.xz
```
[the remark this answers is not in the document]
arithx: Going with `|| :` for this PR.

As far as the `--exit-zero-on-test-failures` it can be hard to make distinctions about what failures are environment (AN: cloud) related and what are legitimate test failures from the runner's perspective. I'll see if I can spend some time thinking up some potential ways that could potentially do something like that and write up a mantle issue if I come up with anything.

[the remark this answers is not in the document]
jlebon: > it can be hard to make distinctions about what failures are environment (AN: cloud) related and what are legitimate test failures from the runner's perspective

Yeah, agreed (we've had to deal with this with PAPR too). Though for the qemu target at least, we should be able to rule those out. :)

Jenkinsfile
```diff
@@ -171,6 +171,22 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
...
+            coreos-assembler kola run || :
+            tar -cf - tmp/kola/ | xz -c9 > _kola_temp.tar.xz
+            """)
+            archiveArtifacts: "_kola_temp.tar.xz"
+        }
+
+        // archive the image if the tests failed
+        def report = readJSON file: "tmp/kola/reports/report.json"
+        if (report["result"] != "PASS") {
+            archiveArtifacts: "builds/${newBuildID}/fedora-coreos-${newBuildID}-qemu.qcow2"
```
jlebon: Let's use `builds/**/*.qcow2` instead to fish out the qemu image? There really should only be one in there, and it's less fragile than specifying the full filename, which could change. Actually, the path here for example is missing `${basearch}`.

jlebon:
Awesome, thanks so much for working on this! I see this as a significant first step towards streamlining the release process.

PR #115, opened 2019-07-30
Jenkinsfile: run Publish step on master

jlebon:
To be able to start the release job, we need `oc` which is not in cosa.
Run that step on the master itself, just like the wrapper jobs do.


jlebon:
Tested working!

PR #116, opened 2019-07-30
Jenkinsfile: compress qcow2 before archiving on failure

jlebon:
Use cosa to compress the qcow2, so we get multi-threading, before
archiving. Otherwise, we risk overloading the Jenkins master PVC before
too long.

Another strategy eventually would be to still upload to S3 (see [1]
which is related to this). But even then, we'd still want to compress
first.

[1] https://github.com/coreos/coreos-assembler/issues/668


jlebon:
Hold until tested (meant to open this in draft mode instead!).

jlebon:
Yup, confirmed working here too. Thanks for testing!

PR #118, opened 2019-07-30
Jenkinsfile: fix cache nuking logic

jlebon:
First, we were using `du` on the symlink, which always returned 0, so we
never actually nuked the cache. (This is a regression from when we
switched to using the PVC for caching the qcow2s only). Moving it to the
`Init` stage means we can directly use the real filename instead.

Second, we would (try to) nuke the cache after a build, but really we
want to handle this situation *before* a build. This code predates cosa
learning to regenerate the qcow2 at build time instead of only at `init`
time.

Finally, be slightly more conservative and knock the limit down to 7G.
The disk is 10G; so even if we just have 3G left, that should be plenty
for composing into (we should still probably implement some pkgcache
cleaning heuristics in rpm-ostree itself to make this less awkward
thought).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -109,6 +109,14 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             mkdir -p \$(dirname ${cache_img})
             ln -s ${cache_img} cache/cache.qcow2
             """)
+
+            // If the cache img is larger than 7G, then nuke it. Otherwise
+            // it'll just keep growing and we'll hit ENOSPC. It'll get rebuilt.
+            utils.shwrap("""
+            if [ -f ${cache_img} ] && [ \$(du ${cache_img} | cut -f1) -gt \$((1024*1024*7)) ]; then
+                rm -vf ${cache_img}
+            fi
+            """)
         }
 
         stage('Fetch') {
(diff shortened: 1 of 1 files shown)


jlebon:
Noticed this when the bodhi-updates stream started failing composes (see https://github.com/projectatomic/rpm-ostree/issues/1878).

Jenkinsfile
```diff
@@ -109,6 +109,14 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             mkdir -p \$(dirname ${cache_img})
             ln -s ${cache_img} cache/cache.qcow2
             """)
+
+            // If the cache img is larger than 7G, then nuke it. Otherwise
+            // it'll just keep growing and we'll hit ENOSPC. It'll get rebuilt.
+            utils.shwrap("""
+            if [ -f ${cache_img} ] && [ \$(du ${cache_img} | cut -f1) -gt \$((1024*1024*7)) ]; then
```
arithx: If the `cache_img` can be up to 7G & the PVC disk is only 10G then wouldn't there potentially be a problem if the sum of the official & developer stream caches add up to 10G (assuming you were running official & developer pipelines in the same instance)?

jlebon: The PVC is 100G. The qcow2 disk is 10G: https://github.com/coreos/coreos-assembler/blob/7f7c61bcad6e36c01c59f4104fcd354e858c83be/src/cmdlib.sh#L173. We also don't share PVCs between fedora-coreos and fedora-coreos-devel.

You're right though that e.g. fedora-coreos-devel (or a local env) then only technically supports 10 different developer prefixes. I think if we even come close to this, we could just bump the PVC or just conditionally disable caching if there's not enough space.

arithx: Ah yeah, I misread the disk size as referring to the PVC.

Issue #7, opened 2018-11-05
make build logs public


dustymabe:
Let's make sure the jenkins build logs public so they can be shared and we can collaborate on issues.

jlebon:
Related: #105 

I say there:

> I'd prefer we don't upload Jenkins logs to the S3 bucket in case credentials get leaked somehow in there.

Though maybe that's too conservative? OTOH, since we don't use regular Jenkins credentials, it doesn't do that cool auto-masking in case creds do end up in the logs. But OTOOH, in practice the only creds we actually use are AWS credentials. And that gets transparently passed through env vars to the AWS CLI/SDKs we use. (And because we're using `AWS_CONFIG_FILE`, even an `env` output would just print the file path).

So I think this is probably OK, but we need to make sure we're very aware of it when hacking on pipeline code that handle creds and by extension cosa and mantle.

jlebon:
Another slightly more complex but more foolproof approach is making the pipeline job entirely credentials-less, and make the uploading part a separate job entirely.

bgilbert:
> So I think this is probably OK, but we need to make sure we're very aware of it when hacking on pipeline code that handle creds and by extension cosa and mantle.

That seems... brittle.  A credentials leak wouldn't be catastrophic, since malicious artifacts wouldn't be signed, but it'd still be a bad day.

I guess we could postprocess the log to explicitly filter out credentials... :confused:

jlebon:
Retitled issue. I still think this would be good to do, but we need to think through how to do it carefully. We've been moving towards native Jenkins credentials recently, which should help with this.

PR #121, opened 2019-07-31
manifests: drop duffy credentials

jlebon:
We don't use the CentOS artifact server anymore. So just drop the secret
mount in the interest of only mounting those we really need.


jlebon:
(Realized we were still mounting it in the container after thinking about credentials in https://github.com/coreos/fedora-coreos-pipeline/issues/7#issuecomment-516974810.)

PR #125, opened 2019-08-12
Jenkinsfile: add stage to replicate AWS AMI

arithx:
Adds a new stage after the `Upload AWS` stage that replicates the newly
created AMI to other regions. Currently the command will attempt to
replicate it to the following regions:

 - `us-east-1`
 - `us-east-2`
 - `us-west-1`
 - `us-west-2`
 - `eu-west-1`
 - `eu-west-2`
 - `eu-west-3`
 - `eu-central-1`
 - `eu-north-1`
 - `ap-south-1`
 - `ap-southeast-1`
 - `ap-southeast-2`
 - `ap-northeast-1`
 - `ap-northeast-2`
 - `sa-east-1`
 - `ca-central-1`

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -244,6 +244,15 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
                         --bucket s3://${s3_bucket}/ami-import
                     """)
                 }
+
+                // Replicate the newly uploaded AMI to other regions. Intentionally
+                // split out from the 'Upload AWS' stage to allow for tests to be added
+                // at a later date before replicating said image.
+                stage('Replicate AWS AMI') {
+                    utils.shwrap("""
+                    coreos-assembler aws-replicate --build=${newBuildID}
+                    """)
+                }
             }
         }
 


arithx:
Still need to test it

jlebon:
(Just thinking out loud here) I wonder if we only want to do this as part of the release job instead. We don't *need* every build to be on every region, just the builds we select for release. We definitely still want to upload to at least one region so we can run tests though.

arithx:
@jlebon: I had thought about that as well. I ended up choosing to put it here so the changes it makes would already be present for the `buildupload` as well as already having the correct contents inside of the builds directory (as `aws-replicate` is a `coreos-assembler` command it works off the build directory, so we'd need to at minimum download back the meta.json [if it doesn't exist] for a given build to be able to do a replicate)

jlebon:
OK, was reading over the new `aws-replicate` code. It should work to just `buildprep` to pull down the `meta.json` and then another `buildupload` to re-upload the updated version.

We'll actually have to deal with "partial uploads" also when we switch to signing via fedora-messaging too: we'll need to upload just the OSTree content, wait for the signature, and then upload the rest of the image artifacts.

jlebon:
Related: https://github.com/coreos/coreos-assembler/pull/709

arithx:
Tested locally. @jlebon do we want to land this as is and handle separating everything out later or would you prefer that done in this PR?

jlebon:
> @jlebon do we want to land this as is and handle separating everything out later or would you prefer that done in this PR?

I'm cool with doing it later if you'd like so we get to test this code out, though maybe a `TODO`-style comment to say we should eventually move this to the release job (or a separate job entirely if we want more flexibility)?

arithx:
Added a `TODO` comment to that effect.

Issue #68, opened 2019-06-06
Allow concurrent builds


jlebon:
Right now, Jenkins tries to be smart and reuse the same container between builds. We don't want that because it means we can't have e.g. a devel build running at the same time as a prod build, even though they don't intersect in any way.

Probably need to tweak some of these settings:
https://github.com/jenkinsci/kubernetes-plugin#over-provisioning-flags

jlebon:
Small refinement upon further inspection; it *does* spawn multiple containers if the pod spec is different, but not if they're the same it seems. In the devel case, it more often than not isn't the same (if e.g. one is running with a custom pipeline or COSA), though it could be.

Another reason to look into this is so that we can have multiple prod streams running at the same time (though we need to make sure we share cache space smartly).

jlebon:
#76 drops the `disableConcurrentBuilds()`, but we still need to tweak Jenkins flags to fully enable this.

dustymabe:
> tweak Jenkins flags to fully enable this.

which jenkins flags?

jlebon:
> which jenkins flags?

See the link in the first comment. Should be pretty straightforward, just haven't gotten around to testing it out.

jlebon:
https://github.com/coreos/fedora-coreos-pipeline/pull/194

jlebon:
This was fixed.

PR #128, opened 2019-08-26
*: setup AWS kola job & move aws replication to release job

arithx:
Creates a new job specifically for running the kola tests on AWS. As a
result moves the AWS replication out of the main `Jenkinsfile` into the
`Jenkinsfile.release`.

We'll likely have to move the AWS replication to it's own job down the
line if we want to start signing the `meta.json` / `release.json` files.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -248,18 +248,6 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
                         --grant-user ${FEDORA_AWS_TESTING_USER_ID}
                     """)
                 }
-
-                // Replicate the newly uploaded AMI to other regions. Intentionally
-                // split out from the 'Upload AWS' stage to allow for tests to be added
-                // at a later date before replicating said image.
-                //
-                // TODO: move to separate job (release?) along with code to buildprep,
-                // regenerate the release.json, and buildupload said meta & release.json
-              //stage('Replicate AWS AMI') {
-              //    utils.shwrap("""
-              //    coreos-assembler aws-replicate --build=${newBuildID}
-              //    """)
-              //}
             }
         }
 
@@ -295,6 +283,19 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
               mkdir -p ${developer_builddir}
               cp -aT builds ${developer_builddir}
               """)
+            }
+        }
+
+
+        if (!params.MINIMAL && s3_stream_dir) {
+            stage('Kola Runs') {
+                container('jnlp') {
+                    utils.shwrap("""
+                        oc start-build fedora-coreos-pipeline-kola-aws \
+                            -e VERSION=${newBuildID} \
+                            -e S3_STREAM_DIR=${s3_stream_dir}
+                    """)
+                }
             }
         }
 
--- a/Jenkinsfile.kola.aws
+++ b/Jenkinsfile.kola.aws
@@ -1,0 +1,69 @@
+def utils, streams
+node {
+    checkout scm
+    utils = load("utils.groovy")
+    streams = load("streams.groovy")
+    pod = readFile(file: "manifests/pod.yaml")
+}
+
+properties([
+    pipelineTriggers([]),
+    parameters([
+      string(name: 'VERSION',
+             description: 'Fedora CoreOS Build ID to test',
+             defaultValue: '',
+             trim: true),
+      string(name: 'S3_STREAM_DIR',
+             description: 'Fedora CoreOS S3 Stream Directory',
+             defaultValue: '',
+             trim: true)
+    ])
+])
+
+currentBuild.description = "[${params.VERSION}] Running"
+
+// substitute the right COSA image into the pod definition before spawning it
+pod = pod.replace("COREOS_ASSEMBLER_IMAGE", "coreos-assembler:master")
+
+podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultContainer: 'jnlp') {
+    node('coreos-assembler') { container('coreos-assembler') {
+        def ami, ami_region
+        def no_ami = false
+
+        stage('Fetch Metadata') {
+            utils.shwrap("""
+            coreos-assembler buildprep s3://${params.S3_STREAM_DIR}/builds
+            """)
+
+            def basearch = utils.shwrap_capture("coreos-assembler basearch")
+            def meta_json = "builds/${params.VERSION}/${basearch}/meta.json"
+            def meta = readJSON file: meta_json
+            if (meta.amis.size() > 0) {
+                ami = meta['amis'][0]['hvm']
+                ami_region = meta['amis'][0]['name']
+            } else {
+                no_ami = true
+            }
+        }
+
+        // fail immediately if the build contained no AMIs
+        if (no_ami) {
+          currentBuild.result = 'FAILURE'
+          return
+        }
+
+        stage('AWS Kola Run') {
+          utils.shwrap("""
+          kola run -p aws --aws-ami ${ami} --aws-region ${ami_region} -b fcos -j 10 || :
+          tar -cf - _kola_temp/ | xz -c9 > _kola_temp.tar.xz
+          """)
+          archiveArtifacts "_kola_temp.tar.xz"
+        }
+
+        def report = readJSON file: "_kola_temp/aws-latest/reports/report.json"
+        if (report["result"] != "PASS") {
+          currentBuild.result = 'FAILURE'
+          return
+        }
+    }}
+}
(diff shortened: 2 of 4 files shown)


Jenkinsfile.kola.aws
```diff
@@ -0,0 +1,59 @@
...
+
+currentBuild.description = "[${params.BUILDID}] Running"
+
+// substitute the right COSA image into the pod definition before spawning it
+pod = pod.replace("COREOS_ASSEMBLER_IMAGE", "coreos-assembler:master")
+
+podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultContainer: 'jnlp') {
+    node('coreos-assembler') { container('coreos-assembler') {
+        def ami, ami_region
+        stage('Fetching Build Details') {
```
jlebon: How about just "Fetch Metadata" ?

Jenkinsfile.kola.aws
```diff
@@ -0,0 +1,59 @@
...
+            coreos-assembler buildprep s3://${params.S3_STREAM_DIR}/builds
+            """)
+
+            def basearch = utils.shwrap_capture("coreos-assembler basearch")
+            def meta_json = "builds/${params.BUILDID}/${basearch}/meta.json"
+            def meta = readJSON file: meta_json
+            if (meta.amis.size() > 0) {
+                ami = meta['amis'][0]['hvm']
+                ami_region = meta['amis'][0]['name']
+            }
```
jlebon: Should there be an early exit or maybe better error out if there are no AMIs in the metadata?

Jenkinsfile.kola.aws
```diff
@@ -0,0 +1,59 @@
...
+        }
+
+        stage('AWS Kola Run') {
+          utils.shwrap("""
+          kola run -p aws --aws-ami ${ami} --aws-region ${ami_region} -b fcos || :
```
jlebon: Don't we want some `--parallel` value here?

Jenkinsfile
```diff
@@ -298,6 +298,19 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             }
         }
 
+
+        if (!params.MINIMAL) {
+            stage('Kola Runs') {
+                container('jnlp') {
+                    utils.shwrap("""
+                        oc start-build --wait fedora-coreos-pipeline-kola-aws \
```
jlebon: Suggestion: let's not `--wait` for this? Essentially, consider kola AWS testing as a follow-up job, not something that's necessarily part of the build process itself.

Jenkinsfile.release
```diff
@@ -30,6 +30,23 @@ pod = pod.replace("COREOS_ASSEMBLER_IMAGE", "coreos-assembler:master")
 
 podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultContainer: 'jnlp') {
     node('coreos-assembler') { container('coreos-assembler') {
+        // Replicate the newly uploaded AMI to other regions. Intentionally
```
jlebon: We should then remove the `aws-replicate` bit from the `Jenkinsfile`, no?

Jenkinsfile.release
```diff
@@ -30,6 +30,23 @@ pod = pod.replace("COREOS_ASSEMBLER_IMAGE", "coreos-assembler:master")
...
+        //
+        // We have to re-run the coreos-meta-translator as aws-replicate
+        // only modifies the meta.json
+        stage('Replicate AWS AMI') {
+            s3_stream_dir = "${s3_bucket}/prod/streams/${params.STREAM}"
+            utils.shwrap("""
+            coreos-assembler buildprep s3://${s3_stream_dir}/builds
+            coreos-assembler aws-replicate --build=${params.VERSION}
+            git clone https://github.com/coreos/fedora-coreos-releng-automation /var/tmp/fcos-releng
+            /var/tmp/fcos-releng/coreos-meta-translator/trans.py --workdir .
```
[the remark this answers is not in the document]
arithx: Yeah, it doesn't fit in cosa directly so I didn't have a good idea for where to fit it. Maybe just abstracting it into a generic function?

Jenkinsfile.kola.aws
```diff
@@ -0,0 +1,59 @@
...
+}
+
+properties([
+    pipelineTriggers([]),
+    parameters([
+      string(name: 'BUILDID',
+             description: 'Fedora CoreOS Build ID to test',
+             defaultValue: '',
+             trim: true),
+      string(name: 'S3_STREAM_DIR',
```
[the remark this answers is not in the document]
arithx: Because this is a generic job it doesn't have a developer split so it cannot construct the `S3_STREAM_DIR`, as such my intention was to be able to still support running AWS tests in developer pipelines.

Jenkinsfile.kola.aws
arithx: Probably but the value itself highly depends on the account it's being ran in. I don't remember the exact numbers on the default resource limits so we probably want to keep it low, maybe 5 tests at a time which shouldn't ever hit more than 15 concurrent instances?

Jenkinsfile.release
arithx: Ah yeah, I haven't refreshed this PR in a while so I think that commit got lost.

Jenkinsfile.release
```diff
@@ -30,6 +30,23 @@ pod = pod.replace("COREOS_ASSEMBLER_IMAGE", "coreos-assembler:master")
...
+        // split out from the 'Upload AWS' stage to allow for tests to be added
+        // at a later date before replicating said image.
+        //
+        // We have to re-run the coreos-meta-translator as aws-replicate
+        // only modifies the meta.json
+        stage('Replicate AWS AMI') {
+            s3_stream_dir = "${s3_bucket}/prod/streams/${params.STREAM}"
+            utils.shwrap("""
+            coreos-assembler buildprep s3://${s3_stream_dir}/builds
+            coreos-assembler aws-replicate --build=${params.VERSION}
```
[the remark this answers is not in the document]
arithx: Sure.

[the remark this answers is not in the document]
arithx: @jlebon: `buildprep` doesn't seem to accept a specific build as a parameter just the URL; does upstream work there need to happen first?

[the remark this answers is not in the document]
jlebon: Right, yeah. Was just spitballing a new UX in those `cosa` commands. Definitely doesn't need to block this.

Jenkinsfile.release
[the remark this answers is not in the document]
jlebon: We can punt on this for now. Once scripts from that repo are used some more, we can think about how to avoid `git clone`ing all over the place.

Jenkinsfile.kola.aws
jlebon: Ahh gotcha. Hmm, I guess this gets us back to https://github.com/coreos/mantle/issues/1037#issuecomment-517299593.

Yeah, let's go with `10` for now and go from there?

arithx:
Rebased & updated to support comments. Note that this shouldn't be merged until the upstream account has the credentials needed for replication added.

dustymabe:
> Rebased & updated to support comments. Note that this shouldn't be merged until the upstream account has the credentials needed for replication added.

👍 - i'm looking at permissions now for the aws replication issue, so can consider this too.

Jenkinsfile
dustymabe: hmm interesting - i would have figured we'd wanted it that way.

dustymabe:
I tested out #130 extensively today, which builds on top of this PR. We should be able to merge them both around the same time and update creds in prod to finally have aws replication and aws kola tests enabled.

Jenkinsfile
```diff
@@ -298,6 +286,19 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
             }
         }
 
+
+        if (!params.MINIMAL) {
```
jlebon: Should also be conditional on `s3_stream_dir`

Jenkinsfile.kola.aws
```diff
@@ -0,0 +1,62 @@
...
+
+            def basearch = utils.shwrap_capture("coreos-assembler basearch")
+            def meta_json = "builds/${params.VERSION}/${basearch}/meta.json"
+            def meta = readJSON file: meta_json
+            if (meta.amis.size() > 0) {
+                ami = meta['amis'][0]['hvm']
+                ami_region = meta['amis'][0]['name']
+            } else {
+                // no AMI found
+                return
```
jlebon: IIRC `return` inside a stage doesn't actually do anything. Also, thinking more on this now, I think we should just error out? It'd be pretty easy to mistake a job ending successfully because of tests passing vs actually a no-op.

Jenkinsfile
jlebon: One of the reasons why I suggested this is that kola AWS tests can take a while, and there's no reason to delay building other things while that's ongoing (though related: https://github.com/coreos/fedora-coreos-pipeline/issues/68). I also think errorring the build job if the AWS test fails isn't practical.

arithx:
Update pushed addressing comments ⬆️

PR #130, opened 2019-09-11
Enable running kola AWS tests using separate credentials

dustymabe:
We'd like to be able to run our kola AWS tests in a separate
account than the prod Fedora AWS account. The following
collection of commits enables us to specify different creds for
our build uploads and our kola tests.

Commit title summary:

- Add new AWS creds secret for running kola tests
- Move current AWS_CONFIG_FILE var to AWS_FCOS_BUILDS_BOT_CONFIG
- Jenkinsfile: remove AWS_SDK_LOAD_CONFIG var
- Jenkinsfile: add context comments to Kola Runs stage


dustymabe:
This builds on #128. 

cc @jlebon @arithx - this should get us all set up and close the book on AWS replication and also AWS kola tests.

arithx:
Any thoughts on using a different profile rather than a secondary credential file? It would allow us to not have to specify the export in every function context while still using the community account in `kola` (via the `--aws-profile` flag).

dustymabe:
> Any thoughts on using a different profile rather than a secondary credential file? It would allow us to not have to specify the export in every function context while still using the community account in `kola` (via the `--aws-profile` flag).



I thought about something similar, which is just exporting the builds-bot config as `AWS_CONFIG_FILE` in the pod.yaml and then overriding that when we call the kola for AWS testing. I decided to go with explicitly setting the variable everywhere. I think it helps us identify where we're calling into AWS (primarily a bunch of S3 operations), but I'm not super tied to it.

There are 3 options:

- Use two configs and two corresponding env vars (the current status of this PR)
- Use two configs and set the builds bot one as the default `AWS_CONFIG_FILE` var. Override `AWS_CONFIG_FILE` with `AWS_FCOS_KOLA_BOT_CONFIG` right before call to `kola`.
- Use one config with profiles where the `default` profile is the builds bot config and the `kola` profile is the kola bot config and then override with the `--aws-profile` flag.

cgwalters:
We could also use a separate pod for testing, without the prod creds even mounted in.

cgwalters:
(I'd like to make more use of the fact that we're using Kubernetes, it's easy to run separated containers, rather than the "one big pod doing everything" flow we have today)

HACKING.md
```diff
@@ -169,7 +169,23 @@ EOF
...
+you can use a second set of credentials for the kola tests.
+
+```
+cat <<'EOF' > /path/to/kola-secret
+[default]
+aws_access_key_id=keyid
+aws_secret_access_key=key
+EOF
+
+oc create secret generic aws-fcos-kola-bot-config --from-file=config=/path/to/kola-secret
```
jlebon: Is the IAM account called `fcos-kola-bot`? I'd like to keep the exact name of the IAM in the secret name to make things easier to track.

jlebon:
> We could also use a separate pod for testing, without the prod creds even mounted in.

Yeah, I think we should do this eventually (I'd like to even make the building and uploading bits separate to narrow it down further -- related: #7).

For reference, here's an approach where we edit the pod template on-the-fly based on some parameters: https://github.com/coreos/coreos-ci-lib/blob/67ec2b13db811ea9c601b210bd52aba51aa2791a/vars/coreos.groovy.

Anyway, I don't think we should block on this, but should probably open a ticket (or we can just re-use #7 for it).

dustymabe:
rebased on top of new changes from #128 - will give this another test and then merge both #128 and this PR

PR #131, opened 2019-09-11
Add files to track our AWS IAM permissions

dustymabe:
These files will help us track what permissions we have and allow us to methodically test and ask for changes to the policies to be made.


jlebon:
Cool, thanks for documenting this stuff!

`notes/` --> `docs/` ?

Also, can we have a tiny README.md somewhere in there to give some context?

PR #132, opened 2019-09-11
Jenkinsfile.release: make AWS AMI replication optional

dustymabe:
And default to true, but make the pipeline default to not replicate
for non-production streams.


dustymabe:
ok you can specify a boolean var for false like this:

```
oc start-build fedora-coreos-pipeline-release -e STREAM=bodhi-updates -e VERSION=30.20190911.dev.0 -e AWS_REPLICATION=
```

dustymabe:
ok I switched to using strings for the parameters. Now ready for review.

Jenkinsfile.kola.aws
```diff
@@ -55,6 +55,7 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
 
         stage('AWS Kola Run') {
           utils.shwrap("""
+          export AWS_CONFIG_FILE=\${AWS_FCOS_KOLA_BOT_CONFIG}
```
arithx: Previously the AWS config was pre-validated by nature of the build having been uploaded with the credentials being used to test. With this change to using a different config we should validate that it exists before attempting this step.

PR #135, opened 2019-09-13
Bump Jenkins to v2.176.3

jlebon:
Not tested yet.

### DIFF
--- a/manifests/pipeline.yaml
+++ b/manifests/pipeline.yaml
@@ -59,8 +59,8 @@ objects:
         - name: stable
           from:
             kind: DockerImage
-            # 2.164.2
-            name: docker.io/openshift/jenkins-2-centos7@sha256:19f37019978e9fa37d5571ab39954db5ed9bcb9d0aaedac2ba8cf6d4b4883284
+            # 2.176.3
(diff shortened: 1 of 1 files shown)


dustymabe:
I can test that this at least comes up in communishift once you feel good about the change in CCI.

jlebon:
Tested working!

jlebon:
Rolled out! :heavy_check_mark: 

(I had to `oc edit dc jenkins` and do `s/:latest/:2/` due to https://github.com/coreos/fedora-coreos-pipeline/pull/70. On that topic, I think we should periodically nuke everything and re-provision to make sure we don't fool ourselves into thinking we're GitOps'ing when really we're relying on hysteresis from e.g. 14 months ago. I think that's the idea with https://github.com/openshift/release too.)

dustymabe:
> I think we should periodically nuke everything and re-provision to make sure we don't fool ourselves into thinking we're GitOps'ing when really we're relying on hysteresis from e.g. 14 months ago

+1. Maybe once a month (or every 4 weeks would be cleaner). The only concern i have here is:

- I'd love to just delete/re-create project to ensure we clean up everything, but that's not easy in an environment we don't own
- PVCs are hard because I think they have to manually recycle them in CCI and we won't be guaranteed to get a new volume

PR #138, opened 2019-09-17
Jenkinsfiles: don't ignore errors from kola command

dustymabe:
If the kola command has trouble setting up for testing we should
make sure this get flagged up in the web UI.


dustymabe:
cc @arithx - we probably did this for a reason.

arithx:
The `kola` command also fails when any tests fail. Without the `|| :` we won't archive the test results (and/or the artifact) when tests fail.

jlebon:
See prior discussions at https://github.com/coreos/fedora-coreos-pipeline/pull/114#discussion_r308408919.

jlebon:
I filed https://github.com/coreos/mantle/issues/1059 to track this.

PR #140, opened 2019-09-17
Move the Kola AWS run back below Archive stage

dustymabe:
Turns out we need the metadata uploaded to s3 so we can fetch it
in the Jenkinsfile.kola.aws.

Reverts 13ff6d3.


dustymabe:
@arithx my personal openshift has become unreliable and the env I was testing in doesn't yet have access to `/dev/kvm` so I was using a previously built fcos and that turned out to be an invalid test.

PR #141, opened 2019-09-20
Split out OSTree compose from building qemu

jlebon:
This is prep for signing, where we'll want to
`cosa sign robosignatory --ostree` in between those two steps.


jlebon:
Currently testing this and #135 together.

dustymabe:
LGTM

jlebon:
Yup, works fine. I'm hitting some kola errors in my local pipeline, though I don't think it's related since I hit those before this patch as well.

Issue #136, opened 2019-09-16
archive stage failing in CCI


dustymabe:
seems like we've started to fail sometimes. I assume it's running out of memory?

```
[fedora-coreos-fedora-coreos-pipeline] Running shell script

+ set -xeuo pipefail

+ coreos-assembler compress --compressor xz

Targeting build: 30.20190916.dev.1

Compressing: builds/30.20190916.dev.1/x86_64

+ xz -c9 -T0 builds/30.20190916.dev.1/x86_64/fedora-coreos-30.20190916.dev.1-qemu.qcow2

+ xz -c9 -T0 builds/30.20190916.dev.1/x86_64/fedora-coreos-30.20190916.dev.1-metal.raw

Error running command xz

script returned exit code 1
```

jlebon:
Hmm, that's likely the case. We should probably add a memory request to our pod spec really.

dustymabe:
from the man page it looks like we can set `XZ_DEFAULTS=--memlimit=150MiB` and change it so that it will use less memory. Ideas for what we should set that value to?

jlebon:
Hmm, I had briefly looked at this, and it seemed like the purpose was so that xz errors out on its own if the limit is reached. But reading more carefully, you're right, it seems like compression at least can respond to it:

> If  the specified memory usage limit is exceeded when decompressing, xz will display an error and
> decompressing the file will fail.  If the limit is exceeded when  compressing,  xz  will  try  to
> scale  the  settings down so that the limit is no longer exceeded (except when using --format=raw
> or --no-adjust).  This way the operation won't fail unless the limit is very small.

So I think if we set a memlimit, as well as request at least that amount in our pod, we should be good to go. We can get a reasonable value to use with e.g. [`time -v`](https://stackoverflow.com/a/774601).

PR #142, opened 2019-09-23
manifest/pod: ensure we always have at least 2.5GiB of RAM

jlebon:
The `resources.requests.memory` tells the scheduler to put us on a node
with enough free RAM. The `XZ_DEFAULT` env tells `xz` not to go over
what we requested from OCP so we reduce the chances of being evicted for
memory consumption.

Closes: #136

---

Haven't tested this yet

### DIFF
--- a/manifests/pod.yaml
+++ b/manifests/pod.yaml
@@ -22,6 +22,10 @@ spec:
          value: /.aws-fcos-builds-bot-config/config
        - name: AWS_FCOS_KOLA_BOT_CONFIG
          value: /.aws-fcos-kola-bot-config/config
+         # used together with the request limit below to ensure xz doesn't
+         # cause us to get evicted
+       - name: XZ_DEFAULTS
+         value: "--memlimit=2150MiB"
      volumeMounts:
      - name: data
        mountPath: /srv/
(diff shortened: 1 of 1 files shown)


jlebon:
OK, this works locally. It's going to affect compilation times, slightly reverting the wins from https://github.com/coreos/coreos-assembler/pull/632. But let's see how it goes and possibly bump the RAM request if needed.

dustymabe:
It seems to be working!

```
[fedora-coreos-fedora-coreos-pipeline] Running shell script
+ set -xeuo pipefail
+ coreos-assembler compress --compressor xz
xz: Adjusted the number of threads from 40 to 1 to not exceed the memory usage limit of 2150 MiB
xz: Adjusted the number of threads from 40 to 1 to not exceed the memory usage limit of 2150 MiB
xz: Adjusted the number of threads from 40 to 1 to not exceed the memory usage limit of 2150 MiB
```

But also going from 40 threads to 1 is a little sad

dustymabe:
The archive run in the `30.20191001.dev.0` build took 40m15s 👎

jlebon:
OK, follow-up in https://github.com/coreos/fedora-coreos-pipeline/pull/145

PR #144, opened 2019-10-02
pipeline: start building live ISO

jlebon:
This was added recently in COSA. Let's try to build it at least for the
next release.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -229,6 +229,12 @@ podTemplate(cloud: 'openshift', label: 'coreos-assembler', yaml: pod, defaultCon
                 """)
             }
 
+            stage('Build Live') {
+                utils.shwrap("""
+                coreos-assembler buildextend-live
+                """)
+            }
+
             stage('Build Openstack') {
                 utils.shwrap("""
                 coreos-assembler buildextend-openstack


jlebon:
Though.. actually having it show up on the website would require hacking on the webpage too. Don't want to gate the next release on doing that, esp. since we haven't even been building them for `testing-devel`. But at least we'll have artifacts built.

bgilbert:
This LGTM (and I actually have the same change in a local branch).  We should probably hold off until https://github.com/systemd/systemd/pull/13524 lands in F30, since otherwise the artifacts won't work with SELinux enabled.

jlebon:
Hmm, coreos-koji-tagger needs to be kicked because of https://github.com/coreos/fedora-coreos-releng-automation/issues/46 to get the new systemd in, but I don't have the right auth it seems:

```
$ oc delete pod coreos-koji-tagger-20-f2jnh
Error from server (Forbidden): pods "coreos-koji-tagger-20-f2jnh" is forbidden: User "jlebon" cannot delete pods in the namespace "coreos-koji-tagger": no RBAC policy matched
$ oc exec -ti coreos-koji-tagger-20-f2jnh /bin/bash
Error from server (Forbidden): pods "coreos-koji-tagger-20-f2jnh" is forbidden: User "jlebon" cannot create pods/exec in the namespace "coreos-koji-tagger": no RBAC policy matched
```

/cc @dustymabe

jlebon:
OK, tagged it into the pool manually for now! :heavy_check_mark:

PR #145, opened 2019-10-02
manifest/pod: bump memory request to 6.5G

jlebon:
Otherwise, `xz` just scales down to 1 thread.

Nodes should have the memory for this since we used to run without any
limits before and only occasionally got evicted.

### DIFF
--- a/manifests/pod.yaml
+++ b/manifests/pod.yaml
@@ -25,7 +25,7 @@ spec:
          # used together with the request limit below to ensure xz doesn't
          # cause us to get evicted
        - name: XZ_DEFAULTS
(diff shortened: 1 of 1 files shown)


dustymabe:
LGTM, though we'll probably need to think about the local developer or "low resources personal openshift server" use case and adapt for that at some point. Maybe a parameter?

jlebon:
Yeah, you're definitely right. I'm just going to merge this for now to make this next release less painful, and then will work on parameterizing it.

jlebon:
OK, follow-up in https://github.com/coreos/fedora-coreos-pipeline/pull/146.

PR #146, opened 2019-10-03
Jenkinsfile: use smaller memory request for developers

jlebon:
Just hardcode this into the pipeline for now. As mentioned in the
comments, I don't think it warrants making it a full-blown template
parameter yet, but we can revisit if needed.

### DIFF
--- a/manifests/pod.yaml
+++ b/manifests/pod.yaml
@@ -22,10 +22,6 @@ spec:
          value: /.aws-fcos-builds-bot-config/config
        - name: AWS_FCOS_KOLA_BOT_CONFIG
          value: /.aws-fcos-kola-bot-config/config
-         # used together with the request limit below to ensure xz doesn't
-         # cause us to get evicted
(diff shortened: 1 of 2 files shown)


manifests/pod.yaml
```diff
@@ -42,9 +42,7 @@ spec:
        privileged: false
      resources:
        requests:
-         # Note the supermin VM just uses 2G. The really hungry part is xz,
-         # which without lots of memory takes lots of time.
-         memory: 6.5Gi
```
dustymabe: previously this value was slightly larger than the actual memlimit. i.e. 6.5GiB vs 6144MiB. Now they would be the same. Is that OK?

jlebon: OK, I just changed this to do the `export` from the shell so we can drop it down a bit. Also, doing some local testing on `xz` shows that for the artifact sizes we're compressing, it shouldn't need more than 5.5G. So I just lowered it to 6.5G again.

PR #148, opened 2019-10-12
pipeline: add support for signing via RoboSignatory

jlebon:
This makes use of the new `cosa sign robosignatory` command to sign both
the OSTree commit and final image artifacts.

Requires: https://github.com/coreos/coreos-assembler/pull/763


jlebon:
In case we want to use this for Monday's release, I created the configmap and secrets in the official namespace already.

PR #153, opened 2019-10-18
Jenkinsfile: only sign if there's a new build

jlebon:
Otherwise, we'll always request signing even if it's a no-op.


jlebon:
Should fix the bunch of failures we're seeing right now.

PR #154, opened 2019-10-22
Jenkinsfile: use versionary to implement version scheme

jlebon:
This uses the new `versionary.py` script which implements the new
versioning scheme:

https://github.com/coreos/fedora-coreos-releng-automation/pull/50

---

Marking as draft for the same reason as https://github.com/coreos/fedora-coreos-releng-automation/pull/50. Just flushing it out since I already had it written up and tested.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -50,10 +50,6 @@ properties([
              // list devel first so that it's the default choice
              choices: (streams.development + streams.production + streams.mechanical),
              description: 'Fedora CoreOS stream to build'),
-      // XXX: Temporary parameter for first few FCOS preview releases. We
-      // eventually want some way to drive this automatically as per the
-      // versioning scheme.
-      // https://github.com/coreos/fedora-coreos-tracker/issues/212
       string(name: 'VERSION',
              description: 'Override default versioning mechanism',
              defaultValue: '',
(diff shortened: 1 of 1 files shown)


jlebon:
Also tested now!
Requires: https://github.com/coreos/fedora-coreos-releng-automation/pull/50

jlebon:
```
[fedora-coreos-devel-jlebon-fedora-coreos-pipeline] Running shell script
+ set -euo pipefail
+ /var/tmp/fcos-releng/scripts/versionary.py
x: 31 (from manifest)
y: 20191211 (unlocked stream)
z: 94 (mapped from stream bodhi-updates)
n: 0 (no previous builds)
[Pipeline] sh
[fedora-coreos-devel-jlebon-fedora-coreos-pipeline] Running shell script
+ set -xeuo pipefail
+ coreos-assembler build ostree --skip-prune --version 31.20191211.94.0
...
```

jlebon:
Rebased!

We'll have to adapt the [release instructions](https://github.com/coreos/fedora-coreos-streams/blob/master/ISSUE_TEMPLATE/testing.md) accordingly. One interesting side-effect though is that titling the GitHub issue will require either renaming it to include the version the pipeline spit out, or just having the user run versionary first?

jlebon:
I'm going to merge this one after #166 since I think they might conflict and we're almost done testing that one.

jlebon:
Trivial rebase. Merging!

PR #156, opened 2019-10-25
HACKING.md: Document S3 bucket

cgwalters:
Since it's important.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -188,6 +188,22 @@ EOF
 oc create secret generic aws-fcos-kola-bot-config --from-file=config=/path/to/kola-secret
 ```
 
+### [OPTIONAL] Allocating S3 storage
+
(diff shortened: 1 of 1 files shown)


HACKING.md
```diff
@@ -188,6 +188,22 @@ EOF
 oc create secret generic aws-fcos-kola-bot-config --from-file=config=/path/to/kola-secret
 ```
 
+### Allocating S3 storage
```
jlebon: Let's add an `[OPTIONAL]` tag to this header. In fact, the pipeline works fine without any AWS account at all. In that case, builds are kept in the PVC, and one can use `simple-httpd` to access them.

Not practical for long-term storage of course, but for local testing and iterating, it works well.

cgwalters: Done.

jlebon:
Thanks!

PR #157, opened 2019-10-25
Add support for KVM device plugin

cgwalters:
Let's focus people on the new way forward for KVM, which is
the KVM device plugin from Kubevirt.  This is also now documented
in coreos-assembler.

Tested in the CoreOS CI cluster, an OpenShift 4.2 GCP nested virt setup following
https://github.com/coreos/coreos-assembler/blob/master/doc/openshift-gcp-nested-virt.md

I didn't test that the OCI KVM hook path still works.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -17,19 +17,32 @@ In the following sections, the section header may indicate whether the
 section applies to the local cluster case (`[LOCAL]`) or the official
 prod case (`[PROD]`).
 
+You'll want to be sure you have KVM available in your cluster.  See
+[this section of the coreos-assembler docs](https://github.com/coreos/coreos-assembler/blob/master/README.md#getting-started---prerequisites).
+
+### Using a production OpenShift cluster
+
+This is recommended for production pipelines, and also gives you
+a lot of flexibility.  The coreos-assembler document above has
+multiple options for this.  To be clear, we would also likely
+support running on "vanilla" Kubernetes if someone interested showed
+up wanting that.
+
 ### [LOCAL] Set up an OpenShift cluster
 
-First, make sure to install `oci-kvm-hook` on your host system (not in a
-pet container). This is required to ensure that the pipeline has access
-to `/dev/kvm`:
+If you're using `oc cluster up` (which is an older OpenShift with Docker)
+the easiest is to install `oci-kvm-hook` on your host system (not in a
+pet container).  NOTE: The production path for this in modern clusters is
+the [KVM device plugin](https://github.com/kubevirt/kubernetes-device-plugins/blob/master/docs/README.kvm.md)
+linked in the `[PROD]` docs.
 
 ```
 rpm-ostree install oci-kvm-hook # if on OSTree-based system
-dnf install -y oci-kvm-hook # if on traditional
+yum -y install oci-kvm-hook # if on traditional
 ```
 
 We will use `oc cluster up` to set up a local cluster for testing. To do
-this, simply obtain the OpenShift v3.6.1 binary from
+this, obtain the OpenShift v3.6.1 binary from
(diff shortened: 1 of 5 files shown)


HACKING.md
```diff
@@ -17,19 +17,30 @@ In the following sections, the section header may indicate whether the
...
+If you're using `oc cluster up` (which is an older OpenShift with Docker)
+the easiest is to install `oci-kvm-hook` on your host system (not in a
+pet container).  NOTE: The production path for this in modern clusters is
+the [KVM device plugin](https://github.com/kubevirt/kubernetes-device-plugins/blob/master/docs/README.kvm.md)
+linked in the `[PROD]` docs.
 
 ```
 rpm-ostree install oci-kvm-hook # if on OSTree-based system
-dnf install -y oci-kvm-hook # if on traditional
+yum -y install oci-kvm-hook # if on traditional
```
cgwalters: Yes, here I am changing a `dnf` to `yum` - *again*.  The reason is quite strong here: People using such an old OpenShift are quite likely also to want to run on CentOS 7, which doesn't speak `dnf`.

HACKING.md
```diff
@@ -17,19 +17,30 @@ In the following sections, the section header may indicate whether the
 section applies to the local cluster case (`[LOCAL]`) or the official
 prod case (`[PROD]`).
 
+You'll want to be sure you have KVM available in your cluster.  See
+[this section of the coreos-assembler docs](https://github.com/coreos/coreos-assembler/blob/master/README.md#getting-started---prerequisites).
+
+### [PROD] Using a production OpenShift 4 cluster
```
jlebon: `[PROD]` right now documents the process for setting up the current official pipeline. So let's leave that off until we actually change the setup.

HACKING.md
jlebon: Yup, makes sense to me.

deploy
```diff
@@ -98,6 +98,9 @@ def parse_args():
                         help="Repo and ref to use for FCOS config")
     parser.add_argument("--bucket", metavar='BUCKET',
                         help="AWS S3 bucket to use")
+    parser.add_argument("--kvm-selector", help="KVM selector",
```
jlebon: Can we document this in the `HACKING.md` doc where we describe the different switches? Also, I think we should drop `required` and default to what the official pipeline needs for now, which would be `legacy-oci-kvm-hook`. That wait, `./deploy --update --official` still works the same.

manifests/pipeline.yaml
```diff
@@ -275,6 +279,7 @@ objects:
       annotations:
         coreos.com/deploy-default: "false"
         coreos.com/s3-bucket: ${S3_BUCKET}
+        coreos.com/kvm-selector: ${KVM_SELECTOR}
```
jlebon: We don't need this one, right? The release job doesn't need KVM.

dustymabe:
Thanks. I was planning to try this out soon!

Do we have to do anything to the host cluster to enable this? I have an [open request out to fedora infra to enable this in communishift](https://pagure.io/fedora-infrastructure/issue/8208) because I assumed something had to be done by the cluster admin.

deploy
```diff
@@ -98,12 +98,15 @@ def parse_args():
...
+                        choices=['kvm-device-plugin', 'legacy-oci-kvm-hook'],
+                        default=['legacy-oci-kvm-hook'])
```
jlebon: Extra brackets

cgwalters: Fixed

cgwalters:
> Do we have to do anything to the host cluster to enable this?

You linked there to the same KVM device plugin as is linked in the assembler docs linked in this PR.

cgwalters:
I tested the KVM device plugin path with this, but not the legacy/currently-official path.

PR #158, opened 2019-10-25
jenkins: add github-oauth plugin

jlebon:
This is the first step towards setting up GitHub auth login for Jenkins.
Still playing around with this, and I'd like first to start with
enabling it for the coreos-ci project (where GitHub PRs will be tested).
But I want that project to use the exact same Jenkins definition as here
to ensure equivalence.


jlebon:
Let's get this in?

PR #161, opened 2019-11-21
Jenkinsfile.kola.aws: Use the new output dir.

dustymabe:
As part of the move to `cosa kola` the output dir is now `tmp/kola`.


dustymabe:
ok this is working in local testing.

PR #164, opened 2019-11-22
manifests: add jenkins.yaml template

jlebon:
There is this great plugin which I've somehow only learned about
recently to make Jenkins configuration much easier:

https://github.com/jenkinsci/configuration-as-code-plugin

The RHCOS pipeline already makes use of it.

To use this though, we need to be able to change the podspec of Jenkins
itself, which until now was embedded in the `jenkins-persistent`
template.

There's more though: I want to be able to define environment variables,
configmaps, secrets, etc... The default template doesn't provide
facilities for this.

So this patch essentially imports the template into our tree so that we
can have more control. The downside of course is that we lose updates to
the template. (Though that could also be seen as an upside too: we get
better reproducibility even if the template is upgrade.)

One convention I'm using is prefacing all the changes from the default
template values with a comment saying "DELTA:". That way, we can more
easily track what we added on top, and make rebasing easier in the
future.

Another advantage is that it makes setting it up easier because we don't
have to customize as many parameters: we can just set the right default
for us.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -220,11 +220,7 @@ And provide it to `--bucket` below.
 ### Create a Jenkins instance with a persistent volume backing store
 
 ```
-oc new-app --template=jenkins-persistent \
-    --param=NAMESPACE=fedora-coreos \
-    --param=MEMORY_LIMIT=2Gi \
-    --param=VOLUME_CAPACITY=2Gi \
-    --param=JENKINS_IMAGE_STREAM_TAG=jenkins:2
+oc new-app --file=manifests/jenkins.yaml --param=NAMESPACE=fedora-coreos
 ```
 
 Notice the `NAMESPACE` parameter. This makes the Jenkins master use the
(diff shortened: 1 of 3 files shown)


jlebon:
OK, this works fine! (Got another PR coming up which I'm almost done testing that works on top of this.)

dustymabe:
ok i'm working through testing this out (all the way to #166). One thing I'd like to consider (I do this occasionally) is to include the original file `jenkins.yaml.org` that is the original file along with a comment at the top about how it was created (where it came from). We should be able to periodically rebase by copying in a new jenkins.yaml.orig and then re-applying all the DELTAs on top of the new file. Keeping around the .orig is nice IMHO so we can vimdiff. Of course we can use git history for that, so I'll understand if this isn't something you think is valuable.

jlebon:
> One thing I'd like to consider (I do this occasionally) is to include the original file `jenkins.yaml.org` that is the original file along with a comment at the top about how it was created (where it came from).

Sure, seems reasonable to me. It's really a shame there isn't a cleaner way to do this. I did chat with the OpenShift Jenkins folks to confirm there wasn't a more supported way.

That said, the upstream template [doesn't change *that* often](https://github.com/openshift/jenkins/commits/master/openshift/templates/jenkins-persistent.json), so I think rebasing won't be too painful either.

jlebon:
Done! :arrow_up:

PR #167, opened 2019-11-25
Jenkinsfile.kola.aws: fix the path to the report

dustymabe:
It's no longer under `_kola_temp` and for some reason it's also no longer
under `aws-latest`.


jlebon:
OK, so I think what happened there is that we used to run `kola` directly, which defaults to creating the `${platform}-latest` symlink. But now that we've switched to `cosa kola`, which always specifies an output dir, that doesn't happen anymore.

PR #168, opened 2019-11-26
Jenkinsfile: allow overriding cosa image to use

jlebon:
Add a `COREOS_ASSEMBLER_IMAGE` parameter that allows overriding the
default one to use. This is useful on its own, but it's necessary right
now for purposely building an FCOS image with older cosa.

In the future, I think we should take the cosa image we used when doing
a release and tag it. And e.g. keep it around for the last 5 releases
per stream? (The next level from this would be using lockfiles for cosa
too... would be cool, but way more overhead!)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -36,6 +36,13 @@ node {
 // Share with the Fedora testing account so we can test it afterwards
 FEDORA_AWS_TESTING_USER_ID = "013116697141"
 
+def coreos_assembler_image
+if (official) {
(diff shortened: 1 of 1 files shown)


jlebon:
On the topic of parameters, I was looking at reworking things here to use the [job DSL plugin](https://github.com/jenkinsci/job-dsl-plugin) to enable some of what we talked about re. unifying the RHCOS and FCOS pipelines. We'd have a single YAML input file, which would get fed through all the jobs containing the key parameters we care about.

PR #171, opened 2019-11-27
Jenkinsfile.release: hardcode regions for now

jlebon:
We don't currently have permissions to query all the regions:
https://github.com/coreos/fedora-coreos-streams/issues/27#issuecomment-558854714


dustymabe:
I think it may be better to not add this as a parameter and just hardcode --regions into the replicate call briefly (to workaround our permissions issue) and then we'll revert it after the release.

If we do decide we need REGIONS it needs to probably be namespaces to AWS.

cgwalters:
xref https://aws.amazon.com/blogs/security/setting-permissions-to-enable-accounts-for-upcoming-aws-regions/

PR #177, opened 2019-12-17
Add stable stream

jlebon:
Add the entry for a stable stream so we can start building it.

---

Hold for now; we need to create the branch first.


jlebon:
OK, this was created now!

Issue #178, opened 2019-12-21
Signing OStree commit never finishing


dustymabe:
I looked at the [current job](https://jenkins-fedora-coreos.apps.ci.centos.org/job/fedora-coreos/job/fedora-coreos-fedora-coreos-pipeline/10835) running and noticed the sign ostree stage has been running for over 13H. The logs show me: 

```
[fedora-coreos-fedora-coreos-pipeline] Running shell script
+ set -xeuo pipefail
+ export AWS_CONFIG_FILE=/.aws-fcos-builds-bot-config/config
+ AWS_CONFIG_FILE=/.aws-fcos-builds-bot-config/config
+ cosa sign robosignatory --s3 fcos-builds/prod/streams/testing-devel/builds --extra-fedmsg-keys stream=testing-devel --ostree --gpgkeypath /etc/pki/rpm-gpg --fedmsg-conf /etc/fedora-messaging-cfg/fedmsg.toml
Successfully started consumer thread
Uploading s3://fcos-builds/prod/streams/testing-devel/builds/tmp/ostree-commit-object
Sending ostree-sign request for build 31.20191220.20.0
Exception in thread CrochetReactor:
Traceback (most recent call last):
  File "/usr/lib64/python3.7/threading.py", line 926, in _bootstrap_inner
    self.run()
  File "/usr/lib64/python3.7/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/usr/lib/python3.7/site-packages/crochet/_eventloop.py", line 412, in <lambda>
    target=lambda: self._reactor.run(installSignalHandlers=False),
  File "/usr/lib64/python3.7/site-packages/twisted/internet/base.py", line 1271, in run
    self.startRunning(installSignalHandlers=installSignalHandlers)
  File "/usr/lib64/python3.7/site-packages/twisted/internet/base.py", line 1251, in startRunning
    ReactorBase.startRunning(self)
  File "/usr/lib64/python3.7/site-packages/twisted/internet/base.py", line 752, in startRunning
    raise error.ReactorAlreadyRunning()
twisted.internet.error.ReactorAlreadyRunning
Process Process-1:
Traceback (most recent call last):
  File "/usr/lib/python3.7/site-packages/fedora_messaging/api.py", line 313, in publish
    eventual_result.wait(timeout=timeout)
  File "/usr/lib/python3.7/site-packages/crochet/_eventloop.py", line 239, in wait
    result = self._result(timeout)
[...]

PR #179, opened 2019-12-22
Jenkinsfile: don't sign the ostree commit for now

dustymabe:
This isn't working for some reason. Let's see if we can unblock the
pipeline. We might need to disable the signing of the artifacts later
as well.

This is related to https://github.com/coreos/fedora-coreos-pipeline/issues/178


dustymabe:
I'm not expecting any reviews since people are out for the holidays. Will merge and see if it unblocks pipeline.

PR #180, opened 2019-12-22
Jenkinsfile: don't sign the artifacts for now

dustymabe:
Followup to 586126c. Looks like the issue affects signing of the artifacts
too. This is related to https://github.com/coreos/fedora-coreos-pipeline/issues/178


dustymabe:
I'm not expecting any reviews since people are out for the holidays. Will merge and see if it unblocks pipeline.

PR #181, opened 2020-01-06
Jenkinsfile: re-enable signing

jlebon:
This should be fixed in the latest cosa now:
https://github.com/coreos/coreos-assembler/pull/1021

Closes: https://github.com/coreos/fedora-coreos-pipeline/issues/178


jlebon:
Adding hold -- requires the aforementioned cosa PR!

PR #187, opened 2020-01-15
pipeline: Run cosa kola --basic-qemu-scenarios

cgwalters:
So we test NVMe root disk, Secure Boot etc.
See https://github.com/coreos/coreos-assembler/pull/1029

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -33,6 +33,21 @@ node {
     }
 }
 
+// Parse and handle the result of Kola
+boolean checkKolaSuccess(currentBuild) {
+    // archive the image if the tests failed
+    def report = readJSON file: "tmp/kola/reports/report.json"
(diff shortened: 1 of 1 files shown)


cgwalters:
Test run of this here https://jenkins-walters-fcos.apps.coreosci.gcp.devcluster.openshift.com/job/walters-fcos/job/walters-fcos-walters-fedora-coreos-pipeline/12/console

Jenkinsfile
```diff
@@ -33,6 +33,21 @@ node {
     }
 }
 
+// Parse and handle the result of Kola
+boolean checkKolaSuccess(currentBuild) {
```
jlebon: Yeah, this is the kind of thing that we can push up into [coreos-ci-lib](https://github.com/coreos/coreos-ci-lib) at some point.

PR #188, opened 2020-01-15
Jenkinsfile: stop ignoring kola errors

jlebon:
Now that https://github.com/coreos/mantle/issues/1059 is fixed (see
https://github.com/coreos/mantle/pull/1153), we can use the new
`--no-test-exit-error` switch to be more strict about kola error
handling.

This way, we immediately fail the build if something fundamental went
wrong with kola.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -299,7 +299,7 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
 
         stage('Kola:QEMU basic') {
             utils.shwrap("""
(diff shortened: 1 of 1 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/1058

dustymabe:
will we still get the tarball of the logs?

jlebon:
> will we still get the tarball of the logs?

Yeah, everything else stays the same.

PR #189, opened 2020-01-15
pipeline: Fix crash when kola fails

cgwalters:
I forgot to commit this apparently, the error was:

`groovy.lang.MissingPropertyException: No such property: utils for class: groovy.lang.Binding`

Which...TBH I don't understand the scoping rules in Groovy.


cgwalters:
OK http://docs.groovy-lang.org/docs/latest/html/documentation/#_methods explains it.

PR #190, opened 2020-01-24
Hook up upgrade testing

jlebon:
Start running the new upgrade test right after building the QEMU image.
In the AWS test job, run the upgrade test on AWS in parallel.

For more information, see:
https://github.com/coreos/mantle/pull/1168

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -33,21 +33,6 @@ node {
     }
 }
 
-// Parse and handle the result of Kola
-boolean checkKolaSuccess(utils, currentBuild) {
-    // archive the image if the tests failed
-    def report = readJSON file: "tmp/kola/reports/report.json"
-    def result = report["result"]
-    print("kola result: ${result}")
-    if (result != "PASS") {
-        utils.shwrap("coreos-assembler compress --compressor xz")
-        archiveArtifacts "builds/latest/**/*.qcow2.xz"
-        currentBuild.result = 'FAILURE'
-        return false
-    }
-    return true
-}
-
 // Share with the Fedora testing account so we can test it afterwards
 FEDORA_AWS_TESTING_USER_ID = "013116697141"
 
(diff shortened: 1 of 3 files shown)


jlebon:
Requires: https://github.com/coreos/mantle/pull/1168

I've tested the main pipeline change successfully. I didn't test the AWS bit, but the change is more or less the same.

jlebon:
Now waiting on https://github.com/coreos/coreos-assembler/pull/1076.

jlebon:
OK, it's merged! Let's get this one in?
This is what it looks like FWIW:

```
[fedora-coreos-devel-jlebon-fedora-coreos-pipeline] Running shell script
+ set -xeuo pipefail
+ kola run-upgrade --cosa-build builds/31.20200123.dev.0/x86_64/meta.json --output-dir tmp/kola -v --find-parent-image --no-test-exit-error
2020-01-23T23:43:58Z cli: Started logging at level INFO
fedora-coreos-31.20200121.20.0-qemu.x86_64.qcow2:  0 B/449 MB
fedora-coreos-31.20200121.20.0-qemu.x86_64.qcow2:  32.8 KB/449 MB
...
fedora-coreos-31.20200121.20.0-qemu.x86_64.qcow2:  449 MB/449 MB 

=== RUN   fcos.upgrade.basic
=== RUN   fcos.upgrade.basic/setup
=== RUN   fcos.upgrade.basic/upgrade-from-previous
=== RUN   fcos.upgrade.basic/upgrade-from-current
--- PASS: fcos.upgrade.basic (84.31s)
    --- PASS: fcos.upgrade.basic/setup (13.56s)
    --- PASS: fcos.upgrade.basic/upgrade-from-previous (22.90s)
    --- PASS: fcos.upgrade.basic/upgrade-from-current (22.34s)
PASS, output in tmp/kola
```

There is an interesting question here, which is: what happens if the latest version "released" on a particular stream is known broken and thus can't upgrade? I don't expect this to happen often from now on, because that release itself must've somehow passed through the `upgrade-from-current` test. But if we do end up in that position, then yeah, I think we'll want a job parameter that allows filling in a previous version to use that is known good.

jlebon:
So actually, with https://github.com/coreos/coreos-assembler/pull/1077, we can simplify this further. So we can wait for that.

jlebon:
Tweaked this now to use `cosa kola --upgrades`! :arrow_up:

PR #191, opened 2020-01-28
Jenkinsfile.kola.aws: hook up creds for upgrade testing

jlebon:
Missed this as part of #190.

### DIFF
--- a/Jenkinsfile.kola.aws
+++ b/Jenkinsfile.kola.aws
@@ -83,6 +83,7 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
             aws_upgrade: {
                 stage('Kola:AWS upgrade') {
                     utils.shwrap("""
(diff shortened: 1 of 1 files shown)


Jenkinsfile.kola.aws
```diff
@@ -83,6 +83,7 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
             aws_upgrade: {
                 stage('Kola:AWS upgrade') {
                     utils.shwrap("""
```
cgwalters: Maybe later `utils.shwrap_aws()` or something like that?

jlebon: Yeah, makes sense. Hmm, or actually now that we're using the code-as-configuration plugin, we could do this the more Jenkins-y way by converting the Kube secrets to Jenkins credentials (like we do the Slack token now) and use `withCredentials() { ... }`. That'll make it easier to limit the creds exposed to just what the script needs.

PR #192, opened 2020-01-28
Jenkinsfile{,.kola.aws}: tweak upgrade test output dir

jlebon:
Cosa now puts the output of upgrade tests in `tmp/kola-upgrade`, so
tweak things here accordingly.

This makes the two jobs more consistent and also fixes the issue of
`Jenkinsfile.kola.aws` running both testsuites in parallel but
outputting into the same directory.


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/1080

PR #193, opened 2020-01-30
ci: hook up to CoreOS CI

jlebon:
For now, this just does a basic sanity checking that the manifest files
in `manifests/` are valid. Though in the future, we can make e.g. it run
through the pipeline itself in dry-run mode.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -1,0 +1,12 @@
+@Library('github.com/coreos/coreos-ci-lib@master') _
+
+node {
+    checkout scm
+
+    stage("Test") {
+        // for now, just sanity check all the manifests
+        coreos.shwrap("""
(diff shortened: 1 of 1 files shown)


.cci.jenkinsfile
```diff
@@ -0,0 +1,12 @@
+@Library('github.com/coreos/coreos-ci-lib@master') _
+
+node {
+    checkout scm
+
+    stage("Test") {
+        // for now, just sanity check all the manifests
+        coreos.shwrap("""
+        find manifests/ -iname '*.yaml' | xargs -n 1 oc create --dry-run -f
```
dustymabe: does this oc command need to be logged in to a remote cluster?

jlebon: No, it doesn't. (It's a good question though, since CI here *is* connected to the CI cluster).

PR #194, opened 2020-01-31
Allow concurrent builds/releases/tests

jlebon:
Sprinkle `runPolicy: Parallel` a bit everywhere, then adapt the various pipeline code to deal with it correctly. See individual commit messages for details!

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -77,8 +77,6 @@ properties([
     durabilityHint('PERFORMANCE_OPTIMIZED')
 ])
 
-currentBuild.description = "[${params.STREAM}] Running"
-
 // Note the supermin VM just uses 2G. The really hungry part is xz, which
 // without lots of memory takes lots of time. For now we just hardcode these
 // here; we can look into making them configurable through the template if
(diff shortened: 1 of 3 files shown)


jlebon:
Related to 2. in https://github.com/coreos/fedora-coreos-pipeline/pull/194/commits/b8037b701afd6dfcf454111a1074ce0f8e9283fe, I'm working on rewiring things here to move away from buildconfigs and towards using the job-dsl plugin in combination with JCasC. The CoreOS CI rework will leverage this too. One outcome of this could be to have separate jobs per stream. This would fix a few things, including:
- dropping the "meta" wrapper jobs (`Jenkinsfile.{production,mechanical,development}`)
- fixing bugs related to the OpenShift Jenkins plugin (e.g. how we have to trigger other jobs using `oc start-build` instead of the `build` step)
- making it trivial to see logs from the last e.g. `stable`/`testing` build and not worrying about having them be GC'ed
- and finally, being able to run concurrently across streams but not within would become trivial since they'd be separate jobs

Anyway, there's of course some cons too and it would affect the developer workflow, but I think it's worth exploring.

jlebon:
Hmm, actually, we don't need the `inversePrecedence: true` stuff and all the special logic supporting it. My initial motivation for doing this was misguided I think. Unlike the "trigger on git push" model of a lot of Jenkins jobs, each build could be unique by their parameters (e.g. version string, force flag, cosa image). So I think overall it makes sense to run builds serially per stream. And worse case, the following runs will just be quick no-op builds anyway (and remember, this shouldn't happen often; the lock there is more of a fail-safe).

jlebon:
> my only wish would be that we added new indentation for new sets of { } but I realize that does mean our code starts getting shifted over quite a bit

Yeah, I think we could do this more liberally in the future once we factor out the meat of the pipeline into a separate function similar to RHCOS.

PR #197, opened 2020-02-07
Prep patches for sharing with CoreOS CI

jlebon:
See individual commit messages, but essentially the end goal here is preparations for sharing things with [CoreOS CI](https://github.com/jlebon/coreos-ci). Specifically, I want CoreOS CI to use the exact same Jenkins build configs, i.e. the inputs to the S2I builder. However, the pipeline and CoreOS CI will have different JCasC configurations (provided via configmaps) and Jenkins app templates (`jenkins.yaml`).

### DIFF
--- a/deploy
+++ b/deploy
@@ -154,13 +154,19 @@ def process_template(args):
         print(f"  {k}={v}")
     print()
 
-    def gen_param_args():
-        return [q for k, v in params.items() for q in ['--param', f'{k}={v}']]
+    def gen_param_args(selected):
+        selected_params = {(k, v) for k, v in params.items() if k in selected}
+        return [q for k, v in selected_params for q in ['--param', f'{k}={v}']]
 
     resources = []
-    for template in ['pipeline.yaml']:
+    for template in ['pipeline.yaml', 'jenkins-s2i.yaml']:
+        # we only want to pass the params which each template actually
+        # supports, so filter it down for this specific template
+        with open(f'manifests/{template}') as f:
+            t = yaml.safe_load(f)
+            tparams = [p['name'] for p in t.get('parameters', [])]
         # and generate the --param FOO=bar ... args for this template
-        param_args = gen_param_args()
+        param_args = gen_param_args(tparams)
         j = json.loads(subprocess.check_output(
             ['oc', 'process', '--filename', f'manifests/{template}'] + param_args))
         resources += j['items']
--- a/deploy
+++ b/deploy
@@ -132,37 +132,44 @@ def get_username():
 
 
 def process_template(args):
-    params = []
+    params = {}
     if not args.official:
-        params += [f'DEVELOPER_PREFIX={args.prefix}']
+        params['DEVELOPER_PREFIX'] = args.prefix
     if args.pipeline:
-        params += params_from_git_refspec(args.pipeline, 'JENKINS_S2I')
-        params += params_from_git_refspec(args.pipeline, 'JENKINS_JOBS')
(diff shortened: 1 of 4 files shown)


jlebon:
Though there's a bunch of other unrelated minor fixes and enhancements in there along the way.

jlebon:
OK, I've split out the preliminary ones that are easier to review in https://github.com/coreos/fedora-coreos-pipeline/pull/198.

jlebon:
Note this is also prep for reworking how jobs are created and managed here too!

jlebon:
Re-tested this successfully now in `fedora-coreos-devel`! :heavy_check_mark:

jlebon:
Rebased!

deploy
```diff
@@ -132,37 +132,49 @@ def get_username():
...
-    params = [q for p in params for q in ['--param', p]]
+    for k, v in params.items():
+        print(f"  {k}={v}")
     print()
 
-    resources = subprocess.check_output(
-        ['oc', 'process', '--filename', 'manifests/pipeline.yaml'] + params)
-    return json.loads(resources)
+    def gen_param_args(selected):
+        return [q for k, v in params.items() if k in selected for q in ['--param', f'{k}={v}']]
```
dustymabe: I find this hard to follow - maybe splitting into two operations would make things clearer

manifests/jenkins-s2i.yaml
```diff
@@ -0,0 +1,109 @@
...
+      annotations:
+        coreos.com/deploy-default: "true"
+  - kind: BuildConfig
+    apiVersion: v1
+    metadata:
+      name: jenkins
+      annotations:
+        coreos.com/deploy-default: "true"
+        coreos.com/jenkins-jobs-url: ${JENKINS_JOBS_URL}
+        coreos.com/jenkins-jobs-ref: ${JENKINS_JOBS_REF}
```
dustymabe: we added these two annotations as well?

deploy
jlebon: Done!

manifests/jenkins-s2i.yaml
jlebon: Yeah sorry, I should've made that a separate commit. I'm not a fan myself of sneaking things into blocks that initially just looks like it was a plain move. :)

I added a comment there. Let me know if it's not clear!

deploy
```diff
@@ -154,13 +154,19 @@ def process_template(args):
         print(f"  {k}={v}")
     print()
 
-    def gen_param_args():
-        return [q for k, v in params.items() for q in ['--param', f'{k}={v}']]
+    def gen_param_args(selected):
+        selected_params = {(k, v) for k, v in params.items() if k in selected}
+        return [q for k, v in selected_params for q in ['--param', f'{k}={v}']]
```
dustymabe: I think the part that was confusing to me here was the double for loop inside a list comprehension.. here's my alternative that I can read better:

```
def gen_param_args(selected):                                             
    selected_params = {(k, v) for k, v in params.items() if k in selected}
    args = []                                                             
    for k, v in selected_params:                                          
        args += ['--param', f'{k}={v}']                                   
    return args                                                           
                                                                          

```

I'll understand if you don't want to make this change.

jlebon: Not sure where this stands on the idiomatic scale (I've definitely seen it used elsewhere), but it's really useful. Here's how to read this: when you see something like `[w for y in z for x in y for w in x]`, you can simply unroll it in that same order:

```python
for y in z:
  for x in y:
    for w in x:
      mylist.append(w)
```

dustymabe:
LGTM - merge away

PR #198, opened 2020-02-07
Prep patches for prep patches for CoreOS CI

jlebon:
Split out of #197.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -242,6 +242,12 @@ reason we create the app first is that otherwise OpenShift will
 automatically instantiate Jenkins with default parameters when creating
 the Jenkins pipeline).
 
+Now, create the Jenkins configmap:
+
+```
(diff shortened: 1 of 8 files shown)


jlebon:
Heh, wanted to use `deploy --dry-run` in CI, but of course, the master doesn't have py3. We can do that in a container, but we'll need to install `oc` too. Anyway, I can do that in a follow-up!

jlebon:
Releases built, merging!

dustymabe:
> Releases built, merging!

but we still haven't run the release pipeline for them? currently waiting on the releing ostree import

jlebon:
> but we still haven't run the release pipeline for them? currently waiting on the releing ostree import

Fair, though the risk should be pretty minimal (and if somehow it does break the release job, I'll be on standby to make sure we get it fixed :)  ).

PR #200, opened 2020-02-11
switch cosa buildprep commands to specify which build

dustymabe:
Now that buildprep supports passing in a build id let's use that
so we're more explicit.

### DIFF
--- a/Jenkinsfile.kola.aws
+++ b/Jenkinsfile.kola.aws
@@ -45,7 +45,7 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
             utils.shwrap("""
             export AWS_CONFIG_FILE=\${AWS_FCOS_BUILDS_BOT_CONFIG}
             coreos-assembler init https://github.com/coreos/fedora-coreos-config
-            coreos-assembler buildprep --ostree s3://${params.S3_STREAM_DIR}/builds
+            coreos-assembler buildprep --ostree --build=${params.VERSION} s3://${params.S3_STREAM_DIR}/builds
             """)
 
             def basearch = utils.shwrap_capture("coreos-assembler basearch")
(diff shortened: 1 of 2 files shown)


Jenkinsfile.kola.aws
```diff
@@ -45,7 +45,7 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
             utils.shwrap("""
             export AWS_CONFIG_FILE=\${AWS_FCOS_BUILDS_BOT_CONFIG}
             coreos-assembler init https://github.com/coreos/fedora-coreos-config
-            coreos-assembler buildprep --ostree s3://${params.S3_STREAM_DIR}/builds
+            coreos-assembler buildprep --ostree --build=${params.VERSION} s3://${params.S3_STREAM_DIR}/builds
```
jlebon: We also need to pass `--build` to the `cosa kola` commands below.

Jenkinsfile.release
```diff
@@ -64,17 +64,14 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
...
-                // operate on the specific build rather than the most
-                // recent build
                 utils.shwrap("""
                 export AWS_CONFIG_FILE=\${AWS_FCOS_BUILDS_BOT_CONFIG}
                 coreos-assembler init https://github.com/coreos/fedora-coreos-config
-                coreos-assembler buildprep s3://${s3_stream_dir}/builds
+                coreos-assembler buildprep --build=${params.VERSION} s3://${s3_stream_dir}/builds
                 coreos-assembler aws-replicate --build=${params.VERSION} --log-level=INFO
                 git clone https://github.com/coreos/fedora-coreos-releng-automation /var/tmp/fcos-releng
                 /var/tmp/fcos-releng/coreos-meta-translator/trans.py --workdir .
```
jlebon: And here, we have to pass `--build-id` too.

Jenkinsfile.kola.aws
dustymabe: I wasn't sure about this because the first one passes in `--aws-ami ${ami}` which I would think is all the information that is being gleaned from the builds directory. Is it ?

jlebon: The thing is that `cosa kola` today automatically passes `--cosa-build` to `kola`. So if we *don't* override it, I'm pretty sure it would fail entirely.

You're correct though that `kola run -p aws` today doesn't actually use the data from `meta.json`. But it will soon, and we'll be able to e.g. drop the whole `--aws-ami` switch and logic for it once we implement it: https://github.com/coreos/mantle/blob/d17c87d029ea905ae88b18d7e9fed7fd7a5e7968/cmd/kola/options.go#L204

dustymabe: ok I think I've addressed this now

Jenkinsfile.release
dustymabe: i think I've fixed this now

PR #203, opened 2020-02-13
Jenkinsfiles: add hint for vim to choose correct sytax highlighting

dustymabe:
Without the .groovy extension these files won't get any highlighting.
Setting this on the modeline helps.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,3 +1,4 @@
+// vim: set syntax=groovy:
 import org.yaml.snakeyaml.Yaml;
 
 def utils, streams, official, official_jenkins, developer_prefix, src_config_url, src_config_ref, s3_bucket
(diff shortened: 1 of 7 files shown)


jlebon:
Not against, though with this you'll get even better highlighting: https://github.com/martinda/Jenkinsfile-vim-syntax. :)

dustymabe:
> Not against, though with this you'll get even better highlighting: https://github.com/martinda/Jenkinsfile-vim-syntax. :)

seems like this `Simply recognises your project's Jenkinsfile as a Groovy script.` So it's effectively the same as what I'm doing here, except the change in this PR will work for anyone who hasn't specifically installed that plugin? It would be nice if we could get that plugin in Fedora.

jlebon:
Hmm, I think that comment might be outdated. It definitely highlights Jenkins-specific things: https://github.com/martinda/Jenkinsfile-vim-syntax/blob/master/syntax/Jenkinsfile.vim.

jlebon:
And the problem is that with those hints, I actually get worse highlighting because it's using the Groovy syntax highlighter. :(

PR #206, opened 2020-02-19
Jenkinsfile: do `oc get pods -o wide` upfront

jlebon:
We're seeing occasional issues where pods can't find `/dev/kvm` even
though we have a `oci_kvm_hook: allowed` selector (and in fact
`oci-kvm-hook` is installed on the whole cluster).

Print out the pod list on startup so that we can figure out on which
node the pod was scheduled to help debugging the next time we hit this.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -137,6 +137,11 @@ lock(resource: "build-${params.STREAM}") {
     podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
     node(pod_label) { container('coreos-assembler') {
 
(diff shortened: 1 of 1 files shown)


cgwalters:
Seems a bit odd to list *all* pods versus just the current pod name perhaps using the [downward API](https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/) ?

jlebon:
Yeah, I wasn't sure how to get the name of the specific pod we're running in so just went with a quick & safe shotgun approach. Looks like that API would do the trick though.

PR #209, opened 2020-03-17
pipeline: Run kola testiso

cgwalters:
On general principle, no reason not to run this test.


cgwalters:
I didn't test this though.

PR #213, opened 2020-03-25
manifests/pod: stop using PVC cache for now

jlebon:
We've been hitting strange I/O related issues recently. We originally
traced it down to a specific node that was cordoned off, but now we're
hitting it on other nodes as well.

I think it might be related to the cache qcow2 being on an NFS and some
I/O patterns we do trigger some issues. Anyway, since the pipeline is
failing pretty bad, let's test that theory and swap it out for just an
`emptyDir`.

Also rename the mount to `cache` rather than `data` to make it clear we
don't actually need it.

### DIFF
--- a/manifests/pod.yaml
+++ b/manifests/pod.yaml
@@ -25,7 +25,7 @@ spec:
        - name: AWS_FCOS_KOLA_BOT_CONFIG
          value: /.aws-fcos-kola-bot-config/config
      volumeMounts:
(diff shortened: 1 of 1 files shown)


jlebon:
And we can tell exactly how much time we lose from not having a cache: the `Fetch` stage went from 48s to 2m30. So less than 2m, which is not bad at all.

Issue #195, opened 2020-02-06
Build should time out if it takes more than X hours


jlebon:
E.g. we had a build recently just hang on S3 uploads.

Or could try to be fancier and do different timeouts per stage. But a big wrapping timeout would work to begin with.

dustymabe:
👍

dustymabe:
hit this again.. I just killed https://jenkins-fedora-coreos.apps.ci.centos.org/job/fedora-coreos/job/fedora-coreos-fedora-coreos-pipeline/11960/  which has been running almost a day.

jlebon:
https://github.com/coreos/fedora-coreos-pipeline/pull/214

PR #214, opened 2020-03-25
pipeline: add overall timeout of 4h

jlebon:
Let's cap build runs to 4h. Builds should never take that long. Ideally,
we'd do timeouts per stage, and that could make sense (e.g. signing for
example could in the worst case take more than 1h, but building the
qcow2 shouldn't really take more than 15m), though just having one
overall timeout to start with will at least prevent jobs from getting
stuck anywhere in the process forever and taking up a pod in a cluster.

Closes: #195

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -145,7 +145,7 @@ lock(resource: "build-${params.STREAM}") {
         // declare this early so we can use it in Slack
         def newBuildID
 
-        try {
+        try { timeout(time: 240, unit: 'MINUTES') {
 
         // Clone the automation repo, which contains helper scripts. In the
         // future, we'll probably want this either part of the cosa image, or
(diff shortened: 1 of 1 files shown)


dustymabe:
can we make it 4 or 5 ? I think the goal here is to make sure builds don't sit there forever. 3h makes me a little worried mostly because robosignatory can get stuck for long periods of time.

jlebon:
> can we make it 4 or 5 ? I think the goal here is to make sure builds don't sit there forever. 3h makes me a little worried mostly because robosignatory can get stuck for long periods of time.

We currently already cap signing to 1h in cosa though. I think you're suggesting making that larger too? Which we could... but part of the reason why we don't is because due to https://pagure.io/robosignatory/issue/3, there's no way to tell the difference between "Robosignatory is down" and "Robosignatory is busy". In the case of the former, failing earlier means we get the attention of sysadmins earlier.

dustymabe:
> We currently already cap signing to 1h in cosa though. I think you're suggesting making that larger too?

No. What I'm saying is that because the robosignatory (currently timing out at 1 hour) operations can take a long time we could approach that 3 hour limit and cancel a build that's progressing just fine. The stable build that ran today took `Took 1 hr 56 min`. So if we have robosig issues that take a long time we could kill the build in home stretch.

Can we just bump to 4 hours? That would make me more comfortable.

jlebon:
Yup, OK sure!

PR #216, opened 2020-03-25
utils: fail the build if kola tests fail

jlebon:
We want to always fail the build if the result is not `PASS`. But we
only want to archive the qcow2 if the platform is `qemu-unpriv`.


dustymabe:
xref: https://github.com/coreos/fedora-coreos-streams/issues/71#issuecomment-604115245

PR #218, opened 2020-04-02
pipeline: also test metal4k installs

jlebon:
`kola testiso` knows how to do this now!

https://github.com/coreos/coreos-assembler/pull/1312

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -345,16 +345,26 @@ lock(resource: "build-${params.STREAM}") {
                 """)
             }
 
-            stage('Test Live ISO') {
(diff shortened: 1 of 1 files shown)


jlebon:
Rebased this now! Though not tested in a developer pipeline.

jlebon:
Actually, we should use https://github.com/coreos/coreos-assembler/pull/1358 first before merging this so we avoid recompressing a bunch of times.

jlebon:
OK, updated this to use `cosa compress --artifact metal --artifact metal4k`!

dustymabe:
still LGTM

PR #219, opened 2020-04-06
Add GCP image uploads to the pipeline

dustymabe:


dustymabe:
Currently WIP, but this should be close. Doing some testing.

Will add kola tests in a followup.

HACKING.md
```diff
@@ -201,6 +201,27 @@ EOF
...
 ```
 
+### [OPTIONAL] Creating GCP credentials configs
+
+If you are in production where we upload images to GCP OR you want to
+test uploading to GCP as part of your pipeline development, you need to
+create a credentials config as a secret within OpenShift.
+
+First grab a json file that represents the credentials for your service account
+from GCP. Then create the secret in OpenShift:
```
jlebon, line 211: Can we put more details here on where to get the JSON file from/what format it is?

manifests/pod.yaml
```diff
@@ -65,6 +77,17 @@ spec:
...
       secretName: aws-fcos-kola-bot-config
       optional: true
+  # This secret is used for uploading to GCP
+  - name: gcp-image-upload-config
+    secret:
+      secretName: gcp-image-upload-config
+      optional: true
+  # This secret is used for running GCP kola tests
+  - name: gcp-image-upload-config
+  - name: gcp-kola-tests-config
```
jlebon, line 87: Two `- name:` here.

manifests/pod.yaml
```diff
@@ -24,6 +24,12 @@ spec:
          value: /.aws-fcos-builds-bot-config/config
        - name: AWS_FCOS_KOLA_BOT_CONFIG
          value: /.aws-fcos-kola-bot-config/config
+       - name: GCP_IMAGE_UPLOAD_CONFIG
+         value: /.gcp-image-upload-config
+       - name: GCP_KOLA_TESTS_CONFIG
+         value: /.gcp-kola-tests-config
+  - name: gcp-image-upload-config
+  - name: gcp-kola-tests-config
```
jlebon, line 32: Two `- name:` here.

manifests/pod.yaml
```diff
@@ -24,6 +24,12 @@ spec:
          value: /.aws-fcos-builds-bot-config/config
        - name: AWS_FCOS_KOLA_BOT_CONFIG
          value: /.aws-fcos-kola-bot-config/config
+       - name: GCP_IMAGE_UPLOAD_CONFIG
+         value: /.gcp-image-upload-config
+       - name: GCP_KOLA_TESTS_CONFIG
+         value: /.gcp-kola-tests-config
```
jlebon, line 30: Yeah... we should really look at constraining each container run to the minimal set of credentials it needs. E.g. no reason the AWS kola test needs GCP creds mounted in.

Jenkinsfile
```diff
@@ -23,6 +23,8 @@ node {
     src_config_ref = utils.get_pipeline_annotation('source-config-ref')
     s3_bucket = utils.get_pipeline_annotation('s3-bucket')
     kvm_selector = utils.get_pipeline_annotation('kvm-selector')
+    gcp_gs_bucket = utils.get_pipeline_annotation('gcp-gs-bucket')
+    gcp_project = utils.get_pipeline_annotation('gcp-project')
```
jlebon, line 27: Hmm, missing new annotations for this in the manifest?

Jenkinsfile
```diff
@@ -416,6 +418,21 @@ lock(resource: "build-${params.STREAM}") {
                     """)
                 }
             }
+
+            // If there is a config for GCP then we'll upload our image to GCP
+            if (utils.path_exists("\${GCP_IMAGE_UPLOAD_CONFIG}")) {
```
jlebon, line 423: Can we run this in parallel with the `Upload AWS` stage?

dustymabe: yeah - I started to copy over some paralllel code and doing it that way to begin with but both of them update the metadata IIUC and I wanted to just sprint to getting this in and being able to do followups later.

manifests/pod.yaml
dustymabe: I assume you mean something for us to do in the future?

Jenkinsfile
dustymabe: fixed

HACKING.md
dustymabe: fixed

manifests/pod.yaml
dustymabe: fixed I think

dustymabe: fixed I think

Jenkinsfile
jlebon: Ahhh right, that issue. OK yup, let's leave that for later. An easy fix for this would be to add locking in cosa for the crucial bits.

manifests/pod.yaml
jlebon: Yup, definitely!

manifests/pod.yaml
dustymabe: fixed I think

dustymabe:
pushed up a few small fixes for this. Now seems to be working in the pipeline!

dustymabe:
depends on https://github.com/coreos/coreos-assembler/pull/1319

dustymabe:
ok added a commit that removes the need to specify the project in various places. One final review before merge?

dustymabe:
https://github.com/coreos/coreos-assembler/pull/1319 merged.

PR #220, opened 2020-04-07
Jenkinsfile: add description for the GCP uploaded image

dustymabe:
For public images GCP would like for the description of the image to be
in the format: `DISTRO, DISTRO RELEASE, VERSION, INFO` so that information
can get pulled into web UIs and documentation. Let's start with something
like:

```
Fedora CoreOS, Fedora CoreOS stable, 31.20200310.3.0, x86_64 published on 2020-03-25
```

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -424,13 +424,15 @@ lock(resource: "build-${params.STREAM}") {
                     utils.shwrap("""
                     # pick up the project to use from the config
                     gcp_project=\$(jq -r .project_id \${GCP_IMAGE_UPLOAD_CONFIG})
+                    today=\$(date +%Y-%m-%d)
                     cosa buildextend-gcp \
                         --build=${newBuildID} \
                         --upload \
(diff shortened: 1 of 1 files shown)


dustymabe:
depends on https://github.com/coreos/coreos-assembler/pull/1328

Jenkinsfile
```diff
@@ -424,13 +424,15 @@ lock(resource: "build-${params.STREAM}") {
                     utils.shwrap("""
                     # pick up the project to use from the config
                     gcp_project=\$(jq -r .project_id \${GCP_IMAGE_UPLOAD_CONFIG})
+                    today=\$(date +%Y-%m-%d)
```
bgilbert, line 427: Are we concerned about time zones?

dustymabe: i'm not. It's purely informational

dustymabe:
force pushed a change to fix quotations.

PR #222, opened 2020-04-09
streams: disable bodhi-updates and next-devel for now

jlebon:
There's more work needed before we can unblock those. For background,
on bodhi-updates, see:

https://github.com/coreos/fedora-coreos-config/pull/335#issuecomment-610634917

For the next-devel stream, see:

https://github.com/coreos/fedora-coreos-tracker/issues/283#issuecomment-610974733


jlebon:
Had to tweak this a bit actually, updated! :arrow_up:

PR #224, opened 2020-04-15
pipeline: handle release index without releases

jlebon:
For example, on `next-devel`, we don't have any releases yet.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -236,10 +236,13 @@ lock(resource: "build-${params.STREAM}") {
             def parent_arg = ""
             if (utils.path_exists("tmp/releases.json")) {
                 def releases = readJSON file: "tmp/releases.json"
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -236,10 +236,12 @@ lock(resource: "build-${params.STREAM}") {
             def parent_arg = ""
             if (utils.path_exists("tmp/releases.json")) {
                 def releases = readJSON file: "tmp/releases.json"
-                def commit_obj = releases["releases"][-1]["commits"].find{ commit -> commit["architecture"] == basearch }
-                parent_commit = commit_obj["checksum"]
-                parent_arg = "--parent ${parent_commit}"
-                parent_version = releases["releases"][-1]["version"]
+                if (releases["releases"].size() > 0) {
```
dustymabe, line 239: optional - add comment here to give more context

PR #225, opened 2020-04-16
Add next stream

jlebon:
Now that we have next-devel setup, let's add the next stream!


jlebon:
Requires: https://github.com/coreos/fedora-coreos-config/pull/353

PR #226, opened 2020-04-20
pipeline: use `cosa build --strict`

jlebon:
We always want to build in strict mode. See:
https://github.com/coreos/fedora-coreos-tracker/issues/454

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -254,7 +254,7 @@ lock(resource: "build-${params.STREAM}") {
                 version = "--version ${new_version}"
             }
             utils.shwrap("""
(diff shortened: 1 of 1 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/1370

dustymabe:
what happens when you pass `--strict` but don't have any lockfiles? i.e. if we try to run bodhi-updates through the pipeline again

jlebon:
> what happens when you pass `--strict` but don't have any lockfiles? i.e. if we try to run bodhi-updates through the pipeline again

It has no effect if no lockfiles are passed. Relatedly, I've improved the help string for the switch in https://github.com/coreos/rpm-ostree/pull/2064.

PR #227, opened 2020-04-21
pipeline: use `cosa fetch --strict`

dustymabe:
We also need to fetch in strict mode. This requires [1]. See that PR
for more information.

[1] https://github.com/coreos/coreos-assembler/pull/1378


jlebon:
LGTM once https://github.com/coreos/coreos-assembler/pull/1378 merges!

PR #229, opened 2020-05-01
jobs: add bump-lockfile

jlebon:
This job will implement lockfile bumping for testing-devel and
next-devel: coreos/fedora-coreos-tracker#293.

The original plan for this functionality was to have it in config-bot:
coreos/fedora-coreos-releng-automation#48

But in the end, I think it's more natural to have it as a Jenkins job
given that it does a lot of the same things as the pipeline/upstream CI
jobs. So that way it looks and feels just like another job that runs
cosa, and we get kola artifacts, we can re-use the shared library,
it's easily inspectable, we can hook it to Slack, etc...

### DIFF
--- a/jenkins/config/github-coreosbot.yaml
+++ b/jenkins/config/github-coreosbot.yaml
@@ -1,0 +1,15 @@
+credentials:
+  system:
+    domainCredentials:
+      - credentials:
+        - usernamePassword:
+            scope: GLOBAL
+            id: github-coreosbot-token
+            username: coreosbot
+            password: ${github-coreosbot-token/token}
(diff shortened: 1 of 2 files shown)


jlebon:
Requires: https://github.com/coreos/fedora-coreos-pipeline/pull/228

Marking as draft, since I haven't tested this at all for now!

jlebon:
OK, this is tested working now! I hit a subtle issue with lockfiles and shim-x64 which I need to follow-up with, though this PR is hacking around that for now.

@dustymabe Want to take a final look before we merge this?

PR #230, opened 2020-05-06
pipeline: add temporary knob to allow upgrade tests to fail

jlebon:
Upgrade tests on the `next` stream are going to fail right now because
of a known regression:

https://github.com/coreos/fedora-coreos-tracker/issues/473

Add a temporary knob to allow continuing the build without it.

I initially tried just blacklisting it, but because it's the only test
in the suite right now, kola errors out due to no tests being available
to run.


sohankunkerkar:
LGTM

PR #233, opened 2020-05-11
Add pipeline to test GCP images

dustymabe:
This adds the Jenkinsfile.kola.gcp file that will allow us to run tests
against GCP images just like we do for AWS images.

### DIFF
--- a/Jenkinsfile.kola.gcp
+++ b/Jenkinsfile.kola.gcp
@@ -1,0 +1,121 @@
+def utils, streams
+node {
+    checkout scm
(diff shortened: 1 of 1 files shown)


dustymabe:
Note that this PR doesn't enable the test to be called by the default pipeline yet. I'll need to debug and make sure everything works first. Since this is a side pipeline I'd like to do it in prod since I currently don't have a secondary pipeline environment to run this in. 

The proposal is that we merge this and then I run it a few times in prod and fix any bugs before we enable it being called from the main Jenkinsfile.

PR #235, opened 2020-05-13
fix variable definition in kola GCP pipeline

dustymabe:


Jenkinsfile.kola.gcp
```diff
@@ -41,7 +41,7 @@ def pod_label = "cosa-${UUID.randomUUID().toString()}"
 
 podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
     node(pod_label) { container('coreos-assembler') {
-        def gcp_image, no_gcp_image
+        def gcp_image
         def no_gcp_image = false
```
jlebon, line 45: I think it'd be cleaner to just get rid of `no_gcp_image` entirely and initialize `gcp_image` to `""` and fail on that.

dustymabe: ok I updated it but I need you to help me to make sure the conditional logic for `""` is correct

Jenkinsfile.kola.gcp
```diff
@@ -61,13 +60,11 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
...
             if (meta.gcp.image) {
                 gcp_image = meta.gcp.image
-            } else {
-                no_gcp_image = true
             }
         }
 
         // fail immediately if the build contained no GCP image
-        if (no_gcp_image) {
+        if (gcp_image = "") {
```
jlebon, line 67:
```suggestion
        if (gcp_image == "") {
```

dustymabe: +1 fixed

PR #236, opened 2020-05-15
Add new sync-stream-metadata job

jlebon:
Add a new job that will automatically sync the fedora-coreos-streams
metadata to S3. This is part of streamlining the release process.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -169,20 +169,28 @@ If you are in production where we upload builds to S3 OR you want to
 test uploading to S3 as part of your pipeline development, you need to
 create a credentials config as a secret within OpenShift.
 
-First create a file with your secret content:
+First create files with your secret content:
 
 ```
-cat <<'EOF' > /path/to/upload-secret
+mkdir dir
+cat <<'EOF' > dir/config
 [default]
 aws_access_key_id=keyid
 aws_secret_access_key=key
 EOF
+echo keyid > dir/accessKey
+echo key > dir/secretKey
 ```
+
+We expose it in different ways (as an AWS config file and as direct
+fields) because those credentials are used both directly as Jenkins
+credentials (which use the direct fields) and by e.g. `cosa`, which
+accepts an AWS config file.
 
 Then create the secret in OpenShift:
 
 ```
-oc create secret generic aws-fcos-builds-bot-config --from-file=config=/path/to/upload-secret
+oc create secret generic aws-fcos-builds-bot-config --from-file=dir
 ```
 
 We also have a second AWS config that can be used for running kola
--- a/jobs/sync-stream-metadata.Jenkinsfile
+++ b/jobs/sync-stream-metadata.Jenkinsfile
@@ -1,0 +1,20 @@
+@Library('github.com/coreos/coreos-ci-lib@master') _
+
+properties([
+    pipelineTriggers([
+        githubPush(),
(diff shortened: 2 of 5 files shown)


dustymabe:
since you're going to be out on monday should we hold this until that testing release is out the door?

jlebon:
> since you're going to be out on monday should we hold this until that testing release is out the door?

Yup, sure. WFM!

jlebon:
Anyone want to review this? No releases this week, so I think we can merge it in time for the releases next week!

jobs/sync-stream-metadata.Jenkinsfile
```diff
@@ -0,0 +1,19 @@
...
+    ])
+])
+
+cosaPod {
+    git(url: 'https://github.com/coreos/fedora-coreos-streams', credentialsId: 'github-coreosbot-token')
+    withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: 'aws-fcos-builds-bot']]) {
+        shwrap("""
+            aws s3 sync --acl public-read --cache-control 'max-age=60' \
+                --exclude '*' --include 'streams/*' --include 'updates/*' \
+                    . s3://fcos-builds
```
dustymabe, line 16: i almost missed the `.` - thoughts on making it `./` ?

HACKING.md
```diff
@@ -172,17 +172,20 @@ create a credentials config as a secret within OpenShift.
...
 ```
-cat <<'EOF' > /path/to/upload-secret
+mkdir dir
+cat <<'EOF' > dir/config
 [default]
 aws_access_key_id=keyid
 aws_secret_access_key=key
 EOF
+echo keyid > dir/accessKey
+echo key > dir/secretKey
```
dustymabe, line 182: To an outside user the change here doesn't really make sense. Can we update the comment to say we're using the creds in multiple ways now? Or is there any way to pick up the creds from the secret we were creating in the past?

jobs/sync-stream-metadata.Jenkinsfile
```diff
@@ -0,0 +1,19 @@
+@Library('github.com/coreos/coreos-ci-lib@master') _
+
+properties([
+    pipelineTriggers([
+        githubPush(),
+        pollSCM('H/15 * * * *')
```
dustymabe, line 6: what does this translate into? "every 15 minutes"?

manifests/jenkins.yaml
```diff
@@ -127,6 +131,11 @@ objects:
           secret:
             secretName: slack-api-token
             optional: true
+        # DELTA: add AWS creds
+        - name: aws-fcos-builds-bot-creds
+          secret:
+            secretName: aws-fcos-builds-bot-config
+            optional: true
```
dustymabe, line 138: should we add a comment here about why we're adding them (i.e. specifically so we can have them for the sync-stream-metadata.Jenkinsfile). Otherwise one might get confused about the source of truth for the configs for the actual build uploads (in pod.yaml)

HACKING.md
jlebon: Added a comment. I think it might work to just drop the config file and have cosa use the env vars, though I haven't tested that. Let's look at that in a follow-up?

jlebon:
Pushed fixes for comments, though still need to test this final version and it looks like the devel Jenkins fell over.

jlebon:
Tested working!

PR #237, opened 2020-05-16
update for GCP image build/release workflow

dustymabe:
In our build/release workflow for GCP we will build and upload/create
a GCP image in the build pipeline. In the build pipeline we will
deprecate the image because we haven't decided to release it yet.
In the release pipeline we will undeprecate the image and deprecate
all other images in that image family using the ore gcloud promote-image
command.

Also attach the image to a license upon creation.


dustymabe:
depends on https://github.com/coreos/coreos-assembler/pull/1456

dustymabe:
Thanks @bgilbert . I tested this pretty well tonight in a separate pipeline (found one missing `}` and some whitespace changes I needed to make). I'm going to merge this so we can use the code in the releases we do tomorrow.

PR #238, opened 2020-05-19
Jenkinsfile.kola.gcp: fix upgrade test

dustymabe:
We were missing a `\` but we also need to specify the `=` in the
--platform calls otherwise cosa kola gets confused.

### DIFF
--- a/Jenkinsfile.kola.gcp
+++ b/Jenkinsfile.kola.gcp
@@ -80,7 +80,7 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
                       -b fcos -j 5 \
                       --no-test-exit-error \
                       --build=${params.VERSION} \
-                      --platform gce \
+                      --platform=gce \
                       --gce-project=\${gcp_project} \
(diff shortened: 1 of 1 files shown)


Jenkinsfile.kola.gcp
```diff
@@ -96,14 +96,13 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
                   gcp_project=\$(jq -r .project_id \${GCP_KOLA_TESTS_CONFIG})
                   # use `cosa kola` here since it knows about blacklisted tests
                   cosa kola run \
-                      -b fcos -j 5 \
```
arithx, line 99: Is dropping the parallelism intentional?

dustymabe: yeah I basically made the arguments match what we're doing for AWS, which may be wrong...

https://github.com/coreos/fedora-coreos-pipeline/blob/ad2579f2a946d582f19f4f93e8131345666e151f/Jenkinsfile.kola.aws#L88

jlebon: There's only one test right now, so there's nothing to parallelize :)

PR #240, opened 2020-05-21
Jenkinsfile.kola.gcp: drop --gce-image

jlebon:
Upgrade tests use the image of the previous release, not the candidate
release we're testing itself. The `--find-parent-image` switch will
automatically find the starting image to use.

Closes: https://github.com/coreos/fedora-coreos-tracker/issues/487

### DIFF
--- a/Jenkinsfile.kola.gcp
+++ b/Jenkinsfile.kola.gcp
@@ -101,7 +101,6 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
                       --no-test-exit-error \
                       --platform=gce \
                       --gce-project=\${gcp_project} \
-                      --gce-image="projects/${gcp_image_project}/global/images/${gcp_image}" \
                       --gce-json-key=\${GCP_KOLA_TESTS_CONFIG}
                   tar -cf - tmp/kola-upgrade | xz -c9 > kola-run-upgrade.tar.xz
                   """)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/1468

dustymabe:
hmm. Note here that the line we are removing is:

```
--gce-image="projects/${gcp_image_project}/global/images/${gcp_image}" \
```

which encodes the project the image belongs to as well. I think we need to hold this, let me revive https://github.com/coreos/coreos-assembler/pull/1335 so we can get project into the meta.json and then update https://github.com/coreos/coreos-assembler/pull/1468 to also consider project.

Do you agree?

jlebon:
> I think we need to hold this, let me revive coreos/coreos-assembler#1335 so we can get project into the meta.json and then update coreos/coreos-assembler#1468 to also consider project.
>
> Do you agree?

Ahh right, the image is not useful without a project. Sure yup, let's do https://github.com/coreos/coreos-assembler/pull/1335. (I guess short-term we could also just hardcode the project name in kola, though https://github.com/coreos/coreos-assembler/pull/1335 looks pretty straightforward.)

dustymabe:
ok https://github.com/coreos/coreos-assembler/pull/1335 and https://github.com/coreos/coreos-assembler/pull/1468 is approved.

just wondering.. do we need to or should we drop `--gce-image` from the other `cosa kola` call?

dustymabe:
as part of this do you want to enable the GCP test in the pipeline too?

dustymabe:
> as part of this do you want to enable the GCP test in the pipeline too?

nvm - let's do it in https://github.com/coreos/fedora-coreos-pipeline/pull/241

jlebon:
> ok coreos/coreos-assembler#1335 merged and coreos/coreos-assembler#1468 is approved.

Awesome, thanks for testing! Let's merge once the latter one is in.

> just wondering.. do we need to or should we drop --gce-image from the other cosa kola call?

The other `cosa kola` call isn't an upgrade test, so we want to bring up the actual image we've just uploaded (though an interesting subject here would be to rerun some of the tests on top of a freshly updated system; e.g. something like `kola run --upgrade-to $URL_TO_OSTREE`).

dustymabe:
> The other `cosa kola` call isn't an upgrade test, so we want to bring up the actual image we've just uploaded

I'm just wondering why kola can find that information (the image/image project) itself since we are passing it the `--build` ?

jlebon:
> I'm just wondering why kola can find that information (the image/image project) itself since we are passing it the `--build` ?

I'm guessing you mean "aren't", right?

`kola` knows to use the latest build if `$PWD` is a cosa workdir.

dustymabe:
Let me start over.. Why can't we remove 

https://github.com/coreos/fedora-coreos-pipeline/blob/85c3ea58e5c5125107cab2f0a75c37d109d9d66a/Jenkinsfile.kola.gcp#L85

jlebon:
> Let me start over.. Why can't we remove

Because right now `kola` doesn't know to just automatically use the GCP image from the build metadata if `--platform=gce` (see how it's done right now for `qemu` at https://github.com/coreos/coreos-assembler/blob/3b691fbedf3ff367098a78bf2a86a169ea44d872/mantle/cmd/kola/options.go#L273). We could easily fix that though, the same way we did it for `--find-parent-image`.

dustymabe:
👍 - thanks. that explains it

PR #242, opened 2020-05-22
Jenkinsfile: attach nested virtualization license to GCP images

dustymabe:
This will allow for launched instances to have access to /dev/kvm
by default. I am not aware of any known drawbacks to doing this as
I've been informed by GCP folks that if there is an environment that
doesn't have nested virt support or if there is a policy that disallows
it the instance simply won't have access to /dev/kvm. No other errors
would be presented to the user (like not being able to launch an
instance).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -448,6 +448,7 @@ lock(resource: "build-${params.STREAM}") {
                         --deprecated \
                         --family fedora-coreos-${params.STREAM} \
                         --license fedora-coreos-${params.STREAM} \
+                        --license "https://compute.googleapis.com/compute/v1/projects/vm-options/global/licenses/enable-vmx" \
                         --project=\${gcp_project} \
                         --bucket gs://${gcp_gs_bucket}/image-import \
                         --json \${GCP_IMAGE_UPLOAD_CONFIG} \


dustymabe:
requires https://github.com/coreos/coreos-assembler/pull/1477

zmarano:
LGTM. This is currently the way to allow nested KVM to work.

dustymabe:
A little more context from @zmarano over in https://github.com/coreos/coreos-assembler/pull/1477#issuecomment-634275570

> From the GCE point of view, enabling nested virt on an image by default (currently via the license resource addition) is perfectly fine for your use case. It will not change behaviors for users except to expose /dev/kvm (its synonymous to turning on the Intel-VTx setting on a physical machine in the firmware). That said, there isn't any guarantee that a given guest VM within the nested virt host VM will work or that the user is using a machine type that supports it. The matrix of possibilities there is enormous. It is therefore similar to saying your OS supports KVM on the hardware it is running on and by default allowing your users to use KVM on GCE VM's. I know this is a bit wishy-washy but let me know if I can clarify any of these points.
> 
> This is the [doc](https://cloud.google.com/compute/docs/instances/enable-nested-virtualization-vm-instances) on this topic and it is largely saying the same things in a different way.

PR #245, opened 2020-05-26
bump-lockfile: fix variable substitution for auth

jlebon:
The credential bindings bind to environment variables that e.g. `bash`
can access. So here we need to escape the variable references to make
sure it's `bash` that dereferences them. Oddly, this somehow did work at
first, so I think it did/does sometimes also bind to Groovy variables.


jlebon:
This should unblock testing-devel and next-devel getting new content.

dustymabe:
> Oddly, this somehow did work at
> first, so I think it did/does sometimes also bind to Groovy variables.

odd

PR #247, opened 2020-06-03
pipeline: pass --parent-build to cosa

jlebon:
Instead of passing the parent commit directory to cosa, pass the parent
build ID itself. This will allow us to get a diff against the parent
build, which we can then use on the release notes pages.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -209,14 +209,31 @@ lock(resource: "build-${params.STREAM}") {
             """)
         }
 
+        def parent_version = ""
+        def parent_commit = ""
         stage('Fetch') {
             if (s3_stream_dir) {
+                utils.aws_s3_cp_allow_noent("s3://${s3_stream_dir}/releases.json", "tmp/releases.json")
+                if (utils.path_exists("tmp/releases.json")) {
+                    def releases = readJSON file: "tmp/releases.json"
+                    // check if there's a previous release we should use as parent
+                    if (releases["releases"].size() > 0) {
+                        def commit_obj = releases["releases"][-1]["commits"].find{ commit -> commit["architecture"] == basearch }
+                        parent_commit = commit_obj["checksum"]
+                        parent_version = releases["releases"][-1]["version"]
+                    }
+                }
+
                 utils.shwrap("""
                 export AWS_CONFIG_FILE=\${AWS_FCOS_BUILDS_BOT_CONFIG}
                 cosa buildprep s3://${s3_stream_dir}/builds
                 """)
-                // also fetch releases.json to get the latest build, but don't error if it doesn't exist
-                utils.aws_s3_cp_allow_noent("s3://${s3_stream_dir}/releases.json", "tmp/releases.json")
+                if (parent_version != "") {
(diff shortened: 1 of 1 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/1507

zonggen:
After --parent-build logic has landed completely, does it make sense for website to support both methods? i.e. check for `parent-pkgdiff` field in metadata, if no such field presents, then calculate the pkgdiff within the browser. 

Or wait for 5 releases to make sure `parent-pkgdiff` is available for all displayed releases and remove the front-end calculating part completely.

Jenkinsfile
```diff
@@ -209,14 +209,31 @@ lock(resource: "build-${params.STREAM}") {
...
                 utils.shwrap("""
                 export AWS_CONFIG_FILE=\${AWS_FCOS_BUILDS_BOT_CONFIG}
                 cosa buildprep s3://${s3_stream_dir}/builds
                 """)
-                // also fetch releases.json to get the latest build, but don't error if it doesn't exist
-                utils.aws_s3_cp_allow_noent("s3://${s3_stream_dir}/releases.json", "tmp/releases.json")
+                if (parent_version != "") {
+                    // also fetch the parent version; this is used by cosa to do the diff
+                    utils.shwrap("""
+                    cosa buildprep s3://${s3_stream_dir}/builds --build ${parent_version}
```
dustymabe, line 234: does this just grab the metadata? i.e. it shouldn't be a lot of data to download.

dustymabe:
> After --parent-build logic has landed completely, does it make sense for website to support both methods? i.e. check for `parent-pkgdiff` field in metadata, if no such field presents, then calculate the pkgdiff within the browser.

I guess it depends on how much work it is. If it's not a lot of work then supported both would make for an easier transition.

> 
> _Or_ wait for 5 releases to make sure `parent-pkgdiff` is available for all displayed releases and remove the front-end calculating part completely.

If we went with this approach I don't think we'd have to wait that long. Just wait for 2 releases to come out (i.e., 6 weeks from now) and then go ahead and deploy it. Having the diff for only 2 would be acceptable in my opinion and then the other ones will cycle out.

Jenkinsfile
jlebon: Yup, exactly. Just the build metadata and the commit object, which houses the full RPM list that `rpm-ostree db diff` knows how to read.

zonggen:
> I guess it depends on how much work it is. If it's not a lot of work then supported both would make for an easier transition.

Not a lot of work actually, it's a matter of adding a check condition. Just wanted to discuss possibilities on future website changes. Will proceed with supporting both then :)

jlebon:
> Not a lot of work actually, it's a matter of adding a check condition. Just wanted to discuss possibilities on future website changes. Will proceed with supporting both then :)

Yeah, if we can support both easily let's do that, and then nuke the old one in a few releases.

To help testing, here's a sample `meta.json` of what the latest `testing` release would've been if it had `parent-pkgdiff`: https://gist.github.com/jlebon/a8eaf409a266679f07fbbbfae50f75a3.

PR #250, opened 2020-06-05
pipeline: generate liveiso after metal4k

jlebon:
We need the metal4k image in place before generating the ISO so that we
can `osmet pack` it. Otherwise understandably, the `iso-offline-install`
test for `metal4k` will fail. We should probably add a switch that makes
it required to `buildextend-live`.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -342,15 +342,15 @@ lock(resource: "build-${params.STREAM}") {
                 """)
             }
 
-            stage('Build Live') {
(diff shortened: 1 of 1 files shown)


cgwalters:
Random aside, I was thinking a bit more lately about https://lwn.net/Articles/821367/ and related things, and might be interesting to try to represent our build as ninja rules?

dustymabe:
🤣 - sometimes you just have to laugh!

PR #254, opened 2020-07-09
jobs: pin to coreos-ci-lib for OCP3 for now

jlebon:
I'd like to have CoreOS CI exercise the new cluster for a bit before
moving the production pipeline itself over. For now, let's pin the
version of coreos-ci-lib we use here which still uses oci-kvm-hook
instead of kubevirt.


jlebon:
See https://github.com/coreos/coreos-ci-lib/pull/31. I guess we could also make it configurable, though I'm hoping to move everything over pretty soon, and this is less churn.

PR #256, opened 2020-07-15
pipeline: ignore failures from pxe-install scenario

jlebon:
This test is flaky right now. See
https://github.com/coreos/coreos-assembler/issues/1597.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -374,7 +374,10 @@ lock(resource: "build-${params.STREAM}") {
                 """)
                 try {
                     parallel metal: {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -374,7 +374,10 @@ lock(resource: "build-${params.STREAM}") {
                 """)
                 try {
                     parallel metal: {
-                        utils.shwrap("kola testiso -S --output-dir tmp/kola-metal")
+                        utils.shwrap("kola testiso -SP --output-dir tmp/kola-metal")
```
dustymabe, line 377: tangent: I kind of wish we didn't have short options here.. `--no-pxe` would be easier to process when reviewing

PR #258, opened 2020-07-20
pipeline: build legacy PXE images for now

bgilbert:
Generate a PXE rootfs image that only contains a flag file for alerting purposes.  We'll switch to a real separate rootfs image after a deprecation period.

For https://github.com/coreos/fedora-coreos-tracker/issues/390.  Must be merged at the same time as https://github.com/coreos/coreos-assembler/pull/1608.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -359,7 +359,7 @@ lock(resource: "build-${params.STREAM}") {
 
             stage('Build Live') {
                 utils.shwrap("""
-                cosa buildextend-live
+                cosa buildextend-live --legacy-pxe
                 """)
             }
 


dustymabe:
If we want to go ahead and do this for the next `next` then we'd just add some conditional logic in here to make that happen right?

bgilbert:
Right.

Ready for review.

jlebon:
> Must be merged at the same time as coreos/coreos-assembler#1608.

Note for the future: we should merge pipeline changes which require cosa changes once a cosa image is available with that change. If we merge the two at the same time, it's unlikely going to be rebuilt in time and the pipeline could go red. (Or as what happened here; the Quay.io build died on a Fedora registry flake -- restarted it!)

dustymabe:
It's eventually consistent :rofl:

bgilbert:
> Note for the future: we should merge pipeline changes which require cosa changes once a cosa image is available with that change.

Noted.  Though in this case, it's probably better to get a failed build than one with incorrect artifacts.

PR #261, opened 2020-08-11
fcos-release-notes: add job for generating release notes

zonggen:
Adds a job for generating `release-notes.yaml` from `release-notes.d/` in `testing-devel` and `next-devel` and puts the output `release-notes.yaml` file to `testing` and `next` branch correspondingly.

Also adds the functionality in `Jenkinsfile` such that after building the image, replaces the premature `next-release` placeholder key with the real build id. Then it uploads the final `release-notes.yaml` to the S3 bucket as well as pushing to `fedora-coreos-config` repo.

Related: https://github.com/coreos/fedora-coreos-tracker/issues/194
Signed-off-by: Allen Bai <abai@redhat.com>

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -500,6 +500,38 @@ lock(resource: "build-${params.STREAM}") {
             /var/tmp/fcos-releng/coreos-meta-translator/trans.py --workdir .
             """)
 
+            // Replace the 'next-release' field in 'release-notes.yaml' field
+            // with the real build id
+            botCreds = "github-coreosbot-token"
+            config_repo = "coreos/fedora-coreos-config"
+            release_notes_yaml = "src/config/release-notes.yaml"
+            utils.shwrap("""
+            /var/tmp/fcos-releng/coreos-release-note-generator/release-note-generator.py update \
+            --release-notes-file ${release_notes_yaml} \
+            --build-id ${buildID} \
+            --output-dir src/config
+            """)
+            shwrap("""
+            cd src/config
+            git commit -am "release-notes.yaml: replace 'next-release' with build id"
+            """)
+            withCredentials([usernamePassword(credentialsId: botCreds,
+                                                usernameVariable: 'GHUSER',
+                                                passwordVariable: 'GHTOKEN')]) {
+                // should gracefully handle race conditions here
+                sh("git push https://\${GHUSER}:\${GHTOKEN}@github.com/${config_repo} ${params.STREAM}")
+            }
+
+            // Push 'release-notes.yaml' to S3 bucket for website consumption
+            withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: 'aws-fcos-builds-bot']]) {
+                shwrap("""
+                cd src/config
+                aws s3 sync --acl public-read --cache-control 'max-age=60' \
+                    --exclude '*' --include 'release-notes.yaml' \
+                        ./ s3://fcos-builds
+                """)
+            }
+
             if (s3_stream_dir) {
               // just upload as public-read for now, but see discussions in
               // https://github.com/coreos/fedora-coreos-tracker/issues/189
--- a/jobs/generate-release-notes.Jenkinsfile
+++ b/jobs/generate-release-notes.Jenkinsfile
@@ -1,0 +1,70 @@
+@Library('github.com/coreos/coreos-ci-lib@master') _
+
+config_repo = "coreos/fedora-coreos-config"
+releng_script_repo = "coreos/fedora-coreos-releng-automation"
+branches = [
+    "testing-devel"
+    "next-devel"
+]
+botCreds = "github-coreosbot-token"
+
+cosaPod {
+    // set up fedora-coreos-releng-automation repository
+    shwrap("""
+    mkdir fcos-releng
+    git clone --branch master https://github.com/${releng_script_repo} fcos-releng"
+    git config --global user.name "CoreOS Bot"
+    git config --global user.email "coreosbot@fedoraproject.org"
+    """)
+
+    parallel branches.collectEntries { branch -> [branch, {
+        // build 'release-notes.yaml' using 'release-notes.d/' under testing-devel and next-devel
+        stage("Build 'release-notes.yaml'") {
+            shwrap("mkdir ${branch}")
+
+            dir(branch) {
+                shwrap("cosa init --branch ${branch} https://github.com/${config_repo}")
+            }
+
+            // generate 'release-notes.yaml' from yaml snippets under 'config/release-notes.d/'
+            dir("fcos-releng") {
+                shwrap("python3 coreos-release-note-generator/release-note-generator.py build \
+                    --config-dir ~/${branch}/src/config \
+                    --output-dir ~/${branch}/src/config")
+            }
+        }
+
+        // delete all 'release-notes.d/*.yaml' files for next release version
+        stage("Clean up 'release-notes.d/'") {
+            shwrap("""
+            cd ${branch}/src/config
+            git checkout ${branch}
+            rm -rf release-notes.d/*.yaml
+            git commit -am "release-notes.d: clean up release-notes.d for next release"
+            """)
+            withCredentials([usernamePassword(credentialsId: botCreds,
+                                                usernameVariable: 'GHUSER',
+                                                passwordVariable: 'GHTOKEN')]) {
+                // should gracefully handle race conditions here
+                sh("git push https://\${GHUSER}:\${GHTOKEN}@github.com/${config_repo} ${branch}")
+            }
+        }
+
+        // push 'release-notes.yaml' from testing-devel and next-devel to testing and next branch correspondingly
+        stage("Push 'release-notes.yaml'") {
+            target_branch = branch.replace("-devel", "")
+            shwrap("""
+            cd ${branch}/src/config
+            git checkout ${target_branch}
+            git add release-notes.yaml
+            git commit -am "release-notes.yaml: generate latest release notes"
+            """)
+            withCredentials([usernamePassword(credentialsId: botCreds,
+                                                usernameVariable: 'GHUSER',
+                                                passwordVariable: 'GHTOKEN')]) {
+                // should gracefully handle race conditions here
+                sh("git push https://\${GHUSER}:\${GHTOKEN}@github.com/${config_repo} ${target_branch}")
+            }
+        }
+    }] }
+}


Jenkinsfile
```diff
@@ -500,6 +500,38 @@ lock(resource: "build-${params.STREAM}") {
...
+            // Push 'release-notes.yaml' to S3 bucket for website consumption
+            withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: 'aws-fcos-builds-bot']]) {
+                shwrap("""
+                cd src/config
+                aws s3 sync --acl public-read --cache-control 'max-age=60' \
+                    --exclude '*' --include 'release-notes.yaml' \
+                        ./ s3://fcos-builds
+                """)
+            }
+
```
zonggen, line 534: Something in my mind keeps telling me this is not correct / desired. Since `Jenkinsfile` primarily used _only_ used `cosa`, and not sure if I could use `botCreds` and `aws-fcos-builds-bot` here. The logic comes from `bump-lockfile.Jenkinsfile` and `sync-stream-metadata.Jenkinsfile`.

Could you sanity check please? @jlebon

zonggen:
The PR makes it possible to:
 - Run the `generate-release-notes.Jenkinsfile` to build a pre-mature `release-notes.yaml` with the field `next-release` as the placeholder before we know the real build id ; also clean up the `release-notes.d/` for next release cycle
 - Run the `Jenkinsfile` to replace the `next-release` field with actual build id and push the finalized `release-notes.yaml` to S3 bucket for website consumption

jobs/generate-release-notes.Jenkinsfile
```diff
@@ -0,0 +1,70 @@
...
+    git config --global user.email "coreosbot@fedoraproject.org"
+    """)
+
+    parallel branches.collectEntries { branch -> [branch, {
+        // build 'release-notes.yaml' using 'release-notes.d/' under testing-devel and next-devel
+        stage("Build 'release-notes.yaml'") {
+            shwrap("mkdir ${branch}")
+
+            dir(branch) {
+                shwrap("cosa init --branch ${branch} https://github.com/${config_repo}")
```
dustymabe, line 26: Do we need to use `cosa` here? Should we just use `git clone` instead?

jobs/generate-release-notes.Jenkinsfile
```diff
@@ -0,0 +1,70 @@
...
+                    --output-dir ~/${branch}/src/config")
+            }
+        }
+
+        // delete all 'release-notes.d/*.yaml' files for next release version
+        stage("Clean up 'release-notes.d/'") {
+            shwrap("""
+            cd ${branch}/src/config
+            git checkout ${branch}
+            rm -rf release-notes.d/*.yaml
```
dustymabe, line 42: we probably need to use `git rm` here and `git commit -m` below.

jobs/generate-release-notes.Jenkinsfile
```diff
@@ -0,0 +1,70 @@
...
+            git commit -am "release-notes.d: clean up release-notes.d for next release"
+            """)
+            withCredentials([usernamePassword(credentialsId: botCreds,
+                                                usernameVariable: 'GHUSER',
+                                                passwordVariable: 'GHTOKEN')]) {
+                // should gracefully handle race conditions here
+                sh("git push https://\${GHUSER}:\${GHTOKEN}@github.com/${config_repo} ${branch}")
```
dustymabe, line 49: doing a `git pull --rebase` before doing the `git push` might help with race conditions.

jobs/generate-release-notes.Jenkinsfile
zonggen: All I needed was a FCOS config repo so no hard need for a `cosa init`.

dustymabe:
In the most recent [design proposal](https://github.com/coreos/fedora-coreos-tracker/issues/194#issuecomment-669524131) we have (snippet):

----

### Proposal:

Have a tool that runs independently of the pipeline like `bump-lockfile` and `sync-stream-metadata` jenkins jobs. This tool will trigger based on a new release of FCOS or can be triggered manually. The responsibility of this tool includes:

----


In this PR the "tool" is `jobs/generate-release-notes.Jenkinsfile`, which is great! However, shouldn't it be able to run independently without the need for the `Jenkinsfile` code in this PR that runs `release-note-generator.py update`? 

If `jobs/generate-release-notes.Jenkinsfile` basically polls the release metadata to see what the latest release is it should be able to run completely independent of the build pipeline.

dustymabe:
For `jobs/generate-release-notes.Jenkinsfile` are we sticking by the [design proposal](https://github.com/coreos/fedora-coreos-tracker/issues/194#issuecomment-669524131) (snippet):

----

#### Git repository manipulation
- look at the `release-notes.yaml` file for the related branch (i.e. `testing`)
    - if an entry for that release exists, it does nothing to the git repo
    - if an entry for that release does not exist
        - create a new entry in `release-notes.yaml` for that release version
            - merge the `release-notes.d/*` files into `release-notes.yaml`
            - delete `release-notes.d/*` files from that branch
            - delete corresponding `release-notes.d/*` files from `testing-devel` branch
----

It seems like we can do these three bullet points:

```
- merge the `release-notes.d/*` files into `release-notes.yaml`
- delete `release-notes.d/*` files from that branch
- delete corresponding `release-notes.d/*` files from `testing-devel` branch
```

in a single stage and not worry about breaking it up further. From those bullets it seems like there should be 2 `git push` operations.

- one git push to `testing` with updated `release-notes.yaml` and deleted `release-notes.d/*` files
- one git push to `testing-devel` with deleted `release-notes.d/*` files
	- We have to be careful with this one. We need to make sure we only delete `release-notes.d/` files that were in the `testing` branch


We really are just primarily operating on the `testing` branch, the only reason we touch `testing-devel` is so we don't have to manually clean up those `release-notes.d/` files.

jobs/generate-release-notes.Jenkinsfile
dustymabe: Yeah, let's just use `git clone --depth=1` to keep things simple then.

zonggen:
> jobs/generate-release-notes.Jenkinsfile basically polls the release metadata to see what the latest release

@dustymabe Yes, the only portion that I depend on `Jenkinsfile` was the retrieving the `build id` part, so this should be fixing the current issue :+1: 

> We really are just primarily operating on the testing branch, the only reason we touch testing-devel is so we don't have to manually clean up those release-notes.d/ files.

I thought we mainly operate on `release-notes.d` in `testing-devel` and generate the corresponding `release-notes.yaml` in `testing` so there would be no `release-notes.d` in `testing` branch. So my question is, what is the relationship between `release-notes.d` in `testing-devel` and `release-notes.d` in `testing` since according from the comment

> We need to make sure we only delete release-notes.d/ files that were in the testing branch

it seems like `release-notes.d` in `testing-devel` is the superset of that in `testing` branch.

dustymabe:
> > jobs/generate-release-notes.Jenkinsfile basically polls the release metadata to see what the latest release
> 
> @dustymabe Yes, the only portion that I depend on `Jenkinsfile` was the retrieving the `build id` part, so this should be fixing the current issue +1

👍

> 
> > We really are just primarily operating on the testing branch, the only reason we touch testing-devel is so we don't have to manually clean up those release-notes.d/ files.
> 
> I thought we mainly operate on `release-notes.d` in `testing-devel` and generate the corresponding `release-notes.yaml` in `testing` so there would be no `release-notes.d` in `testing` branch. So my question is, what is the relationship between `release-notes.d` in `testing-devel` and `release-notes.d` in `testing` since according from the comment

The `release-notes.d/` files from `testing-devel` will get copied into `testing` when the release person [runs the promote-config.sh script](https://github.com/coreos/fedora-coreos-streams/blob/master/ISSUE_TEMPLATE/testing.md#promote-testing-devel-changes-to-testing). So we'll be primarily operating on the `testing` branch.

Though, now that I think about it, we might need to leave the `release-notes.d/` in the `testing` branch even after we've processed them, so that they can be used later for the `stable` branch. We'd just want to make sure any processed files get removed from `testing-devel`. We'd just need the `promote-config.sh` to clean up all the `release-notes.d/` files before promoting content to the `stable`, `testing`, `next` branches. If this doesn't make sense let's grab some time tomorrow and talk it through.

> 
> > We need to make sure we only delete release-notes.d/ files that were in the testing branch
> 
> it seems like `release-notes.d` in `testing-devel` is the superset of that in `testing` branch.

Right. Basically `testing-devel` continuously gets new PRs, so a new PR could have come in since we promoted content to `testing` (for the next release). We'd want to make sure that we only remove `release-notes.d/` files that we actually made it into a release and was processed by our tools.

zonggen:
> we might need to leave the release-notes.d/ in the testing branch even after we've processed them, so that they can be used later for the stable branch

That makes a lot of sense. I was concerning about `testing` -> `stable` transition and `promote-config.sh` seems like the piece of information I was missing.

> new PR could have come in since we promoted content to testing (for the next release). We'd want to make sure that we only remove release-notes.d/ files that we actually made it into a release and was processed by our tools.

I thought twice about the process, and it makes sense now. Thanks for explaining!

zonggen:
From https://github.com/coreos/fedora-coreos-config/pull/550#discussion_r472494658 I've reorganized the logic to reflect the workflow without a `next-release` field:

```
# Before promotion, engineers add yaml snippets under release-notes.d in testing-devel
testing-devel:
  release-notes.d/1.yaml
  release-notes.d/2.yaml
testing:
  release-notes.d/

stable:
  release-notes.d/

---

# During promotion from testing-devel -> testing stream, 1.yaml and 2.yaml would be moved to testing branch and removed from testing-devel
# New release-notes.yaml is also generated / updated
  release-notes.d/
testing:
  release-notes.d/1.yaml
  release-notes.d/2.yaml
  release-notes.yaml
stable:
  release-notes.d/
 
---

# During promotion from testing -> stable stream, 1.yaml and 2.yaml would be moved to stable branch and removed from testing
# New release-notes.yaml is also generated / updated
# We no longer need to aggregate yaml snippets in stable, since we get future release notes from testing branch and directly generate them with latest release id. Remove yaml files under release-notes.d/, though in reality we probably should just not move the yaml files to stable in the first place.
testing-devel
  release-notes.d/
testing:
  release-notes.d/
  release-notes.yaml
stable:
  release-notes.d/1.yaml (removed)
  release-notes.d/2.yaml (removed)
  release-notes.yaml
```

_Note_: all of the `release-notes.yaml` will be updated with latest release id.

dustymabe:
> From [coreos/fedora-coreos-config#550 (comment)](https://github.com/coreos/fedora-coreos-config/pull/550#discussion_r472494658) I've reorganized the logic to reflect the workflow without a `next-release` field:
> 
> ```
> # Before promotion, engineers add yaml snippets under release-notes.d in testing-devel
> testing-devel:
>   release-notes.d/1.yaml
>   release-notes.d/2.yaml
> testing:
>   release-notes.d/
> 
> stable:
>   release-notes.d/
> ```

Looks good.

> ```
> 
> # During promotion from testing-devel -> testing stream, 1.yaml and 2.yaml would be moved to testing branch and removed from testing-devel
> # New release-notes.yaml is also generated / updated
> testing-devel:
>   release-notes.d/
> testing:
>   release-notes.d/1.yaml
>   release-notes.d/2.yaml
>   release-notes.yaml
> stable:
>   release-notes.d/
>  
> ```

Looks good

> ```
> # During promotion from testing -> stable stream, 1.yaml and 2.yaml would be moved to stable branch and removed from testing
> # New release-notes.yaml is also generated / updated
> # We no longer need to aggregate yaml snippets in stable, since we get future release notes from testing branch and directly generate them with latest release id. Remove yaml files under release-notes.d/, though in reality we probably should just not move the yaml files to stable in the first place.
> testing-devel
>   release-notes.d/
> testing:
>   release-notes.d/
>   release-notes.yaml
> stable:
>   release-notes.d/1.yaml (removed)
>   release-notes.d/2.yaml (removed)
>   release-notes.yaml
> ```

I'm a bit torn on the "we no longer need to aggregate yaml snippets in stable" bit. I think it might more simple to just generate them the same way as we do for `testing`, but I do see value in just taking the release notes directly from the corresponding release in `testing` because it's possible we've gone in and fixed up formatting or added new information.

Though the more I think about it, we often do multiple `testing` releases before we do a stable. I think the snippets approach might be the way to go.


> 
> _Note_: all of the `release-notes.yaml` will be updated with latest release id.


I think there are a few things that are important to call out to see if we are on the same page.

1. For `testing-devel` and `next-devel` `release-notes.d/{1,2}.yaml` files should get removed by the release-notes-generator job
2. For `stable`, `testing`, and `next` the `release-notes.d/{1,2}.yaml` files should only get removed by the `promote-config.sh` process

Does that match with what you are thinking?

Jenkinsfile
jlebon: Still trying to understand how everything fits together (so the next statements might not fit with the intended design), but ISTM like we shouldn't do anything related to release notes as part of the main build pipeline. That should instead be in the release job (or a job triggered by the release job), and only on production streams.

dustymabe: Agreed. Discussed at https://github.com/coreos/fedora-coreos-pipeline/pull/261#issuecomment-675104281.

PR #263, opened 2020-08-24
Move pipeline to new OCP4 cluster

jlebon:
```
commit cc4b2095febdfdc697b8bffe86a68421126791eb
Date:   Mon Aug 24 09:44:59 2020 -0400

    jenkins: rebase to latest OCP4 template

    And as usual, keep marking all the deltas with `DELTA`.

commit 9564c1a289d2044593a5aec9dac2e052d599c6d1
Date:   Mon Aug 24 09:45:00 2020 -0400

    pipeline: adapt to new cluster URL and version

    Update the Jenkins URL and cluster URL to the new cluster, move the jobs
    which use coreos-ci-lib to go back to the master branch, and change the
    default KVM selector to kvm-device-plugin. (We'll rip out that switch
    altogether in a following patch).

commit 516091e19a778d18bd84f1d19e4c24049d1e4df3
Date:   Mon Aug 24 09:45:01 2020 -0400

    plugins: bump configuration-as-code to 1.35

    Otherwise we get version dependency issues with the other new plugins.

commit dd744d540f4bb4fa5077a20ab7563e1df6352bd1
Date:   Mon Aug 24 09:45:02 2020 -0400

    jenkins: switch to in-cluster Jenkins imagestream

    A nice change in OCP4 is that the S2I Jenkins image meant to match the
    cluster version is tracked by the cluster itself via an ImageStream.
    Let's make use of it instead of doing our own tracking.

commit 06a587e51f5267509f730884a30ec5900d7ae9f6
Date:   Mon Aug 24 09:45:03 2020 -0400

    pipeline: drop support for oci-kvm-hook

    Now that we're on OCP4, let's only support the kubevirt way of things.

commit 9c36085f257288d0faf8fc5f14b0c5b3ab9b325d
Date:   Mon Aug 24 09:45:04 2020 -0400

    HACKING: drop `oc cluster up` path

    It doesn't exist anymore in OCP4. Direct people to CRC or the installer
    and the OpenShift CNV docs.

```

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -17,7 +17,7 @@ In the following sections, the section header may indicate whether the
 section applies to the local cluster case (`[LOCAL]`) or the official
 prod case (`[PROD]`).
 
-You'll want to be sure you have KVM available in your cluster.  See
+You'll want to be sure you have kubevirt available in your cluster.  See
 [this section of the coreos-assembler docs](https://github.com/coreos/coreos-assembler/blob/master/README.md#getting-started---prerequisites).
 
 ### Using a production OpenShift cluster
@@ -306,9 +306,6 @@ circumstances. Here are some of them:
     - Git source URL and optional git ref for pipeline Jenkinsfile.
 - `--config <URL>[@REF]`
     - Git source URL and optional git ref for FCOS config.
-- `--kvm-selector=kvm-device-plugin`:
-    - Use this if you're using the KVM device plugin (modern
-      Kubernetes/OpenShift 4+).
 - `--pvc-size <SIZE>`
     - Size of the cache PVC to create. Note that the PVC size cannot be
       changed after creation. The format is the one understood by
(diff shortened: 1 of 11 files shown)


jlebon:
Let's hold this until we've gone over our migration strategy.

This [works fine in the new cluster](https://jenkins-fedora-coreos.apps.ocp.ci.centos.org/job/fedora-coreos/job/fedora-coreos-fedora-coreos-pipeline/) (builds error out in the end on missing credentials because I didn't add them so as not to conflict with the existing cluster).

For migrating, here's what I'm currently thinking:
1. get releases for this week out using old cluster
2. keep an `ocp3` branch around at current git master
3. scale Jenkins on old cluster down to 0
4. merge this PR
5. deploy onto new cluster and sanity-check everything works fine
6. cleanup old cluster and branch

If something goes wrong/becomes a blocker, we can scale down on the new cluster and scale back up the old cluster pointing at the old branch while we fix it.

dustymabe:
I wonder if I should try this out on my OKD cluster tomorrow just to give it a spin in a different environment.

jlebon:
> I wonder if I should try this out on my OKD cluster tomorrow just to give it a spin in a different environment.

Sure thing! Would be a nice sanity-check. I was trying to test this with CRC locally, and while CRC itself worked well, I simply didn't have enough RAM to get the full pipeline up.

PR #264, opened 2020-08-24
Jenkins.release: error out if VERSION is not provided

jlebon:
Let's be explicit and require that users/machines specify which version
to release instead of having a dangerous/racy "use latest" fallback.

### DIFF
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -36,6 +36,12 @@ properties([
     ])
 ])
 
+// no way to make a parameter required directly so manually check
+// https://issues.jenkins-ci.org/browse/JENKINS-3509
+if (params.VERSION == "") {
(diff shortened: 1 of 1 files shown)


Jenkinsfile.release
```diff
@@ -36,6 +36,10 @@ properties([
     ])
 ])
 
+if (params.VERSION == "") {
+    throw new Exception("Missing VERSION parameter!")
+}
+
```
dustymabe, line 42: Is there a way to make it required in the properties above instead of doing the check here?

jlebon: Heh, yeah I was looking at that, and it led me to https://issues.jenkins-ci.org/browse/JENKINS-3509.

dustymabe: Nice.. comment in the code here?

PR #270, opened 2020-08-26
Convert kola test jobs to vanilla pipeline jobs

jlebon:
This is part of the push to move away from buildconfig-based jobs
towards the same model as what upstream CI does. For more details, see:

https://github.com/coreos/fedora-coreos-pipeline/pull/228

As a first victim, convert the kola tests since they're mostly
straight-forward. A huge part of this is re-using coreos-ci-lib the same
way upstream CI does. And this helps save us a lot of boilerplate (see
diffstat for this patch).

Converting to vanilla jobs also means we can now use the `build` step as
intended, instead of hacking around it with `oc start-build`.


jlebon:
Requires: https://github.com/coreos/coreos-ci-lib/pull/36
Requires: https://github.com/coreos/coreos-ci-lib/pull/37

jlebon:
(I'll admit I didn't test this. Since it's not the main pipeline itself, was thinking of just doing fixups as needed on top.)

jobs/kola-gcp.Jenkinsfile
```diff
@@ -0,0 +1,66 @@
...
+cosaPod(image: params.COREOS_ASSEMBLER_IMAGE, memory: "256Mi",
+        secrets: ["aws-fcos-builds-bot-config", "gcp-kola-tests-config"]) {
+
+    def gcp_image, gcp_image_project, gcp_project
+    stage('Fetch Metadata') {
+        shwrap("""
+        export AWS_CONFIG_FILE=/run/kubernetes/secrets/aws-fcos-builds-bot-config/config
+        cosa init --branch ${params.STREAM} https://github.com/coreos/fedora-coreos-config
+        cosa buildprep --ostree --build=${params.VERSION} s3://${s3_stream_dir}/builds
+        """)
```
miabbott, line 48: maybe in the future this can be a function in coreos-ci-lib that can be reused

miabbott:
Looks pretty straight-forward to me.  :+1:

jobs/kola-gcp.Jenkinsfile
jlebon: Yeah, makes sense. Maybe an `fcosBuildPrep` or something? I think the f-c-c CI could use this as well.

Jenkinsfile
```diff
@@ -537,35 +537,25 @@ lock(resource: "build-${params.STREAM}") {
...
-                    utils.shwrap("""
-                        # We consider the GCP kola tests to be a followup job
-                        # so we aren't adding a `--wait` here.
-                        oc start-build fedora-coreos-pipeline-kola-gcp \
-                            -e STREAM=${params.STREAM} \
-                            -e VERSION=${newBuildID} \
-                            -e S3_STREAM_DIR=${s3_stream_dir}
-                    """)
-                }
+                // We consider the AWS kola tests to be a followup job, so we use `wait: false` here.
```
dustymabe, line 553:
```suggestion
                // We consider the GCP kola tests to be a followup job, so we use `wait: false` here.
```

jobs/kola-aws.Jenkinsfile
```diff
@@ -0,0 +1,62 @@
...
+      choice(name: 'STREAM',
+             // list devel first so that it's the default choice
+             choices: (streams.development + streams.production + streams.mechanical),
+             description: 'Fedora CoreOS stream to test'),
+      string(name: 'VERSION',
+             description: 'Fedora CoreOS Build ID to test',
+             defaultValue: '',
+             trim: true),
+      string(name: 'S3_STREAM_DIR',
+             description: 'Fedora CoreOS S3 Stream Directory',
```
dustymabe, line 22:
```suggestion
             description: 'Optional: Override the Fedora CoreOS S3 Stream Directory',
```

jobs/kola-aws.Jenkinsfile
```diff
@@ -0,0 +1,62 @@
...
+    s3_stream_dir = "fcos-builds/prod/streams/${params.STREAM}"
+}
+
+cosaPod(image: params.COREOS_ASSEMBLER_IMAGE, memory: "256Mi",
+        secrets: ["aws-fcos-builds-bot-config", "aws-fcos-kola-bot-config"]) {
+
+    def ami, ami_region
+    stage('Fetch Metadata') {
+        shwrap("""
+        export AWS_CONFIG_FILE=/run/kubernetes/secrets/aws-fcos-builds-bot-config/config
```
dustymabe, line 45:
```suggestion
        export AWS_CONFIG_FILE=\${AWS_FCOS_BUILDS_BOT_CONFIG}
```

Is what we had in the past. Shall we keep it the same in case something changes we won't need to update this code, just the variable.

jobs/kola-aws.Jenkinsfile
dustymabe: Same suggestion in GCP file

jobs/kola-aws.Jenkinsfile
dustymabe: Same suggestion in GCP file

jobs/kola-gcp.Jenkinsfile
```diff
@@ -0,0 +1,66 @@
...
+            throw new Exception("No GCP image found in metadata for ${params.VERSION}")
+        }
+
+        // pick up the project to use from the config
+        def config = readJSON file: "/run/kubernetes/secrets/gcp-kola-tests-config/config"
+        gcp_project = config.project_id
+    }
+
+    fcosKola(cosaDir: env.WORKSPACE, parallel: 5,
+             extraArgs: "-p gce --build ${params.VERSION} --gce-project ${gcp_project} --gce-image project/${gcp_image_project}/global/images/${gcp_image} --gce-json-key=/run/kubernetes/secrets/gcp-kola-tests-config")
```
dustymabe, line 65: any way to break this up into multiple lines? tends to make it easier to read when we have the `--` options on multiple lines like they were before.

jobs/kola-aws.Jenkinsfile
```diff
@@ -0,0 +1,62 @@
...
+        if (meta.amis.size() > 0) {
+            ami = meta['amis'][0]['hvm']
+            ami_region = meta['amis'][0]['name']
+        } else {
+            throw new Exception("No AMI found in metadata for ${params.VERSION}")
+        }
+    }
+
+    fcosKola(cosaDir: env.WORKSPACE, parallel: 5,
+             extraArgs: "-p aws --build ${params.VERSION} --aws-ami ${ami} --aws-region ${ami_region}")
```
dustymabe, line 61: how are the credentials getting passed in here?

previously we had `export AWS_CONFIG_FILE=\${AWS_FCOS_KOLA_BOT_CONFIG}` right before we called `cosa kola run`.

For GCP we pass in `--gce-json-key=` so that's taken care of.

jlebon:
Updated this now, but I'd like to sanity-check those `kola run` invocations locally.

jobs/kola-aws.Jenkinsfile
jlebon: Yup, good catch!

jobs/kola-aws.Jenkinsfile
```diff
@@ -0,0 +1,67 @@
...
+def s3_stream_dir = params.S3_STREAM_DIR
+if (s3_stream_dir == "") {
+    s3_stream_dir = "fcos-builds/prod/streams/${params.STREAM}"
+}
+
+cosaPod(image: params.COREOS_ASSEMBLER_IMAGE, memory: "256Mi",
+        secrets: ["aws-fcos-builds-bot-config", "aws-fcos-kola-bot-config"]) {
+
+    def AWS_FCOS_BUILDS_BOT_CONFIG = "/run/kubernetes/secrets/aws-fcos-builds-bot-config/config"
+    def AWS_FCOS_KOLA_BOT_CONFIG = "/run/kubernetes/secrets/aws-fcos-kola-bot-config/config"
```
dustymabe, line 42: hmm. That's not exactly what I had in mind. I think I'm bumping into a limitation in my knowledge on how things work.

I was thinking we'd be somehow re-using 

https://github.com/coreos/fedora-coreos-pipeline/blob/35f1f18598cbe39601d71f5a181248288e6357bc/manifests/pod.yaml#L23-L24

rather than having to define it here again.

jlebon: We no longer use that pod definition in those jobs. Instead, we're telling `cosaPod` that we need access to the specific secrets (see https://github.com/coreos/coreos-ci-lib/pull/37). That way, pods only get the secrets they need mounted in.

I guess an improvement we could make here is to expose the path to the secret in environment variables as part of that wrapper.

jlebon: E.g. in this file, notice:

```groovy
cosaPod(image: params.COREOS_ASSEMBLER_IMAGE, memory: "256Mi",
        secrets: ["aws-fcos-builds-bot-config", "aws-fcos-kola-bot-config"]) {
```

dustymabe: OK, sorry for the noise on this. I probably wouldn't have made the [other comment](https://github.com/coreos/fedora-coreos-pipeline/pull/270#discussion_r494999308) if I had realized how it worked.

dustymabe (approved):
LGTM

jobs/kola-aws.Jenkinsfile
jlebon: No worries, I think a good idea came out of it regardless: https://github.com/coreos/coreos-ci-lib/pull/39.

cgwalters (approved):
I am not an expert in this pipeline, but the high level direction looks really good to me, and superficially code looks good too.

darkmuggle:
LGTM....any thing hold up the merge?

dustymabe:
Maybe let's hold just a day or two til we get the `next` release out the door and then merge away!

jlebon:
OK, manually tested this locally now and fixed a bunch of issues. Also realized there's a gap in `fcosKola` we need to address first: https://github.com/coreos/coreos-ci-lib/pull/41. Once that PR is merged, I think we can merge this.

dustymabe (approved):
LGTM

PR #272, opened 2020-08-27
bump-lockfile: explicitly mark build as SUCCESS

jlebon:
In a good build, in the `finally` block the current build's status isn't
*yet* `SUCCESS`... because the build is still running. And so the
conditional we have there fails. Fix this by explicitly setting the
status to `SUCCESS` at the end of the job.


miabbott:
LGTM

PR #273, opened 2020-08-30
README.md: Update links/info for new CentOS CI 4.4 OCP cluster

cgwalters:
I actually can't log into the new Jenkins (or see the `fedora-coreos`
project) but I'm pretty sure this is the right one.  Add a link
to the "new" OCP4 docs I found.


jlebon (approved):
Thanks!

PR #276, opened 2020-09-14
pipeline: switch testing builds back to legacy PXE artifacts

bgilbert:
We need to do an out-of-cycle testing release this week.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -340,7 +340,7 @@ lock(resource: "build-${params.STREAM}") {
             stage('Build Live') {
                 utils.shwrap("""
                 case "${params.STREAM}" in
-                stable)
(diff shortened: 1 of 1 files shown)


sohankunkerkar (commented):
Is this change related to https://github.com/coreos/fedora-coreos-streams/issues/186#issuecomment-691587765?

bgilbert:
#275 is for https://github.com/coreos/fedora-coreos-streams/issues/186#issuecomment-691587765.  #276 is a partial reversion because we're discussing doing an out-of-cycle release to fix https://bugzilla.redhat.com/show_bug.cgi?id=1875699.

sohankunkerkar (approved):

PR #278, opened 2020-09-16
manifests/jenkins: use emptyDir for now

jlebon:
We're seeing intermittent issues where the NFS PVC mount seems too slow
for Jenkins to function properly:

https://pagure.io/centos-infra/issue/26#comment-685779

Switch over to an emptyDir for now. (This means that on restarts Jenkins
will lose all job history.)


dustymabe (approved):
LGTM

PR #283, opened 2020-10-05
utils: add umask workaround

jlebon:
This is affecting the pipeline too now:
https://pagure.io/centos-infra/issue/79

Add the hack here too until we fully drop utils.groovy in favour of
coreos-ci-lib (where this hack already exists).


lucab (approved):

dustymabe (approved):
LGTM

PR #286, opened 2020-10-23
jobs/kola-{aws,gcp}: use `--arg=val` format for platform args

jlebon:
Otherwise, it confuses the `cosa kola` wrapper which tries to process
the arguments.

We should fold the denylist handling and basic scenario testing into
kola proper so we can drop this wrapper.


dustymabe (approved):
LGTM

PR #287, opened 2020-10-23
jobs/kola-gcp: use `jq` instead of `readJSON`

jlebon:
We can't use `readJSON` because IIRC that runs on the master but the
config file is mounted in the pod. Just use `jq`. And that way we get to
use the automatically generated env var too.


dustymabe (approved):
LGTM

PR #288, opened 2020-10-30
Add new fedmsgs for builds, releases, and metadata

jlebon:
This patch adds four new fedmsgs:
- `build.state.change`: emitted at the start and end of a build
- `stream.release`: emitted when a build is released
- `stream.metadata.update`: emitted for stream/graph metadata updates

This will allow other services to hook more closely into our process.
Some use cases right now are AWS Marketplace listing updates, mirrors
triggering a sync, and OpenQA testing the latest artifacts.

I started with a helper for sending the messages in `utils`, but I'd
like to get rid of `utils` down the line in favour of moving all that
functionality into the shared library.

Closes: coreos/fedora-coreos-tracker#225

### DIFF
--- a/jobs/sync-stream-metadata.Jenkinsfile
+++ b/jobs/sync-stream-metadata.Jenkinsfile
@@ -8,9 +8,15 @@ properties([
     ])
 ])
 
-cosaPod {
+cosaPod(configMaps: ["fedora-messaging-cfg"], secrets: ["fedora-messaging-coreos-key"]) {
     git(url: 'https://github.com/coreos/fedora-coreos-streams', credentialsId: 'github-coreosbot-token')
     withCredentials([[$class: 'AmazonWebServicesCredentialsBinding', credentialsId: 'aws-fcos-builds-bot']]) {
+        // XXX: update configMap to use the /run/secrets value
+        shwrap("ln -sf \${FEDORA_MESSAGING_COREOS_KEY} /run/fedora-messaging-coreos-key")
+
+        // XXX: eventually we want this as part of the pod or built into the image we use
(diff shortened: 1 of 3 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-ci-lib/pull/47
Requires: https://github.com/coreos/fedora-coreos-releng-automation/pull/125

AdamWill:
awesome! thanks for this. [openQA scheduling ticket](https://pagure.io/fedora-qa/fedora_openqa/issue/85).

jobs/sync-stream-metadata.Jenkinsfile
```diff
@@ -8,13 +8,28 @@ properties([
...
+
+        cmd = """
             aws s3 sync --acl public-read --cache-control 'max-age=60' \
                 --exclude '*' --include 'streams/*' --include 'updates/*' \
                     ./ s3://fcos-builds
-        """)
+        """
+
+        def dry_run = shwrapCapture("${cmd} --dryrun").trim()
+        if (dry_run != "") {
```
dustymabe, line 27: so if you run it with `--dry-run` and there's output then we need to send a message? Could we just capture the output of the command without dry-run and have it work? either way might be worth a comment.

dustymabe (approved):
LGTM. One comment

jobs/sync-stream-metadata.Jenkinsfile
jlebon: I think that would work, but I don't like the idea of capturing output of the real run (though I guess we could print it out, but I find this cleaner). Added a comment!

dustymabe: 👍

jlebon:
Alrighty, with releases out, let's do this!

PR #289, opened 2020-11-09
Point fedmsg config to new k8s secret path

jlebon:
This is the path under which `coreos-ci-lib` puts the secret. We want
the config to work in both the "old world" (where it uses `pod.yaml`)
and the "new world" (where it uses `pod(secret: ...)` from
`coreos-ci-lib`), so let's update the path in the config and `pod.yaml`
to match `coreos-ci-lib`.

This fixes one `XXX` in the `sync-stream-metadata` job which turns out
to be a hard blocker because the container runs as non-root and so
doesn't have the privs to link at `/run/fedora-messaging-coreos-key`.


jlebon:
That *should* fix the `sync-stream-metadata` job failure.

dustymabe (approved):
LGTM

PR #290, opened 2020-11-09
jobs/sync-stream-metadata: fix --dryrun arg passing

jlebon:
We need the appended `--dryrun` to be on the same line as `aws s3 sync`.

### DIFF
--- a/jobs/sync-stream-metadata.Jenkinsfile
+++ b/jobs/sync-stream-metadata.Jenkinsfile
@@ -14,11 +14,14 @@ cosaPod(configMaps: ["fedora-messaging-cfg"], secrets: ["fedora-messaging-coreos
         // XXX: eventually we want this as part of the pod or built into the image we use
         shwrap("git clone --depth=1 https://github.com/coreos/fedora-coreos-releng-automation /var/tmp/fcos-releng")
 
-        cmd = """
+        def cmd = """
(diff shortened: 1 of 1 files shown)


jobs/sync-stream-metadata.Jenkinsfile
```diff
@@ -17,8 +17,7 @@ cosaPod(configMaps: ["fedora-messaging-cfg"], secrets: ["fedora-messaging-coreos
         cmd = """
             aws s3 sync --acl public-read --cache-control 'max-age=60' \
                 --exclude '*' --include 'streams/*' --include 'updates/*' \
-                    ./ s3://fcos-builds
-        """
+                    ./ s3://fcos-builds"""
```
dustymabe, line 20: I could see someone mistakenly putting back the `"""` on the next line. Any way we could trim the newline in the `def dry_run = shwrapCapture("${cmd} --dryrun").trim()` instead?

jlebon: I changed this to just explicitly `trim()` the `cmd` after defining it.

dustymabe (approved):
LGTM. I agree it seems silly to `.trim()` right after you define the variable. A comment would help explain to people why we did that (optional).

jlebon:
Added comment!

PR #291, opened 2020-11-09
jobs/sync-stream-metadata: sync one stream at a time

jlebon:
One obvious thing that I missed while working on #288 was that this job
doesn't actually have a `STREAM` parameter since it syncs all the
streams in one shot. So the fedmsg it emits will have `null` as the
stream name.

To make this stream-sensitive, let's change to doing `aws s3 sync`
separately for each stream with change detection for each. That way we
only emit fedmsgs for streams which actually had their metadata changed.

### DIFF
--- a/jobs/sync-stream-metadata.Jenkinsfile
+++ b/jobs/sync-stream-metadata.Jenkinsfile
@@ -1,5 +1,11 @@
 @Library('github.com/coreos/coreos-ci-lib') _
 
+def streams
+node {
+    checkout scm
+    streams = load("streams.groovy")
+}
+
 properties([
     pipelineTriggers([
         githubPush(),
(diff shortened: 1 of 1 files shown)


jlebon:
Diff with whitespace ignored: https://github.com/coreos/fedora-coreos-pipeline/pull/291/files?w=1

jobs/sync-stream-metadata.Jenkinsfile
```diff
@@ -14,25 +20,27 @@ cosaPod(configMaps: ["fedora-messaging-cfg"], secrets: ["fedora-messaging-coreos
...
-            """)
+            // Do a dry run first to see if there's any work that actually needs to
+            // be done. `aws s3 sync` doesn't print anything if everything is
+            // synced up.
+            def dry_run = shwrapCapture("${cmd} --dryrun").trim()
+            if (dry_run != "") {
+                shwrap("${cmd}")
+                utils.shwrap("""
+                /var/tmp/fcos-releng/scripts/broadcast-fedmsg.py --fedmsg-conf=\${FEDORA_MESSAGING_CFG}/fedmsg.toml \
+                    stream.metadata.update --stream ${it}
```
dustymabe, line 41: dumb question.. Where is `${it}` defined?

dustymabe: Is it an automatic variable that gets created from `streams.production.each{}` ?

jlebon: > Is it an automatic variable that gets created from `streams.production.each{}` ?

Yup, exactly! See https://riptutorial.com/groovy/example/32092/each-and-eachwithindex#example. I'll name it `stream` to be more explicit.

dustymabe (approved):
LGTM

dustymabe:
👍

PR #292, opened 2020-11-09
pipeline: declare basearch earlier

jlebon:
Otherwise, the `finally` block where we emit a message won't be able to
access it.


dustymabe (approved):
LGTM

PR #293, opened 2020-11-09
jobs/sync-stream-metadata: fix no-change detection

jlebon:
Switching to syncing per-stream in #291 worked, but it broke change
detection using `--dryrun` such that now `aws s3 sync` always uploads
files even if nothing changed. Some bug in using
`--exclude '*' --include a/specific/file` maybe? I also noticed that the
invocation takes much longer, I think because we're syncing at the root
of the bucket and it needs to get a full listing of the files in there
or something?

Anyway, switching to syncing the `streams` and `updates` directories
specifically is much faster and restores the no-change detection, so
let's do that.

I went with Dusty's suggestion in #288 to just do the operation without
`--dryrun` and determine based on that whether to emit a fedmsg.


dustymabe (approved):
LGTM

PR #294, opened 2020-11-09
jobs/kola-openstack: add job for testing OpenStack on VexxHost

dustymabe:


dustymabe:
Needs to be tested properly before merging.

HACKING.md
```diff
@@ -131,6 +131,19 @@ assuming they have enough permissions.
 oc create secret generic gcp-kola-tests-config --from-file=config=/path/to/kola-secret
 ```
 
+### [OPTIONAL] Creating OpenStack credentials configs
+
+If you want to run kola tests against OpenStack images you need to
+create a secret (`clouds.yaml` format) for talking with the openstack
```
jlebon, line 137: Minor/optional: is there a page we could link to which describes `clouds.yaml`?

Also

```suggestion
create a secret (`clouds.yaml` format) for talking with the OpenStack
```

:)

manifests/pod.yaml
```diff
@@ -28,6 +28,8 @@ spec:
          value: /.gcp-image-upload-config/config
        - name: GCP_KOLA_TESTS_CONFIG
          value: /.gcp-kola-tests-config/config
+       - name: OPENSTACK_KOLA_TESTS_CONFIG
+         value: /.openstack-kola-tests-config/config
```
jlebon, line 32: Hmm, we shouldn't need to touch `pod.yaml` if it's just the `kola-openstack` job that uses it (and actually, we should remove the GCP and AWS test secrets as well from there).

jobs/kola-openstack.Jenkinsfile
```diff
@@ -0,0 +1,104 @@
...
+        def meta = readJSON file: "builds/${params.VERSION}/${basearch}/meta.json"
+        if (meta.images.openstack) {
+            openstack_image_filename = meta.images.openstack.path
+            openstack_image_sha256 = meta.images.openstack.sha256
+            openstack_image_filepath = builds/${params.VERSION}/${basearch}/${openstack_image_filename}
+        } else {
+            throw new Exception("No OpenStack artifacts found in metadata for ${params.VERSION}")
+        }
+
+        # Copy down the openstack image from S3, verify, uncompress
```
jlebon, line 59: Minor/optional: maybe an `XXX` here for folding this into `cosa` (i.e. linking to https://github.com/coreos/coreos-assembler/issues/1508)?

jlebon (commented):
LGTM overall from a quick review. Mostly nits!

HACKING.md
dustymabe: Yep will link.

jlebon:
Should this patch series also be modifying the main pipeline to schedule an OpenStack run?

dustymabe:
> Should this patch series also be modifying the main pipeline to schedule an OpenStack run?

yep. Check out the latest upload.

jobs/kola-openstack.Jenkinsfile
```diff
@@ -0,0 +1,111 @@
...
+                    --openstack-floating-ip-network=public                   \
+                    --openstack-image=${openstack_image_name}""")
+    } finally {
+        stage('Delete Image') {
+            // Delete the image in OpenStack
+            shwrap("""
+            ore openstack --config-file=\${OPENSTACK_KOLA_TESTS_CONFIG}/config \
+                 delete-image --id=${openstack_image_name}
+            """)
+        }
```
jlebon, line 109: Would be nice if there were a more foolproof way to do this so that we're sure we'll never leak images. E.g. right now if this job gets interrupted before we delete the image, it'll never get cleaned up.

Hmm, how about this: name the image `${params.STREAM}` and delete any image of that name before creating it (using `ore openstack delete-image`). That way, we're sure that any leaks eventually gets cleaned up. The downside is that we wouldn't be able to run multiple jobs testing the same stream at different versions, though not sure how likely that is.

jobs/kola-openstack.Jenkinsfile
```diff
@@ -0,0 +1,111 @@
...
+
+currentBuild.description = "[${params.STREAM}] - ${params.VERSION}"
+
+def s3_stream_dir = params.S3_STREAM_DIR
+if (s3_stream_dir == "") {
+    s3_stream_dir = "fcos-builds/prod/streams/${params.STREAM}"
+}
+
+cosaPod(image: params.COREOS_ASSEMBLER_IMAGE,
+        memory: "256Mi", kvm: false,
```
jlebon, line 39: Doesn't have to be in this PR, though we need to remember to use `kvm: false` in other cloud kola jobs too.

jlebon (approved):
Nice work! Minor comments, but OK to merge as is too.

jobs/kola-openstack.Jenkinsfile
dustymabe: added a new commit to do this.

jobs/kola-openstack.Jenkinsfile
dustymabe: added in a new commit

dustymabe:
Will do one more round of testing once https://github.com/coreos/coreos-assembler/pull/1853 lands and then merge.

jlebon (approved):

manifests/pod.yaml
dustymabe: we ended up needing to because the main Jenkinsfile cues running the tests off of them 

https://github.com/coreos/fedora-coreos-pipeline/blob/390a7f45d3cec9dc70f8911e13424adbbbdfe9e6/Jenkinsfile#L499

PR #295, opened 2020-11-10
jobs/sync-stream-metadata: use `aws s3 cp` instead of `aws s3 sync`

jlebon:
`aws s3 sync` doesn't sync based on file contents, but rather just on
file size and timestamps. And because our git checkout will always be
fresh, it'll always seem newer than the S3 copy and thus we'll always
upload new copies.

There doesn't seem to be a `--checksum` flag like `rsync` supports.

Instead, let's just do this a really basic way:
1. download the S3 version
2. check if there's a diff
3. if so, upload the new version and emit a fedmsg


dustymabe (approved):
LGTM

PR #296, opened 2020-11-10
pipeline: pass URL to build dir in fedmsgs

jlebon:
Otherwise it will require services which want to make use of these
fedmsgs to hardcode knowledge of our bucket layout.


jlebon:
Requires: https://github.com/coreos/fedora-coreos-releng-automation/pull/126

jlebon:
Worth nothing that this new fedmsg is exposing the cosa build dir for the first time to be consumed by code outside our direct control. (Though note RHCOS made that jump a long time ago.)

dustymabe (approved):
LGTM

PR #298, opened 2020-11-17
manifests/pod: use jnlp image from quay.io

jlebon:
We're hitting pull limit errors on Docker Hub, and anyway that one is
deprecated in favour of the one on Quay.io so let's switch to that.

### DIFF
--- a/manifests/pod.yaml
+++ b/manifests/pod.yaml
@@ -12,7 +12,7 @@ spec:
   serviceAccountName: jenkins
   containers:
    - name: jnlp
-     image: jenkins-slave-base-centos7:latest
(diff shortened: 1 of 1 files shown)


manifests/pod.yaml
```diff
@@ -12,7 +12,7 @@ spec:
   serviceAccountName: jenkins
   containers:
    - name: jnlp
-     image: jenkins-slave-base-centos7:latest
+     image: quay.io/openshift/origin-jenkins-agent-base
```
dustymabe, line 15: should we include the tag, like we used to (just to be more explicit?)

dustymabe (approved):
trivial LGTM. I trust @Jlebon is pointing us to the right image :)

PR #299, opened 2020-11-17
jobs/sync-stream-metadata: use --acl and --cache-control args

jlebon:
These were accidentally dropped in #295.

### DIFF
--- a/jobs/sync-stream-metadata.Jenkinsfile
+++ b/jobs/sync-stream-metadata.Jenkinsfile
@@ -32,9 +32,12 @@ cosaPod(configMaps: ["fedora-messaging-cfg"], secrets: ["fedora-messaging-coreos
             if (shwrapRc("git diff --exit-code") != 0) {
                 shwrap("git reset --hard HEAD")
                 for (subdir in ["streams", "updates"]) {
-                    shwrap("aws s3 cp ${subdir}/${stream}.json s3://fcos-builds/${subdir}/${stream}.json")
+                    shwrap("""
+                        aws s3 cp --acl public-read --cache-control 'max-age=60' \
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM.

jobs/sync-stream-metadata.Jenkinsfile
```diff
@@ -32,9 +32,12 @@ cosaPod(configMaps: ["fedora-messaging-cfg"], secrets: ["fedora-messaging-coreos
...
                 shwrap("git reset --hard HEAD")
                 for (subdir in ["streams", "updates"]) {
-                    shwrap("aws s3 cp ${subdir}/${stream}.json s3://fcos-builds/${subdir}/${stream}.json")
+                    shwrap("""
+                        aws s3 cp --acl public-read --cache-control 'max-age=60' \
+                            ${subdir}/${stream}.json s3://fcos-builds/${subdir}/${stream}.json
+                    """)
                 }
-                utils.shwrap("""
+                shwrap("""
```
dustymabe, line 40: the move from `utils.shwrap` to just `shwrap` significant?

jlebon: TBH, I'm actually not sure how it works currently with `utils.shwrap`. `utils` here is mapping to https://github.com/coreos/coreos-ci-lib/blob/master/vars/utils.groovy, but `shwrap` is not defined in that file -- I'm guessing Groovy somehow still figures out we mean https://github.com/coreos/coreos-ci-lib/blob/master/vars/shwrap.groovy?

Anyway, it should've been `shwrap()` from the start like the other calls in this file.

PR #300, opened 2020-12-11
Jenkinsfile: Create and upload keylime hashlist

ashcrow:
When a builds occur we now generate a keylime hashlist and upload it to an expected location. We also upload a sha256sum.

See:
- https://github.com/keylime/enhancements/blob/master/16_remote_allowlist_retrieval.md
- https://github.com/coreos/coreos-assembler/blob/master/src/cmd-generate-hashlist
- ~https://github.com/coreos/coreos-assembler/pull/1970~

When closing [the upload PR](https://github.com/coreos/coreos-assembler/pull/1970) we decided to modify the pattern for POC as such:
- Output the hashlist as `exp-hash.json`
- Ensure the `exp-hash.json` file is in the builds/ directory so it is uploaded as a sibling of `{meta|commitmeta}.json`
- sha256sum the `exp-hash.json` file as `exp-hash.json-CHECKSUM`
- Once this work moves beyond POC we will update the output file to an expected location + sign the hashlist

cc @mpeters

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -444,6 +444,16 @@ lock(resource: "build-${params.STREAM}") {
             }
         }
 
+        // Generate KeyLime hashes for attestation on official builds
+        // This is a POC setup and will be modified over time
+        // See: https://github.com/keylime/enhancements/blob/master/16_remote_allowlist_retrieval.md
+        stage('KeyLime Hash Generation') {
+            utils.shwrap("""
+            cosa generate-hashlist --release=${newBuildID} --output=builds/${newBuildID}/${basearch}/exp-hash.json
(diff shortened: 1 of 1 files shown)


ashcrow:
Looking for feedback before lifting WIP.

Jenkinsfile
```diff
@@ -477,6 +477,17 @@ lock(resource: "build-${params.STREAM}") {
             }
         }
 
+        if (official) {
```
jlebon, line 480: Meh, let's just do it for all builds. That way if it breaks, we don't find out when we're trying to release.

jlebon (approved):
Sorry, totally missed that this was updated already! (I think when it's in draft, notifications don't go out).

One comment, LGTM otherwise!

ashcrow:
@jlebon updated

PR #302, opened 2020-12-18
streams: add rawhide mechanical stream

jlebon:
We need to start tracking changes that happen in rawhide earlier so that
we can better participate in the branching and stabilization process of
Fedora releases.

Let's create a mechanical rawhide stream.


dustymabe (approved):
LGTM

dustymabe:
👍 for every 24H

PR #304, opened 2020-12-21
pipeline: don't upload to clouds for rawhide for now

jlebon:
It doesn't quite work yet because e.g. in GCP we need to create a
license. But at least, let's push out builds which passed local QEMU
tests.

(Related to this, adding another stream like this reminds we really need
to tackle GC at some point. E.g. for rawhide, we should have an
aggressive policy.)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -83,6 +83,8 @@ properties([
     durabilityHint('PERFORMANCE_OPTIMIZED')
 ])
 
+def is_mechanical = (params.STREAM in streams.mechanical)
+
 // Note the supermin VM just uses 2G. The really hungry part is xz, which
 // without lots of memory takes lots of time. For now we just hardcode these
 // here; we can look into making them configurable through the template if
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -395,7 +395,7 @@ lock(resource: "build-${params.STREAM}") {
             // Key off of s3_stream_dir: i.e. if we're configured to upload artifacts
             // to S3, we also take that to mean we should upload an AMI. We could
             // split this into two separate developer knobs in the future.
-            if (s3_stream_dir) {
+            if (s3_stream_dir && !(params.STREAM in streams.mechanical)) {
```
cgwalters, line 398: Maybe hoist out a
`Boolean not_mechanical = !(params.STREAM in streams.mechanical);` and use it elsewhere?

cgwalters (approved):
This says "for rawhide" but it's really all mechanical streams right?

Jenkinsfile
jlebon: Thanks, did that (though went with `is_mechanical`).

jlebon:
> This says "for rawhide" but it's really all mechanical streams right?

Yeah, though right now the only mechanical stream is rawhide. In a "jobspec" world, I would've done it for rawhide only, though I'm just taking a shortcut right now since ideally we revert this soon-ish once we fix the GCP thing.

Thanks for the review!

cgwalters:
> Yeah, though right now the only mechanical stream is rawhide.

Ahh, I missed the `/* */` originally when looking at
`mechanical = ['rawhide' /*'bodhi-updates', 'bodhi-updates-testing', 'branched' */]`

PR #306, opened 2021-01-13
pipeline: stop triggering OpenStack tests

jlebon:
They're currently failing because of
https://github.com/coreos/coreos-assembler/issues/1977.


jlebon:
(I already disabled the job itself, but this is more explicit and in git.)

lucab (approved):

PR #307, opened 2021-01-14
pipeline: move from `trans.py` to `cosa generate-release-meta`

jlebon:
This was moved to cosa in
https://github.com/coreos/coreos-assembler/pull/2006.


jlebon:
Hold until that PR merges.

cgwalters (approved):

PR #308, opened 2021-01-18
jobs/kola: add KOLA_TESTS param

jlebon:
Sometimes we just want to rerun some tests only. Add a parameter to make
that easy.


sinnykumari (approved):
lgtm

PR #309, opened 2021-02-16
jenkins: add timestamper plugin

jlebon:
This will allow us to add timestamps to all build logs.

Prompted by: https://github.com/coreos/rpm-ostree/pull/2583


jlebon:
Follow-ups to this in https://github.com/coreos/coreos-ci/pull/24 and https://github.com/coreos/fedora-coreos-pipeline/pull/314.

PR #310, opened 2021-02-18
streams: add branched mechanical stream

jlebon:
We're adding a `branched` stream:
https://github.com/coreos/fedora-coreos-tracker/issues/743


cgwalters (approved):
Looks sane

PR #313, opened 2021-02-22
plugins: add build-token-root

jlebon:
This plugin allows scripts to trigger jobs using a secret token without
having to set up authentication. I'd like to use this with GitHub
Actions in the future.


miabbott (approved):
LGTM

PR #316, opened 2021-03-02
Temporarily use --no-default-checks

jlebon:
We're hitting SELinux flakes in the releases:
https://github.com/coreos/fedora-coreos-tracker/issues/751

Working on a better way to allow known SELinux failures, but for now
let's just use the new `--no-default-checks` to unblock releases.


cverna (approved):

cverna:
LGTM

PR #317, opened 2021-03-02
Revert "Temporarily use --no-default-checks"

jlebon:
This reverts commit 74a78a3858038cdf738b3e4952e8c94e8ebd50d5.

We're dropping SELinux checking and support for that switch from kola,
so stop using it:

https://github.com/coreos/coreos-assembler/pull/2066


jlebon:
Quay.io rebuilt with that cosa revert PR, so we can get this in now!

miabbott:
LGTM

PR #318, opened 2021-03-12
Jenkinsfile.production: trigger on GitHub pushes

jlebon:
So far, we've been triggering new production builds manually during the
release process.

In an effort to streamline that process, let's now switch to
automatically triggering a new Jenkins build whenever one of the
production branches is pushed to.

This does mean that we'll be triggering builds even when making changes
on the branch between releases (e.g. tweaking `manifest.yaml`). However,
this happens very rarely, and even then, it's totally fine if it
triggers a build. The pipeline is designed to allow multiple release
candidate builds.

### DIFF
--- a/Jenkinsfile.mechanical
+++ b/Jenkinsfile.mechanical
@@ -13,22 +13,6 @@ properties([
 ])
 
 node {
-    change = checkout(
-        [$class: 'GitSCM',
-         userRemoteConfigs: [
-            [url: 'https://github.com/coreos/fedora-coreos-config']
-         ],
-         branches: streams.as_branches(streams.mechanical)
-        ]
-    )
-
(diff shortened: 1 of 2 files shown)


jlebon:
I was initially working on a fancier solution with a GitHub Action which triggers builds in Jenkins remotely based on a `trigger-build`-type label on the promotion PR. But there's no straight-forward way to set up an auth token for the pipeline job via Jenkinsfile's `pipelineTriggers`. And anyway, I think it's over-complicating it since (as mentioned in the commit message) in the majority of cases a push means a promotion/new release, and it's safe to generalize. (It also avoids us having to maintain *yet another* secret, which is nice).

jlebon:
I think this needs to be part of a larger discussion about release process streamlining.

PR #319, opened 2021-03-15
jobs/bump-lockfile: fix artifact archiving

jlebon:
We weren't correctly archiving artifacts because we run inside a
`dir(branch) { ... }` block, which causes `archiveArtifacts` to
interpret paths relative to that dir, but `fcosKola()` outputs the
artifact at the top-level workspace directory.

Work around this by just calling `fcosKola` from the workspace.


miabbott (approved):
LGTM

Issue #320, opened 2021-03-19
Prevent reuse of version numbers from failed builds


bgilbert:
A build can fail before uploading to the S3 bucket but after uploading to AWS and GCP.  In that case, rerunning the build will reuse the same version number (unless overridden), but ore upload will skip the AWS/GCP reupload because of its idempotency rules.  This could potentially cause divergent build artifacts to be shipped on those platforms.

Add an interlock which prevents version numbers from being reused across builds.

jlebon:
Related: https://github.com/coreos/coreos-assembler/issues/668

What would be pretty easy to do is a first early S3 upload of the artifacts *before* we do any cloud uploads (but importantly, after it passed all the QEMU tests). This adds the build ID to the canonical `builds.json` so that any failure after that cannot result in an ID reuse.

jlebon:
Draft in https://github.com/coreos/fedora-coreos-pipeline/pull/321.

dustymabe:
At least for GCP we run with `--force`. We could easily start defaulting to that for AWS too to mitigate this potential problem (I think).

https://github.com/coreos/coreos-assembler/blob/67e71ce5376b3ca03077f8897989f8da6f29e8e2/src/cosalib/gcp.py#L64

dustymabe:
I think this is addressed in https://github.com/coreos/fedora-coreos-pipeline/commit/08b3ebcdcbef1b149d51df090705778565a9c36c - We now "reserve" our build ID by doing an early upload of the ostree and also updating builds.json. This happens before we upload to any cloud providers.

PR #321, opened 2021-03-22
pipeline: archive to S3 before uploading to clouds

jlebon:
That way if the cloud uploads fail (or any step after that), we still
have the build ID reserved in the `builds.json` and don't risk re-using
it.

Closes: https://github.com/coreos/fedora-coreos-pipeline/issues/320

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -394,6 +394,17 @@ lock(resource: "build-${params.STREAM}") {
             }
             parallel pbuilds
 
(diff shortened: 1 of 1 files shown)


dustymabe:
This early archive was achieved in https://github.com/coreos/fedora-coreos-pipeline/commit/08b3ebcdcbef1b149d51df090705778565a9c36c so we don't need this anymore.

Issue #232, opened 2020-05-08
Job bump-lockfile shouldn't push changes that are just metadata date updates


jlebon:
See e.g. https://github.com/coreos/fedora-coreos-config/commit/e34014c4812acb854a0acbd5e337096eddf51b6d:

```diff
diff --git a/manifest-lock.x86_64.json b/manifest-lock.x86_64.json                                                                                                                                                                           
index e1d65d0..91e58c5 100644                                                                                                                                                                                                               
--- a/manifest-lock.x86_64.json                                                                                                                                                                                                             
+++ b/manifest-lock.x86_64.json                                                                                                                                                                                                             
@@ -1088,7 +1088,7 @@                                                                                                                                                                                                                       
     }
   },
   "metadata": {
-    "generated": "2020-05-05T19:53:47Z",                                                                                                                                                                                                   
+    "generated": "2020-05-05T21:07:22Z",                                                                                                                                                                                                   
     "rpmmd_repos": {
       "fedora": {
         "generated": "2019-10-23T22:52:47Z"
```

We should just `cosa buildprep` from the latest build and use change detection the way the main pipeline does it.

dustymabe:
@sohankunkerkar - would you want to work on this when you get free?

dustymabe:
Is this fixed by #322. This "just metadata timestamp bump" commit just landed in `next-devel`: https://github.com/coreos/fedora-coreos-config/commit/c374504ad3fdc3d868dbeb273e05f43cb5c2afb0

jlebon:
Yup, you're right. There's another subtle issue here. Will fix.

PR #322, opened 2021-04-06
bump-lockfile: strengthen no-op detection

jlebon:
Right now, we try to no-op early if we detect that the base lockfiles
didn't change. However, `--update-lockfile` will still touch the
lockfiles if just the repo timestamps changed (#232). This then causes
us to run `cosa build`, which no-ops, and then subsequent tests fail
because there is no e.g. qcow2 or ostree artifact that was built.

Fix this by detecting the no-op case just like the main pipeline does.

Closes: #232


dustymabe (commented):
LGTM

PR #325, opened 2021-04-12
jobs/bump-lockfile: set up git user upfront

jlebon:
These are global settings, so they just need to be done once, not
per-branch.

This avoids git hitting locking issues by racing with itself across
branches.


dustymabe (approved):
LGTM

dustymabe:
still LGTM

PR #330, opened 2021-04-13
manifest/jenkins: bump readiness timeout to 20 mins

jlebon:
The default is 10 mins, which is too short for Jenkins to come up
(likely because of slow NFS). So then OCP considers the rollout failed
and reverts to the previous version.


jlebon:
(Had to do this on the live dc object yesterday to get Jenkins back up.)

dustymabe (approved):
LGTM

Issue #331, opened 2021-04-16
break bump-lockfile into multiple jobs


dustymabe:
It currently updates both testing-devel and next-devel in a single job. If one of these streams fails for any reason then bump-lockfile looks like the entire job fails, even though half of it succeeded. Having them co-mingled means that it's also hard to tell which of the two failed (or maybe both failed?) and in what way. Let's break them into two separate jobs to ease up on the confusion.

jlebon:
Yeah, and it doesn't help that the BlueOcean view gets super confused by the parallel + nested stages we got going on in that job.

Easiest way to do this is: "un-parallelize" bump-lockfile, drop the cron trigger, and make it take the target stream as a parameter. Then add another job with a cron trigger which just triggers bump-lockfile jobs for each stream.

PR #333, opened 2021-04-26
bump-lockfile: de-parallelize

jlebon:
Right now, the job tries to update both `testing-devel` and `next-devel`
in parallel, but this makes the output messy and hard to follow. It also
makes it harder to tell from the job status page which of the two
branches failed (and the BlueOcean UI isn't much help either).

Instead, let's make `bump-lockfile` focus on a single stream, and have a
driver job which simply triggers `bump-lockfile` once per stream so that
they're separate jobs. This mirrors how the main pipeline jobs
themselves are set up.

Closes: #331

### DIFF
--- a/jobs/bump-lockfile.Jenkinsfile
+++ b/jobs/bump-lockfile.Jenkinsfile
@@ -1,23 +1,32 @@
 @Library('github.com/coreos/coreos-ci-lib') _
 
+def streams
+node {
+    checkout scm
+    streams = load("streams.groovy")
+}
+
 repo = "coreos/fedora-coreos-config"
-branches = [
-    "testing-devel",
-    "next-devel"
-]
 botCreds = "github-coreosbot-token"
 
 // Base URL through which to download artifacts
 BUILDS_BASE_HTTP_URL = "https://builds.coreos.fedoraproject.org/prod/streams"
 
 properties([
-    pipelineTriggers([
-        // we don't need to bump lockfiles any more often than daily
-        cron("H H * * *")
+    // we're only triggered by bump-lockfiles
+    pipelineTriggers([]),
+    parameters([
+        choice(name: 'STREAM',
+               choices: streams.development,
+               description: 'Fedora CoreOS development stream to bump'),
     ])
 ])
 
-try { timeout(time: 120, unit: 'MINUTES') { cosaPod {
+echo "Waiting for bump-${params.STREAM} lock"
+currentBuild.description = "[${params.STREAM}] Waiting"
+
+try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTES') { cosaPod {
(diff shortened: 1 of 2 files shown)


jobs/bump-lockfile.Jenkinsfile
```diff
@@ -1,121 +1,124 @@
...
-                shwrap("cosa fetch --strict")
-                shwrap("cp src/config/fedora-coreos-pool.repo{.bak,}")
-            }
-
-            stage("Build") {
-                shwrap("cosa build --strict")
-            }
-
-            // need to run this in the workspace context because `fcosKola()`
-            // archives directly there but archiveArtifacts is dir-sensitive
```
dustymabe, line 74:
```
// need to run this in the workspace context because `fcosKola()`
// archives directly there but archiveArtifacts is dir-sensitive
```

No longer applicable?

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -1,121 +1,124 @@
...
+        shwrap("git -C src/config commit -am 'lockfiles: bump to latest' -m 'Job URL: ${env.BUILD_URL}' -m 'Job definition: https://github.com/coreos/fedora-coreos-pipeline/blob/master/jobs/bump-lockfile.Jenkinsfile'")
+        withCredentials([usernamePassword(credentialsId: botCreds,
+                                          usernameVariable: 'GHUSER',
+                                          passwordVariable: 'GHTOKEN')]) {
+          // should gracefully handle race conditions here
+          sh("git -C src/config push https://\${GHUSER}:\${GHTOKEN}@github.com/${repo} ${branch}")
         }
-    }] }
+    }
+    currentBuild.description = "[${params.STREAM}] ⚡ (pushed)"
```
dustymabe, line 119: ❤️

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -1,121 +1,124 @@
...
+    // do a first fetch where we only fetch metadata; no point in
+    // importing RPMs if nothing actually changed
+    stage("Fetch Metadata") {
+        shwrap("cosa fetch --update-lockfile --dry-run")
+    }
+
+    def newPkgChecksum = shwrapCapture("jq -c .packages src/config/manifest-lock.*.json | sha256sum")
+    if (newPkgChecksum == prevPkgChecksum) {
+        println("No changes")
+        currentBuild.description = "[${params.STREAM}] 💤 (no change)"
```
dustymabe, line 52: ❤️

dustymabe (approved):
LGTM - one question..

jobs/bump-lockfile.Jenkinsfile
jlebon: Yeah, because now the workdir is directly in the workspace instead of a subdir.

jlebon:
Let's try this out!

PR #334, opened 2021-04-26
bump-lockfile: add stream name to Slack message

jlebon:
Now that the job is stream-specific, we can be more specific in the
Slack message.


dustymabe (approved):
LGTM

PR #337, opened 2021-05-04
migration: branch renaming to `main`

dustymabe:


dustymabe:
prep work for renaming our branches to `main`. `hold` this until we're ready to mrege all relevant repos.

Jenkinsfile
```diff
@@ -41,9 +41,9 @@ BUILDS_BASE_HTTP_URL = "https://builds.coreos.fedoraproject.org/prod/streams"
 
 def coreos_assembler_image
 if (official) {
-    coreos_assembler_image = "coreos-assembler:master"
+    coreos_assembler_image = "coreos-assembler:main"
```
jlebon, line 44: Should we focus this PR on adjustments to this repo only and then have separate PRs for updating links and references to other repos? Are we planning to switch everything over at the same time? Might make for a bumpy migration. :)

jlebon:
There's also `jenkins/master/` which needs to be renamed (and the `contextDir` in the `jenkins-s2i.yaml` adjusted).

Jenkinsfile
dustymabe: I think I'd prefer to move everything over in one step in each repo. If we need to order some repos after the others that's fine. For this particular item we can use `coreos-assembler:latest` instead of `coreos-assembler:main` and that should alleviate some of our migration pain.

dustymabe: ok switched to `:latest` tag.

jlebon (approved):

PR #338, opened 2021-05-05
jobs/bump-lockfile: explicitly set SUCCESS status in no-op case

jlebon:
Otherwise, the Slack notifier code will think that the job failed and
emit a trashfire message.


dustymabe:
>  and emit a trashfire message.

🔥🗑️

dustymabe (approved):
LGTM

PR #339, opened 2021-05-06
jenkins.yaml: use `fsGroup: 0` for Jenkins

jlebon:
Jenkins already runs as gid 0, but we need to specify `fsGroup: 0` in
the pod definition too so that Kubernetes will make sure that the local
PVC mounted in will be chown'ed to the root gid so we can access it.


dustymabe (approved):
LGTM

PR #341, opened 2021-05-12
Revert "COSA image: use :latest tag to ease migration"

jlebon:
This reverts commit 0f8cf3a.

The `:latest` tag has fuzzy semantics. Let's be more explicit here and
use the actual branch-derived name.


dustymabe:
would another option be to revert 0f8cf3a ?

dustymabe (approved):
LGTM

PR #342, opened 2021-05-13
jobs/sync-stream-metadata: specify `main` branch when cloning fedora-coreos-streams

jlebon:
Otherwise, surprisingly the `git` step defaults to `master` instead of
whatever the default branch upstream is.


dustymabe (approved):
LGTM

PR #347, opened 2021-06-02
Drop Jenkinsfile.stream.metadata.generator and stop mounting GitHub token

jlebon:
The idea with this job was that we'd run it and it would create a
streams PR for us. In practice, we never really incorporated it into our
release workflow. We usually do triple releases in a single PR, and it
also doesn't handle the update metadata.

Down the line we do want to automate this stuff, but ideally make it
even more hands off so that e.g. Jenkins automatically opens an initial
PR after the build and the releaser interacts with the process through
the PR (i.e. doesn't have to go into Jenkins at all).

So parts of this code may come back, but for now let's remove it since
it's not used anyway. My immediate reason for doing this is that right
now, because of this job we're also mounting the GitHub token into the
pods we use to build and release FCOS even if they don't really need it.

We have Jenkins credentials now for these and that's what jobs should
use to access the GitHub token. That way it can be scoped to just the
part of the job which needs the token. (See e.g. the `bump-lockfile`
job.)


jlebon:
> My immediate reason for doing this is that right
now, because of this job we're also mounting the GitHub token into the
pods we use to build and release FCOS even if they don't really need it.

And my immediate reason for looking at that is that we're currently in the process of rotating the GitHub coreosbot token used by various bits in this repo and CoreOS CI.

dustymabe (approved):
LGTM

Thanks for the detailed reasoning.

PR #349, opened 2021-06-03
Use shwrap* helpers from coreos-ci-lib in prod pipelines

jlebon:
This introduces coreos-ci-lib in the production jobs and starts making
use of its shwrap helpers. This allows us to drain the duplicates we
have here in the "utils" module.

There's more stuff we can drain from "utils" into coreos-ci-lib, but
this at least paves the way for using more coreos-ci-lib helpers in
these jobs. (For example, we should switch over to `fcosBuild`,
`fcosKola`, etc... the way it's used in upstream CI). Notably, this will
also allow us to leveraging the gangplank work going on in
coreos-ci-lib.

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -1,52 +1,12 @@
-import org.yaml.snakeyaml.Yaml;
+@Library('github.com/coreos/coreos-ci-lib@main') _
 
-def shwrap(cmds) {
-    sh """
-        set -xeuo pipefail
-        # https://pagure.io/centos-infra/issue/79
-        if [ `umask` = 0000 ]; then
-          umask 0022
-        fi
-        ${cmds}
-    """
-}
+import org.yaml.snakeyaml.Yaml
 
-// Useful when we don't want to show confidential information in logs
-def shwrap_quiet(cmds) {
-    sh """
-        set +x -euo pipefail
-        # https://pagure.io/centos-infra/issue/79
-        if [ `umask` = 0000 ]; then
-          umask 0022
-        fi
-        ${cmds}
-    """
-}
-
-def shwrap_capture(cmds) {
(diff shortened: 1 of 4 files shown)


dustymabe (approved):
LGTM

utils.groovy
```diff
@@ -1,52 +1,9 @@
-import org.yaml.snakeyaml.Yaml;
+@Library('github.com/coreos/coreos-ci-lib@main') _
```
jlebon, line 1: I'm not 100% sure if this will work from within a separate module like this. But we can try it out and at worst, we'll just use `sh` directly in those few places for now.

jlebon:
(Will merge this tomorrow morning in case there's fallout.)

cgwalters:
Looks like this needs a rebase :surf

dustymabe:
Is there any way to factor out all of the `@Library('github.com/coreos/coreos-ci-lib@main') _` into a single place. I'm thinking about if I have an experimental branch of coreos-ci-lib that I want to test against. Then again, it's most likely I'd just be testing a single job and not all of them. Not sure, but figured I'd ask.

jlebon:
> Is there any way to factor out all of the `@Library('github.com/coreos/coreos-ci-lib@main') _` into a single place. I'm thinking about if I have an experimental branch of coreos-ci-lib that I want to test against. Then again, it's most likely I'd just be testing a single job and not all of them. Not sure, but figured I'd ask.

Yup that's possible, by putting it as a central lib. Though it'd need some more tweaks I can't easily test right now. Let's do that as a separate PR?

dustymabe:
This seems to be working in a dev pipeline I've got going. Let's merge and address any fallout in followups

dustymabe (approved):

PR #350, opened 2021-06-11
pipeline: print coreos-assembler-git.json

jlebon:
This is helpful for debugging.


cgwalters:
(I'd still like to consider building assembler in OpenShift and not quay, then we'd have nice git metadata in the container image itself)

PR #351, opened 2021-06-14
jobs/kola-openstack: disable concurrent builds

jlebon:
Run them one at a time so we don't hit VexxHost resource limits.


dustymabe (approved):
LGTM

PR #353, opened 2021-06-24
bump-lockfile: update metadata timestamp periodically

dustymabe:
If we don't have package updates for the contentset of FCOS for a few
days, which is unusual but not unheard of (right now we haven't had
an updated RPM for 5 days), then bump the timestamp anyway. It's an
indicator of freshness for FCOS so even if no packages have changed
let's update it to bump the date in the FCOS version numbers.

### DIFF
--- a/jobs/bump-lockfile.Jenkinsfile
+++ b/jobs/bump-lockfile.Jenkinsfile
@@ -35,10 +35,12 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
     """)
 
     def branch = params.STREAM
+    def timestampOnly = false
     shwrap("cosa init --branch ${branch} https://github.com/${repo}")
     shwrap("cosa buildprep ${BUILDS_BASE_HTTP_URL}/${branch}/builds")
 
     def prevPkgChecksum = shwrapCapture("jq -c .packages src/config/manifest-lock.*.json | sha256sum")
+    def prevPkgTimestamp = shwrapCapture("jq -c .metadata.generated src/config/manifest-lock.*.json | xargs -I{} date --date={} +%s")
 
     // do a first fetch where we only fetch metadata; no point in
     // importing RPMs if nothing actually changed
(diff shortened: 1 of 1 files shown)


jobs/bump-lockfile.Jenkinsfile
```diff
@@ -35,10 +35,12 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
     """)
 
     def branch = params.STREAM
+    def timestampOnly = false
     shwrap("cosa init --branch ${branch} https://github.com/${repo}")
     shwrap("cosa buildprep ${BUILDS_BASE_HTTP_URL}/${branch}/builds")
 
     def prevPkgChecksum = shwrapCapture("jq -c .packages src/config/manifest-lock.*.json | sha256sum")
+    def prevPkgTimestamp = shwrapCapture("jq -c .metadata.generated src/config/manifest-lock.*.json | xargs -I{} date --date={} +%s")
```
jlebon, line 43: The logic below assumes that it's a single value, but once we support multi-arch it'll actually be multiple newline-separated values. Let's just hardcode x86_64 for this? Even with multi-arch, I think we should enforce that the top-level `metadata.generated` timestamp be the same to retain our sanity, even if the per-yumrepo timestamps naturally differ.

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -47,65 +49,77 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
     }
 
     def newPkgChecksum = shwrapCapture("jq -c .packages src/config/manifest-lock.*.json | sha256sum")
+    def newPkgTimestamp = shwrapCapture("jq -c .metadata.generated src/config/manifest-lock.*.json | xargs -I{} date --date={} +%s")
```
jlebon, line 52: Here too.

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -47,65 +49,77 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
...
     if (newPkgChecksum == prevPkgChecksum) {
         println("No changes")
-        currentBuild.result = 'SUCCESS'
-        currentBuild.description = "[${params.STREAM}] 💤 (no change)"
-        return
+        if ((newPkgTimestamp - prevPkgTimestamp) > 172800) {
```
jlebon, line 55: This would be easier to read/verify as

```suggestion
        if ((newPkgTimestamp - prevPkgTimestamp) > (2*24*60*60)) {
```

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -47,65 +49,77 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
...
+            // Let's update the timestamp after two days even if no packages were updated.
+            // This will bump the date in the version number for FCOS, which is an indicator
+            // of how fresh the package set is.
+            println("2 days and no package updates. Force pushing to update timestamps.")
+            runTests = false // no need to run tests
```
jlebon, line 60: Unused variable AFAICT.

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -117,7 +131,11 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
           sh("git -C src/config push https://\${GHUSER}:\${GHTOKEN}@github.com/${repo} ${branch}")
         }
     }
-    currentBuild.description = "[${params.STREAM}] ⚡ (pushed)"
+    if (timestampOnly) {
+            currentBuild.description = "[${params.STREAM}] ⚡ (pushed timestamp update)"
```
jlebon, line 135: Extra indentation.

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -47,65 +49,77 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
...
-        shwrap("tar -cf - tmp/kola-testiso-metal/ | xz -c9 > ${env.WORKSPACE}/kola-testiso-metal.tar.xz")
-        shwrap("tar -cf - tmp/kola-testiso-metal4k/ | xz -c9 > ${env.WORKSPACE}/kola-testiso-metal4k.tar.xz")
-        archiveArtifacts allowEmptyArchive: true, artifacts: 'kola-testiso*.tar.xz'
     }
 
-    // OK, it passed kola: just push to the branch. In the future, we might be
+    // OK, we're ready to push: just push to the branch. In the future, we might be
     // fancier here; e.g. if tests fail, just open a PR, or if tests passed but a
     // package was added or removed.
     stage("Push") {
```
jlebon, line 125: Should we change the commit message based on whether it was a timestamp-only change? E.g. instead of "lockfiles: bump to latest", something like "lockfiles: bump timestamp".

jobs/bump-lockfile.Jenkinsfile
dustymabe: Yeah, we'll need to revisit this soon for multi-arch. Hardcoding to x86_64 for now.

dustymabe:
ok @jlebon I think I addressed all of your comments.

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -35,89 +35,110 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
...
-        shwrap("git -C src/config commit -am 'lockfiles: bump to latest' -m 'Job URL: ${env.BUILD_URL}' -m 'Job definition: https://github.com/coreos/fedora-coreos-pipeline/blob/main/jobs/bump-lockfile.Jenkinsfile'")
+        def message="lockfiles: bump to latest"
+        if (timestampOnly) {
+            message="lockfiles: bump timestamp"
+        }
+        shwrap("git -C src/config commit -am '$message' -m 'Job URL: ${env.BUILD_URL}' -m 'Job definition: https://github.com/coreos/fedora-coreos-pipeline/blob/main/jobs/bump-lockfile.Jenkinsfile'")
```
jlebon, line 129: I think this might work, but I'd go with

```suggestion
        shwrap("git -C src/config commit -am '${message}' -m 'Job URL: ${env.BUILD_URL}' -m 'Job definition: https://github.com/coreos/fedora-coreos-pipeline/blob/main/jobs/bump-lockfile.Jenkinsfile'")
```

to be safe.

http://groovy-lang.org/syntax.html#_string_interpolation says:

> The placeholder expressions are surrounded by ${}. The curly braces may be omitted for unambiguous dotted expressions, i.e. we can use just a $ prefix in those cases.

But here we're not using a dotted expression, so I'm not sure if it'll still interpolate or not.

jlebon (approved):
One final nit, LGTM otherwise!

jobs/bump-lockfile.Jenkinsfile
dustymabe: fixed

dustymabe:
This finally was exercised: https://github.com/coreos/fedora-coreos-config/commit/1d8cebb0991c174d5607b7e99a5522bc488562f5 🎉

PR #354, opened 2021-06-25
bump-lockfile: convert timestamps into integers

dustymabe:
So they can be used in arithmetic.


dustymabe:
@jlebon approved this patch yesterday before I opened the PR: 

Screenshot_2021-06-25 Slack Jonathan Lebon CoreOS

Merging....

PR #357, opened 2021-07-19
pipeline: Ship Azure Stack

cgwalters:
This was added in https://github.com/coreos/coreos-assembler/pull/1566
and recently enabled for RHCOS.  Let's ship it here too to keep
things aligned upstream/downstream.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -391,7 +391,7 @@ lock(resource: "build-${params.STREAM}") {
 
             // parallel build these artifacts
             def pbuilds = [:]
-            ["Aliyun", "AWS", "Azure", "DigitalOcean", "Exoscale", "GCP", "IBMCloud", "OpenStack", "VMware", "Vultr"].each {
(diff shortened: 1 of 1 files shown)


bgilbert:
We also need to update `cmd-generate-release-meta`, fedora-coreos-stream-generator, the Go and Rust stream libraries, the website code, and the build browser.

cgwalters:
The stream metadata bindings in both Go and Rust are stringly-typed, not an enum or equivalent.  So we don't need to update those.

Some other PRs are linked from https://github.com/coreos/stream-metadata-go/issues/26#issuecomment-882845836

bgilbert:
Ah, yeah, you're right.

jlebon:
Browser support: https://github.com/coreos/fedora-coreos-browser/pull/28
Website support: https://pagure.io/fedora-web/websites/pull-request/192

cgwalters:
And we need https://github.com/coreos/coreos-assembler/pull/2308

jlebon (approved):
Looks like we're ready to do this now!

PR #358, opened 2021-07-20
Add instructions for deploying a multi-arch builder.

dustymabe:
We can use these machines via gangplank to do our multi-arch builds
remotely. The first will be an aarch64 builder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+fcos-aarch64-builder.ign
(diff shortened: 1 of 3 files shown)


dustymabe:
Marking for draft for now as I tweak things.

cgwalters (approved):
Seems sane!

dustymabe:
Thanks @cgwalters. I'll post a followup if I need to tweak anything.

PR #361, opened 2021-07-28
bump-lockfile: force gangplank to not upload artifacts

dustymabe:
We don't want the bits, we just need to know if they passed tests or
not. Currently Gangplank doesn't support this [1] but we can force
it by just deleting the builds.json in a post command.

[1] https://github.com/coreos/coreos-assembler/issues/2317

### DIFF
--- a/jobs/bump-lockfile.Jenkinsfile
+++ b/jobs/bump-lockfile.Jenkinsfile
@@ -133,6 +133,7 @@ stages:
   build_artifacts: [base]
   post_commands:
     - cosa kola run --basic-qemu-scenarios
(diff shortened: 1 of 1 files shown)


jlebon (approved):
Hmm, seems sane to me. Though feels like for the main pipeline build, we'll need to figure out how updating `builds.json` will work.

dustymabe:
> Hmm, seems sane to me. Though feels like for the main pipeline build, we'll need to figure out how updating `builds.json` will work.

Yeah. I think it won't be a big problem initially because we're only planning to trigger the `aarch64` pipeline after `x86_64` has been built and smoketested. It will be a problem when we add a 3rd arch, though.

PR #365, opened 2021-08-05
release: automate waiting for import and pull test

jlebon:
In the FCOS release process, we have this step after the release job
runs where we want to wait until the OSTree commit was successfully
imported into the production repo and verify that it can be pulled and
that it's signed (though that last bit is also implicit with the ability
to even pull).

This requires human releasers to download the previous QEMU image, then
run `rpm-ostree upgrade` multiple times until it succeeds, and then look
at `rpm-ostree status`.

This patch enhances the release job to automate all that so that the
releaser just needs to verify the output.

This uses the new `cosa run -x` added in
https://github.com/coreos/coreos-assembler/pull/2338.

### DIFF
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -70,6 +70,7 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
 
         def s3_stream_dir = "${s3_bucket}/prod/streams/${params.STREAM}"
         def gcp_image = ""
+        def ostree_prod_refs = [:]
 
         // Clone the automation repo, which contains helper scripts. In the
         // future, we'll probably want this either part of the cosa image, or
@@ -103,6 +104,8 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
                         --fedmsg-conf=/etc/fedora-messaging-cfg/fedmsg.toml
                     """)
                 }
+
+                ostree_prod_refs[meta.ref] = meta["ostree-commit"]
             }
 
             // For production streams, promote the GCP image so that it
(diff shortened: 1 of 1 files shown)


jlebon:
I did test the `coreos-installer download` and `cosa run -x` commands manually, but I didn't test this as part of a bona fide release job.

cgwalters:
This seems sane, but I probably would have just run `ostree pull` directly from the container image instead of spawning a VM.

jlebon:
Yeah, it's debatable I guess. I wanted to maintain the "end-to-end-ness" of the human check. E.g. we know for sure we're testing the pull with the right repo config, signatures, etc... just as a client machine would. And now with `cosa run -x`, it's pretty trivial to spawn a VM to do this.

jlebon:
Let's hold this until I'm back online and I can help deal with any fallout.

jlebon:
Ahh right, I need to rebase this on the multi-arch work now.

Jenkinsfile.release
```diff
@@ -172,5 +175,33 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
...
+                ostree init --mode=archive --repo=.
+                cat /etc/ostree/remotes.d/fedora.conf >> config
+                """)
+
+                timeout(time: 20, unit: 'MINUTES') {
+                    for (ref in ostree_prod_refs) {
+                        shwrap("""
+                        cd ${tmpd}
+                        while true; do
+                            ostree pull --commit-metadata-only fedora:${ref.key}
```
dustymabe, line 194: might be worth mentioning here in a comment that if there is no signature or the signature is invalid then the pull will fail (correct?) which is why we can rely on this and remove the manual step from the release checklist

Jenkinsfile.release
```diff
@@ -172,5 +175,33 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
...
+            stage("OSTree Import: Wait and Verify") {
+                def tmpd = shwrapCapture("mktemp -d")
+
+                shwrap("""
+                cd ${tmpd}
+                ostree init --mode=archive --repo=.
+                cat /etc/ostree/remotes.d/fedora.conf >> config
+                """)
+
+                timeout(time: 20, unit: 'MINUTES') {
```
dustymabe, line 189: let's add a comment here on why we need the timeout. basically it takes some time for the infra changes to propagate and the summary file to get updated

Jenkinsfile.release
```diff
@@ -172,5 +175,33 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
...
+                    for (ref in ostree_prod_refs) {
+                        shwrap("""
+                        cd ${tmpd}
+                        while true; do
+                            ostree pull --commit-metadata-only fedora:${ref.key}
+                            chksum=\$(ostree rev-parse fedora:${ref.key})
+                            if [ "\${chksum}" == "${ref.value}" ]; then
+                                break
+                            fi
+                            sleep 10
```
dustymabe, line 199: this seems a little aggressive - maybe `sleep 30` or `sleep 60` instead?

dustymabe (approved):
LGTM requested one small change and a few added comments

Jenkinsfile.release
jlebon: This was implicit a bit in the commit message, but I mentioned it in the code and made it more explicit in the commit message too.

dustymabe (approved):

jlebon:
Let's merge after releases.

PR #368, opened 2021-08-06
Add multi-arch-pipeline job

dustymabe:
This is the first attempt at adding a multi-arch pipeline. We'll
call the multi-arch-pipeline job from the main pipeline after QEMU
tests have passed.

This is currently mostly a copy/paste of the original pipeline. In
the future we want to factor out these bits and re-use them between
the main pipeline and the multi-arch-pipeline job.

### DIFF
--- a/jobs/multi-arch-pipeline.Jenkinsfile
+++ b/jobs/multi-arch-pipeline.Jenkinsfile
@@ -2,12 +2,15 @@
 
 import org.yaml.snakeyaml.Yaml;
 
-def pipeutils, streams, official, official_jenkins, developer_prefix, src_config_url, src_config_ref, s3_bucket
+def pipeutils, streams, official, official_jenkins, developer_prefix, repo, src_config_url, src_config_ref, s3_bucket, gp
 node {
     checkout scm
     pipeutils = load("utils.groovy")
     streams = load("streams.groovy")
     pod = readFile(file: "manifests/pod.yaml")
+    gp = load("gp.groovy")
+    repo = "coreos/fedora-coreos-config"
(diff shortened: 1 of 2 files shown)


dustymabe:
This is a rollup of a few other PRs. I'll remove those commits when those PRs merge. Focus on the last two commits for now.

jobs/multi-arch-pipeline.Jenkinsfile
```diff
@@ -0,0 +1,493 @@
...
+// here; we can look into making them configurable through the template if
+// developers really need to tweak them (note that in the default minimal devel
+// workflow, only the qemu image is built).
+def cosa_memory_request_mb
+if (official) {
+    cosa_memory_request_mb = 6.5 * 1024
+} else {
+    cosa_memory_request_mb = 2.5 * 1024
+}
+cosa_memory_request_mb = cosa_memory_request_mb as Integer
```
jlebon, line 111: Since everything happens on the remote, we don't actually need to allocate a lot of memory here.

jlebon (approved):
This is OK to start with, but we need to quickly move towards sharing the main business logic.

Until then, I wish there was a way to ensure that any changes done in the primary pipeline are also done in the multi-arch version if necessary. I don't have a good idea for enforcing that though apart from some large comment blocks at the beginning of both files explaining the situation.

jobs/multi-arch-pipeline.Jenkinsfile
dustymabe: yeah i can bump it down +1

PR #369, opened 2021-08-11
utils.groovy: update get_annotation() to be job independent

dustymabe:
Right now it tries to guess the name of the buildconfig based on
the name of the job that's being run. With the introduction of the
multi-arch pipeline there is no BuildConfig that backs the job, so
we'd have to do heuristics to find the main pipeline job name.

Instead of doing that let's just assume that there's only one
BuildConfig in the project with the needed information and use that
so that it doesn't matter what jenkins job we're running we can
still find the information we're looking for.

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -5,18 +5,28 @@ import org.yaml.snakeyaml.Yaml
 // Only add pipeline-specific things here. Otherwise add to coreos-ci-lib
 // instead.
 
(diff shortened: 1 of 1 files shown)


dustymabe:
Note.. we should probably use configmaps or something else other than annotations to hold information like this in the future so we don't have to guess a job name.

PR #370, opened 2021-08-11
multi-arch-pipeline: handle "official" logic different

dustymabe:
The job name for this job won't be *fedora-coreos-pipeline so let's
treat it differently and just queue off of the jekins URL.


ravanelli (approved):
/lgtm

dustymabe:
> Let's just do this in the main pipeline too to be consistent?

yeah, honestly I was thinking of a larger change to the pipeline and wanted to discuss it with you first.. was wondering if we could just drop all of the "developer_prefix" stuff - let's discuss in IRC

PR #372, opened 2021-08-11
multi-arch-pipeline: don't pass --arch twice

dustymabe:
We're already passing it earlier in the cosa sign command invocation.


darkmuggle (approved):
/lgtm

PR #373, opened 2021-08-11
utils: cp builds.json with public access

dustymabe:
Otherwise things that pull from it (like the builds browser)
will break.


darkmuggle (approved):
/lgtm

PR #376, opened 2021-08-16
seed: fix call to get_annotation

dustymabe:
The call interface changed in 13b9d25 to just take one argument.

### DIFF
--- a/jenkins/config/seed.yaml
+++ b/jenkins/config/seed.yaml
@@ -18,8 +18,8 @@ jobs:
                 // XXX: hack, should put this in coreos-ci-lib
                 sh("curl -LO https://raw.githubusercontent.com/coreos/fedora-coreos-pipeline/main/utils.groovy")
                 def pipeutils = load("utils.groovy")
-                def url = pipeutils.get_annotation("jenkins", "jenkins-jobs-url")
-                def ref = pipeutils.get_annotation("jenkins", "jenkins-jobs-ref")
+                def url = pipeutils.get_annotation("jenkins-jobs-url")
+                def ref = pipeutils.get_annotation("jenkins-jobs-ref")
                 shwrap("rm -rf source")
                 shwrap("git clone -b ^${ref} ^${url} source")
 


jlebon:
The problem is deeper than this. The seed job is querying the `jenkins` BC, while the pipeline jobs are querying the `fedora-coreos-pipeline` BC. That same seed job is for example used by https://github.com/coreos/coreos-ci; it just sets up the annotations to point to its repo: https://github.com/coreos/coreos-ci/blob/b0d1323751ef4793bb4e8395c0acc2497d95a950/HACKING.md#L72.

Maybe easiest is to revert https://github.com/coreos/fedora-coreos-pipeline/commit/13b9d25700dfc29f30f02ab4d614eaad99c6b0e0 and instead move the pipeline bits which require annotations to use the Jenkins configmap instead to create global env vars.

dustymabe:
> The problem is deeper than this. The seed job is querying the `jenkins` BC, while the pipeline jobs are querying the `fedora-coreos-pipeline` BC. That same seed job is for example used by https://github.com/coreos/coreos-ci; it just sets up the annotations to point to its repo: https://github.com/coreos/coreos-ci/blob/b0d1323751ef4793bb4e8395c0acc2497d95a950/HACKING.md?plain=1#L72.

Hmm. I still think what I propose should work, we'd just need to update coreos-ci to also remove the second argument to the fucntion call. 

> 
> Maybe easiest is to revert [13b9d25](https://github.com/coreos/fedora-coreos-pipeline/commit/13b9d25700dfc29f30f02ab4d614eaad99c6b0e0) and instead move the pipeline bits which require annotations to use the Jenkins configmap instead to create global env vars.

Yes, totally agree that using a configmap instead would be the right move here. I also think we should rip out the developer prefix handling and logic. I think it hurts more than it helps.

jlebon (approved):
OK right yup, agreed this should work. I hadn't fully understood https://github.com/coreos/fedora-coreos-pipeline/commit/13b9d25700dfc29f30f02ab4d614eaad99c6b0e0 at first.

But yeah, let's work towards moving to a configmap which then should make it easier to move the pipeline to a regular Jenkins job.

PR #377, opened 2021-08-17
multi-arch-pipeline: start building other artifacts for aarch64

dustymabe:
This adds the metal artifacts as well as the openstack image.

### DIFF
--- a/jobs/multi-arch-pipeline.Jenkinsfile
+++ b/jobs/multi-arch-pipeline.Jenkinsfile
@@ -285,13 +285,20 @@ stages:
 - id: ExecOrder 1 Stage
   execution_order: 1
   description: Stage 1 execution base
-  build_artifacts: [qemu]
-  require_artifacts: [ostree]
   prep_commands:
     - cat /cosa/coreos-assembler-git.json
+  require_artifacts: [ostree]
   post_commands:
-    - cosa kola run --basic-qemu-scenarios
+    - cosa buildextend-qemu
+    - cosa kola run --basic-qemu-scenarios --output-dir tmp/kola-basic
+    - cosa kola run --parallel 4 --output-dir tmp/kola
+    - cosa buildextend-metal
+    - cosa buildextend-metal4k
+    - cosa buildextend-live
+    - kola testiso -S --output-dir tmp/kola-metal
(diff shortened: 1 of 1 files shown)


dustymabe:
There are better ways to do this. We'll start with this for now.

jobs/multi-arch-pipeline.Jenkinsfile
```diff
@@ -285,13 +285,20 @@ stages:
...
+  require_artifacts: [ostree]
   post_commands:
+    - cosa buildextend-qemu
     - cosa kola run --basic-qemu-scenarios
+    - cosa kola run --parallel 4
+    - cosa buildextend-metal
+    - cosa buildextend-metal4k
+    - cosa buildextend-live
+    - kola testiso -S --output-dir tmp/kola-metal
+    - kola testiso -SP --qemu-native-4k --scenarios iso-install --output-dir tmp/kola-metal4k
```
jlebon, line 299: Are the output dirs here copied off the worker so we can archive them? Feels like we're also missing the `archiveArtifacts` step here.

dustymabe: agree, yes they are copied back. I've got a few more followups to make (kola aws uploads and testing) and some cycles to run in my dev jenkins. Can I add that step then?

dustymabe: actually there is some more I need to do to get them to get copied back - but yeah will do that in a followup

jobs/multi-arch-pipeline.Jenkinsfile
```diff
@@ -285,13 +285,20 @@ stages:
...
   description: Stage 1 execution base
-  build_artifacts: [qemu]
-  require_artifacts: [ostree]
   prep_commands:
     - cat /cosa/coreos-assembler-git.json
+  require_artifacts: [ostree]
   post_commands:
+    - cosa buildextend-qemu
     - cosa kola run --basic-qemu-scenarios
+    - cosa kola run --parallel 4
```
dustymabe, line 294: Do these two write/overwrite the same directory? Do I need to specify `--output-dir` here too?

dustymabe: updated it so that they all specify `--output-dir`

jlebon (approved):

PR #378, opened 2021-08-17
seed: don't redefine existing jobs by default

jlebon:
Otherwise, we lose all the triggers defined in those jobs. This is a
natural consequence of pipeline jobs being able to describe their own
triggers as imperative code.

Add a `FORCE` parameter if we really want to update existing jobs, with
a helper text about the trigger subtlety.


dustymabe (approved):
LGTM

PR #379, opened 2021-08-17
streams: re-enable `branched` stream

jlebon:
Fedora 35 has branched now, so let's track it.


jlebon:
Requires: https://github.com/coreos/fedora-coreos-config/pull/1166

dustymabe (approved):
LGTM

PR #380, opened 2021-08-19
kola-aws,kola-gcp: support arch, let kola inspect meta.json

dustymabe:
```
commit 3af893efef8095bf58e92f01298c7e02a6502f0e
Date:   Thu Aug 19 16:49:52 2021 -0400

    kola-gcp: support arch, get image from meta.json
    
    Similar to what we just did for AWS let's support passing in the
    architecture (even though GCP only supports one) and also let kola
    pick up the image from the meta.json rather than digging for it here
    in the job.

commit 475f622705ba80d4327f849d48c19c2f65d8e407
Date:   Tue Aug 10 15:53:36 2021 -0400

    kola-aws: allow kola to introspect instance type and AMI
    
    kola recently gained the ability to pick up the AMI ID from
    the specified build metadata and also switch the instance type
    based on the target architecture. Let kola do the heavy lifting
    here for us.
    
    See https://github.com/coreos/coreos-assembler/pull/2353

commit 32d349bf5372082dfca4be614cb98d870436023c
Date:   Mon Aug 9 17:12:34 2021 -0400

    kola-aws: support passed in architecture
    
    Also manually override the instance type for aarch64 because the
    default m5.large won't work.
```

### DIFF
--- a/jobs/kola-aws.Jenkinsfile
+++ b/jobs/kola-aws.Jenkinsfile
@@ -17,6 +17,10 @@ properties([
              description: 'Fedora CoreOS Build ID to test',
              defaultValue: '',
              trim: true),
+      string(name: 'ARCH',
+             description: 'Target architecture',
+             defaultValue: 'x86_64',
+             trim: true),
       string(name: 'S3_STREAM_DIR',
              description: 'Override the Fedora CoreOS S3 stream directory',
              defaultValue: '',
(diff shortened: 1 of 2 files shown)


jobs/kola-aws.Jenkinsfile
```diff
@@ -43,28 +47,19 @@ cosaPod(image: params.COREOS_ASSEMBLER_IMAGE,
         memory: "256Mi", kvm: false,
         secrets: ["aws-fcos-builds-bot-config", "aws-fcos-kola-bot-config"]) {
 
-    def ami, ami_region
+    def ami, ami_region, instance_type_arg
```
jlebon, line 50: Unused arg

jlebon (approved):
One nit, LGTM otherwise.

dustymabe:
requires https://github.com/coreos/coreos-ci-lib/pull/84

PR #381, opened 2021-08-23
Jenkinsfile: limit multi-arch-pipeline to non production builds

dustymabe:
We're not quite ready for them on the production streams just yet.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -371,14 +371,18 @@ lock(resource: "build-${params.STREAM}") {
             }
         }
 
-        stage('Fork AARCH64 Pipeline') {
-            build job: 'multi-arch-pipeline', wait: false, parameters: [
-                booleanParam(name: 'FORCE', value: params.FORCE),
-                booleanParam(name: 'MINIMAL', value: params.MINIMAL),
-                string(name: 'STREAM', value: params.STREAM),
-                string(name: 'VERSION', value: newBuildID),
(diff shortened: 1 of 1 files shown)


cgwalters (approved):

Jenkinsfile
```diff
@@ -371,14 +371,18 @@ lock(resource: "build-${params.STREAM}") {
...
-            build job: 'multi-arch-pipeline', wait: false, parameters: [
-                booleanParam(name: 'FORCE', value: params.FORCE),
-                booleanParam(name: 'MINIMAL', value: params.MINIMAL),
-                string(name: 'STREAM', value: params.STREAM),
-                string(name: 'VERSION', value: newBuildID),
-                string(name: 'ARCH', value: 'aarch64')
-            ]
+        // Don't run the multi-arch-pipeline for prod streams just
+        // yet. We're still working out all the details.
+        if (official && !(params.STREAM in streams.production)) {
```
jlebon, line 376: Don't we want

```suggestion
        if (!official || !(params.STREAM in streams.production)) {
```

so that it still runs in developer envs?

jlebon (commented):
Typo in commit title.

Jenkinsfile
dustymabe: fixed

dustymabe:
> Typo in commit title.

fixed

PR #385, opened 2021-08-31
multi-arch-pipeline: drive exact config git commit for arch pipeline runs

dustymabe:
multi-arch pipeline runs execute some time after the x86_64 pipeline
runs. Let's make sure they run from the same config git version to
mitigate any drift that could have happened.

This is mostly useful for the devel branches that change often. It's
less likely to affect the prod branches that don't really change much
(especially over time course of a release).

It also isn't really that useful for our mechanical refs that don't have
full lockfile definitions as a change on the yum repo side would still
creep in. Either way it's at least useful for a development refs/branches.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -2,7 +2,8 @@
 
 import org.yaml.snakeyaml.Yaml;
 
-def pipeutils, streams, official, official_jenkins, developer_prefix, src_config_url, src_config_ref, s3_bucket
+def pipeutils, streams, official, official_jenkins, developer_prefix
+def src_config_url, src_config_ref, s3_bucket, config_git_commit
 node {
     checkout scm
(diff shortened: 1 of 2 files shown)


dustymabe:
also:

```
multi-arch-pipeline: fix tar invocation and reorder post_commands

The previous iteration `ls` was actually returning a bad exit code
if some directories didn't exist, which meant the first list item
in `post_commands` would fail and the compress wouldn't happen. Let's
move the compress to the first step and also change the tar command
to use `|| true` so it won't "fail" if we add any more post commands
in the future.

Note that we can get rid of the `compress` in the future when we have
https://github.com/coreos/coreos-assembler/issues/2389

```

miabbott (approved):
Looks sane to me

dustymabe:
thanks @miabbott

PR #389, opened 2021-09-07
Test ISO boot on UEFI in pipeline, and ISO boot from CD/HD on BIOS/UEFI when bumping lockfile

bgilbert:


cgwalters (approved):
Nice!

Jenkinsfile
```diff
@@ -420,12 +420,17 @@ lock(resource: "build-${params.STREAM}") {
                         shwrap("kola testiso -S --output-dir tmp/kola-metal")
                     }, metal4k: {
                         shwrap("kola testiso -SP --qemu-native-4k --output-dir tmp/kola-metal4k")
+                    }, uefi: {
+                        shwrap("mkdir -p tmp/kola-uefi")
+                        shwrap("kola testiso -S --qemu-firmware=uefi --scenarios iso-live-login,iso-as-disk --output-dir tmp/kola-uefi/insecure")
+                        shwrap("kola testiso -S --qemu-firmware=uefi-secure --scenarios iso-live-login,iso-as-disk --output-dir tmp/kola-uefi/secure")
```
cgwalters, line 426: I am thinking we need to deduplicate this across repos, and since we aren't likely to go Gangplank-only in the near future (and even if we are?) it may make the most sense to add at least e.g. `cosa kola testiso` that auto-generates `--output-dir` for example.

Or maybe better, have `kola testiso --all` or something?

bgilbert: I think that makes sense.  https://github.com/coreos/coreos-ci-lib/pull/88 was intended to perform that deduplication for upstream repos (not affecting the FCOS or RHCOS pipelines) but doing it in cosa would be even more general.

PR #391, opened 2021-09-13
bump-lockfile: force build if we get there

dustymabe:
If we end up pushing an update even if no package changed
(see 544a2cc) then we may need to force the cosa build so it doesn't
re-use the previous iteration. If it does that then it will fail
because the previous build had been compressed already:

```
harness.go:1124: Cluster failed starting machines:
    lstat /home/jenkins/agent/workspace/bump-lockfile/builds/35.20210909.10.0/x86_64/fedora-coreos-35.20210909.10.0-qemu.x86_64.qcow2.xz:
    no such file or directory
```

### DIFF
--- a/jobs/bump-lockfile.Jenkinsfile
+++ b/jobs/bump-lockfile.Jenkinsfile
@@ -149,7 +149,7 @@ EOF
             }
 
             stage("Build") {
-                shwrap("cosa build --strict")
+                shwrap("cosa build --force --strict")
             }
 
             fcosKola(cosaDir: env.WORKSPACE)


jlebon (approved):
Thanks! For the gangplank side, do we need something similar or does it always force?

dustymabe:
> For the gangplank side, do we need something similar or does it always force?

I think we don't hit this on the gangplank side because there's no local cache, so `rpm-ostree` is composing into a fresh repo and the "nothing changed" logic doesn't apply.

PR #393, opened 2021-09-13
kola-{aws,gcp,openstack}: don't run basic qemu tests

dustymabe:
The fcosKola library function was changed to default to running basic
qemu tests in [1]. Let's skip those in our cloud provider tests because
obviously we're not testing in qemu.

[1] https://github.com/coreos/coreos-ci-lib/commit/33a06a663519315877e54c9e123627b03ff50826


jlebon (approved):
Makes sense. We should probably fold the basic scenarios stuff into kola, and then on any platform other than qemu it'll do something more sensible. E.g. on AWS, testing different instance types, etc...

Issue #352, opened 2021-06-15
Webhook events from coreos/fedora-coreos-streams to prod Jenkins failing with "Error 400 Provided signature [...] did not match to calculated"


jlebon:
```
<html>
<head>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8"/>
<title>Error 400 Provided signature [<hash>] did not match to calculated</title>
</head>
<body><h2>HTTP ERROR 400 Provided signature [<hash>] did not match to calculated</h2>
<table>
<tr><th>URI:</th><td>/github-webhook/</td></tr>
<tr><th>STATUS:</th><td>400</td></tr>
<tr><th>MESSAGE:</th><td>Provided signature [<hash>] did not match to calculated</td></tr>
<tr><th>SERVLET:</th><td>Stapler</td></tr>
</table>
<hr><a href="https://eclipse.org/jetty">Powered by Jetty:// 9.4.33.v20201020</a><hr/>

</body>
</html>
```

jlebon:
OK yup, looks like fallout from https://github.com/jenkinsci/github-plugin/pull/242, which was reverted in https://github.com/jenkinsci/github-plugin/pull/246, which is in the Github plugin v1.33.1. We're at v1.33.0. Let's update it.

jlebon:
Let's see if we can fix it in the source image itself: https://github.com/openshift/jenkins/pull/1279. If it takes too long, we can bump it here in the interim.

dustymabe:
hit this in `coreos-ci` today. I had to restart it and use `emptyDir` to get it running because of https://pagure.io/centos-infra/issue/423 (we'll need to get the PVC unbound before we can switch back to actual storage).

Do we need to apply https://github.com/openshift/jenkins/pull/1279 to coreos-ci repo ?

cgwalters:
Heh awesome.  Short term, perhaps we can just strip all unicode from our repo descriptions?  I would be very sad, but I am more sad about CI not working.

dustymabe:
> (we'll need to get the PVC unbound before we can switch back to actual storage)

also reminder here that because of https://pagure.io/centos-infra/issue/423 I applied a nodeSelector to our jenkins container definition so it wouldn't select the bad node (and thus never come up). Let's remove the nodSelector from the container definition when we move it back to being backed by the PV (and thus are doing a restart anyway).

jlebon:
1.33.1 is now on CoreOS CI.

[...]

PR #394, opened 2021-09-13
plugins: bump github to 1.33.1

jlebon:
It contains a fix for #352. Once the base Jenkins imagestream is updated
in our cluster, we can drop this so we don't lock on 1.33.1.


dustymabe (approved):
LGTM

PR #395, opened 2021-09-13
plugins: bump workflow-api to 2.46

jlebon:
Otherwise, the current version (2.40) is too old for the timestamper
plugin to work with.


dustymabe (approved):
LGTM - I guess these all get cleaned up when we move to a newer container?

jlebon:
> LGTM - I guess these all get cleaned up when we move to a newer container?

Yeah, though sadly we need to manually check the version to see what is safe to drop and what isn't. We need something like FCOS override fast-tracks for Jenkins. :)

PR #398, opened 2021-09-21
Use PERFORMANCE_OPTIMIZED for all jobs

jlebon:
Otherwise, the default is to have Jenkins retain the state of jobs so
that it can resume them in the event of a loss of power/restart. But we
don't really need that feature because our jobs are designed to be
idempotent and e.g. we need to be able to handle restarting from
scratch.

So disable it so we can win the performance gains.

Prompted by recently-added (#365) code which breaks the job because
Jenkins isn't capable of serializing all the state to disk.

### DIFF
--- a/Jenkinsfile.development
+++ b/Jenkinsfile.development
@@ -5,7 +5,8 @@ node {
 }
 
 properties([
-    pipelineTriggers(streams.get_push_trigger())
+    pipelineTriggers(streams.get_push_trigger()),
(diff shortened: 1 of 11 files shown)


jlebon:
For context: note that a few of our jobs (notably, the primary pipeline job) were already using this. This PR is just applying everywhere else.

dustymabe (approved):
LGTM

PR #399, opened 2021-09-22
manifest/jenkins: bump HEARTBEAT_CHECK_INTERVAL to 15 mins

jlebon:
We've been seeing the dreaded "process apparently never started" error
more frequently lately, which is likely due to the cluster being under
heavier load.

Let's bump the check interval to 15 minutes to be even more tolerant.


dustymabe (approved):

jlebon:
If agreed, I'll just push this patch to CoreOS CI as well upon merge.

PR #402, opened 2021-09-23
multi-arch-pipeline: turn on testing for aarch64 openstack images

dustymabe:
We can test aarch64 OpenStack images in VexxHost same
as our x86_64 images.


dustymabe:
marking as draft because this is currently blocked on the kernel not panicking on boot (tracked over in https://github.com/coreos/fedora-coreos-tracker/issues/965)

PR #405, opened 2021-09-27
kola-{aws,gcp,openstack}: use a specific config git commit for tests

dustymabe:
This will allow us to only run tests that were for the specific config
repo version that we built against. This prevents issues like what we saw
recently where tests got kicked off later and the config repo had merged
a new change that had a test. The test failed because the test was ran
but the corresponding needed change was not in the built FCOS version.

### DIFF
--- a/jobs/kola-aws.Jenkinsfile
+++ b/jobs/kola-aws.Jenkinsfile
@@ -32,7 +32,11 @@ properties([
       string(name: 'COREOS_ASSEMBLER_IMAGE',
              description: 'Override the coreos-assembler image to use',
              defaultValue: "coreos-assembler:main",
-             trim: true)
+             trim: true),
+      string(name: 'CONFIG_GIT_COMMIT',
+             description: 'The exact config repo git commit to run tests against',
+             defaultValue: '',
+             trim: true),
     ]),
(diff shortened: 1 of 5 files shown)


jobs/kola-aws.Jenkinsfile
```diff
@@ -32,7 +32,11 @@ properties([
       string(name: 'COREOS_ASSEMBLER_IMAGE',
              description: 'Override the coreos-assembler image to use',
              defaultValue: "coreos-assembler:main",
-             trim: true)
+             trim: true),
+      string(name: 'CONFIG_GIT_COMMIT',
```
jlebon, line 36: Minor/optional: WDYT about renaming this `FCOS_CONFIG_COMMIT` everywhere instead? We can do that as a follow-up.

jlebon (approved):
LGTM!

jobs/kola-aws.Jenkinsfile
dustymabe: > Minor/optional: WDYT about renaming this `FCOS_CONFIG_COMMIT` everywhere instead? We can do that as a follow-up.

i'm not opposed - want to open a follow-up PR?

PR #406, opened 2021-09-27
Rename `CONFIG_GIT_COMMIT` parameter to `FCOS_CONFIG_COMMIT`

jlebon:
Makes it more clear that it's specifically a commit from
https://github.com/coreos/fedora-coreos-config.


dustymabe (approved):
LGTM

PR #407, opened 2021-09-30
plugins: fast-track ssh-credentials v1.19

jlebon:
We're currently at v1.18.1. And while the changelogs don't mention
anything to that effect, let's fast-track the latest and see if it helps
https://github.com/jenkinsci/configuration-as-code-plugin/issues/1646.


dustymabe (approved):
LGTM

PR #409, opened 2021-10-05
Keep artifacts for all builds

jlebon:
We no longer archive heavy files like disk images, just kola logs. Those
should be small so let's disable artifact discarding so they're always
available for all builds, until the builds themselves are discarded.


dustymabe:
How big is our backing store? are we concerned at all?

dustymabe (approved):
LGTM

jlebon:
> How big is our backing store? are we concerned at all?

OK yeah, I was looking at this and the PVC is 5G, which is not small, but not very large either. At its max, we're looking at 2 jobs * 60 builds per job * 15M per build = ~1.8G of kola logs which is a bit much. So dropped it to 20 now to be safe.

saqibali-2k:
Clarification: the commit message says we want to keep artifacts for 20 builds while the title says we want to keep all builds? 

Otherwise lgtm!

PR #411, opened 2021-10-06
Revert "kola-openstack: switch to sjc1 region"

dustymabe:
The networking issues in the region have now been fixed. According
to the rep: " there was an issue in our internal BGP E-VPN control plane
that was causing this issue, this has been resolved and it's now
working".

Let's switch back to ca-ymq-1 since the infra is a little faster and
instances were less likely to time out there.

This reverts commit c4bdf1f42f70d9ac2bbd5a1ce2825c9d9581b854.


dustymabe:
marking as draft since I need to press the merge button in order to coordinate this change with a config change as well.

bgilbert (approved):

PR #412, opened 2021-10-13
multi-arch-pipeline: tell gangplank to copy the parent build metadata

dustymabe:
This works around corner cases like described in:
https://github.com/coreos/coreos-assembler/issues/2380

### DIFF
--- a/jobs/multi-arch-pipeline.Jenkinsfile
+++ b/jobs/multi-arch-pipeline.Jenkinsfile
@@ -244,6 +244,7 @@ recipe:
   git_ref: ${params.STREAM}
   git_url: https://github.com/${repo}
   git_commit: ${params.FCOS_CONFIG_COMMIT}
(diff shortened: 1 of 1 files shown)


jobs/multi-arch-pipeline.Jenkinsfile
```diff
@@ -244,6 +244,7 @@ recipe:
   git_ref: ${params.STREAM}
   git_url: https://github.com/${repo}
   git_commit: ${params.FCOS_CONFIG_COMMIT}
+copy-build: ${parent_version}
```
jlebon, line 247: Will gangplank correctly handle the empty string case? (Which obviously won't happen very often, just when starting up new streams, but it's good to be resilient to it.)

dustymabe: yep - it handles that.

PR #413, opened 2021-10-14
plugins: add pipeline-github

jlebon:
This plugin will make it easier to interact with GitHub. E.g. adding
comments, adding additional CI contexts, etc...

Planning to use this for
https://github.com/coreos/fedora-coreos-tracker/issues/990 but we could
also use this for enhancing other things, e.g. adding a comment with the
RPM diff, or having a separate CI context if a downgrade is detected to
make it easier to tell, etc...


bgilbert (approved):
LGTM

Issue #414, opened 2021-10-18
Run kola AWS tests on different instance types


jlebon:
This would've caught https://github.com/coreos/fedora-coreos-tracker/issues/997. We suspect it's hardware/hypervisor-related. Would be enough to start to just run `basic` on these other instance types.

PR #416, opened 2021-10-19
kola-aws: also run on m4.large

jlebon:
Likely this should live in a `fcosKolaAws` wrapper instead, but at least
for the short-term it's trivial to get coverage for Xen and Nitro types
so let's do that.

This would've caught
https://github.com/coreos/fedora-coreos-tracker/issues/997.

Closes: #414

### DIFF
--- a/jobs/kola-aws.Jenkinsfile
+++ b/jobs/kola-aws.Jenkinsfile
@@ -73,6 +73,23 @@ try { timeout(time: 90, unit: 'MINUTES') {
                     --aws-credentials-file=\${AWS_FCOS_KOLA_BOT_CONFIG}/config \
                     --aws-region=us-east-1""")
 
+        if (params.ARCH == "x86_64") {
+            stage('Xen') {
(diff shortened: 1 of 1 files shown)


jobs/kola-aws.Jenkinsfile
```diff
@@ -73,6 +73,23 @@ try { timeout(time: 90, unit: 'MINUTES') {
...
+            stage('Xen') {
+                def tests = params.KOLA_TESTS
+                if (tests == "") {
+                    tests = "basic"
+                }
+                fcosKola(cosaDir: env.WORKSPACE,
+                         build: params.VERSION, arch: params.ARCH,
+                         extraArgs: tests,
+                         skipUpgrade: true,
+                         skipBasicScenarios: true,
```
dustymabe, line 86:
```suggestion
                         skipBasicScenarios: false,
```

We just want it to run basic scenarios, right? Is there a way through coreos-ci-lib to get it to just run basic scenarios?

PR #419, opened 2021-11-01
jobs: add `LEAK_ON_FAIL` parameter to kola jobs

jlebon:
That way, we can run jobs targeting failing tests which we can then
manually debug into when it does fail. Likely we'll want a
"multiply"-like argument with this too for flaky tests.

For more information, see
https://github.com/coreos/coreos-assembler/pull/2511

### DIFF
--- a/jobs/kola-aws.Jenkinsfile
+++ b/jobs/kola-aws.Jenkinsfile
@@ -1,9 +1,10 @@
 @Library('github.com/coreos/coreos-ci-lib@main') _
 
-def streams
+def streams, pipeutils
 node {
     checkout scm
     streams = load("streams.groovy")
+    pipeutils = load("utils.groovy")
 }
 
 properties([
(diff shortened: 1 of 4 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/2511
Requires: https://github.com/coreos/coreos-ci-lib/pull/93

jlebon:
(I haven't added the key to OpenShift yet.)

jlebon:
Will rework this, I think it should instead be a higher-level `DEBUG_TEST_FAILURE` checkbox.

cgwalters (commented):
Superficial LGTM.

I think we need to make the pipeline involve less "parameter threading" for changes like this.  More  "convention not configuration" probably and less Jenkins parameters.  I think Gangplank was trying to go this route, deriving from the original rhcos pipeline "jobspec".

PR #420, opened 2021-11-02
pipeline: add some debugging info re. umask and permissions

jlebon:
To help debug
https://github.com/coreos/fedora-coreos-tracker/issues/1003.


jlebon:
(We can revert this once we're done with it.)

dustymabe (approved):
LGTM

PR #426, opened 2021-11-15
jobs/kola-openstack: allow concurrent builds

jmarrero:


jobs/kola-openstack.Jenkinsfile
```diff
@@ -43,97 +41,100 @@ properties([
     durabilityHint('PERFORMANCE_OPTIMIZED')
 ])
 
-currentBuild.description = "[${params.STREAM}][${params.ARCH}] - ${params.VERSION}"
+lock(resource: "build-${params.STREAM}-${params.ARCH}") {
```
dustymabe, line 44: Let's adjust the lock name slightly.

- Switch to denote which job we are locking (`kola-openstack`).
- We want a single run on each architecture at any given time. Then the code will switch the region based on the architecture.

```suggestion
lock(resource: "kola-openstack-${params.ARCH}") {
```

dustymabe:
Looks like we need a second commit to say something along the lines of:

```
if arch == x86_64; then
    region = ams1
else
    region = ca-ymq-1
```

Then we can update the calls to ore and kola to specify the region.

jobs/kola-openstack.Jenkinsfile
```diff
@@ -43,97 +41,105 @@ properties([
...
 
-currentBuild.description = "[${params.STREAM}][${params.ARCH}] - ${params.VERSION}"
+lock(resource: "kola-openstack-${params.ARCH}") {
 
-def s3_stream_dir = params.S3_STREAM_DIR
-if (s3_stream_dir == "") {
-    s3_stream_dir = "fcos-builds/prod/streams/${params.STREAM}"
-}
+    currentBuild.description = "[${params.STREAM}][${params.ARCH}] - ${params.VERSION}"
+    region = "ams1"
```
dustymabe, line 47: Might need `def` in the front like is done for other variables.

```suggestion
    def region = "ams1"
```

jobs/kola-openstack.Jenkinsfile
```diff
@@ -43,97 +41,105 @@ properties([
...
+    if (params.ARCH != "x86_64") {
+        region = "ca-ymq-1"
+    }
```
dustymabe, line 50: Here it might be better to have a full if/elseif/else for each arch just so your brain doesn't have to thing too much when reading the code. Can still have a default region at the top (in case a new architecture ever gets thrown at us).

jobs/kola-openstack.Jenkinsfile
```diff
@@ -43,97 +41,105 @@ properties([
...
+                // Create the image in OpenStack
                 shwrap("""
-                ore openstack --config-file=\${OPENSTACK_KOLA_TESTS_CONFIG}/config \
-                     delete-image --id=${openstack_image_name}
+                # First delete it if it currently exists, then create it.
+                ore openstack --region=${region} --config-file=\${OPENSTACK_KOLA_TESTS_CONFIG}/config \
+                    delete-image --id=${openstack_image_name} || true
+                ore openstack --region=${region} --config-file=\${OPENSTACK_KOLA_TESTS_CONFIG}/config \
+                    create-image --file=${openstack_image_filepath} \
+                    --name=${openstack_image_name} --arch=${params.ARCH}
```
dustymabe, line 101: Maybe let's add an extra line:

```suggestion
                ore openstack --region=${region} \
                    --config-file=\${OPENSTACK_KOLA_TESTS_CONFIG}/config \
                    delete-image --id=${openstack_image_name} || true
                ore openstack --region=${region} \
                    --config-file=\${OPENSTACK_KOLA_TESTS_CONFIG}/config \
                    create-image --file=${openstack_image_filepath} \
                    --name=${openstack_image_name} --arch=${params.ARCH}
```

jobs/kola-openstack.Jenkinsfile
```diff
@@ -43,97 +41,105 @@ properties([
...
+                            --openstack-region=${region}                             \
+                            --openstack-floating-ip-network=public                   \
+                            --openstack-image=${openstack_image_name}""")
+            } finally {
+                stage('Delete Image') {
+                    // Delete the image in OpenStack
+                    shwrap("""
+                    ore openstack --region=${region} --config-file=\${OPENSTACK_KOLA_TESTS_CONFIG}/config \
+                        delete-image --id=${openstack_image_name}
+                    """)
```
dustymabe, line 131: same as above
```suggestion
                    ore openstack --region=${region} \
                        --config-file=\${OPENSTACK_KOLA_TESTS_CONFIG}/config \
                        delete-image --id=${openstack_image_name}
                    """)
```

dustymabe (commented):
Looking good. Some suggestions:

jobs/kola-openstack.Jenkinsfile
dustymabe: Might also be good to have a comment above this block that explains **why** we want to send different architectures to different regions. Basically that we don't have enough quota in a single region to do multiple runs at the same time but we want to speed up our testing so we'll target two regions (i.e. we can run 2 runs at the same time (an `aarch64` and `x86_64`)).

dustymabe: Maybe that comment should go above the `lock()` block.

jobs/kola-openstack.Jenkinsfile
```diff
@@ -43,97 +41,114 @@ properties([
     durabilityHint('PERFORMANCE_OPTIMIZED')
 ])
 
-currentBuild.description = "[${params.STREAM}][${params.ARCH}] - ${params.VERSION}"
+// Locking so we don't run multiple times on the same region we are speeding up our testing by dividing
+// the load to different regions depending on the architecture. This allow us not to exceed our quota in 
+// a single region and not having to wait for sequential runs on the same region.
```
dustymabe, line 46:
```suggestion
// Locking so we don't run multiple times on the same region. We are speeding up our testing by dividing
// the load to different regions depending on the architecture. This allows us not to exceed our quota in 
// a single region while still being able to execute two test runs in parallel.
```

dustymabe (approved):
LGTM - one suggestion. Once we merge this we'll fixup any issues that arise after the next pipeline run.

dustymabe:
Nice work @jmarrero!

Issue #423, opened 2021-11-09
Support specifying a custom cosa image for multi-arch pipeline


jlebon:
We'll need this for the next stable release so that we build with the same pre-f35 cosa we used for testing.

PR #427, opened 2021-11-16
multi-arch-pipeline: derive the COSA image from params.COREOS_ASSEMBLER_IMAGE

dustymabe:
We currently have a limitation where we aren't building and
pushing multi-arch COSA containers to quay. For multi-arch
we're currently building images once a day on the local
multi-arch builders. See https://github.com/coreos/coreos-assembler/issues/2470

Until `#2470` is resolved let's do the best thing we can do
which is derive the multi-arch container name from the
given x86_64 COSA container. We'll translate
`quay.io/coreos-assembler/coreos-assembler:$tag -> localhost/coreos-assembler:$tag`
This assumes that the desired tagged image has been built
on the multi-arch builder already, which most likely means
someone did it manually.

Fixes https://github.com/coreos/fedora-coreos-pipeline/issues/423

### DIFF
--- a/jobs/multi-arch-pipeline.Jenkinsfile
+++ b/jobs/multi-arch-pipeline.Jenkinsfile
@@ -145,6 +145,23 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}") {
         def newBuildID = params.VERSION
         def basearch = params.ARCH
 
+        // We currently have a limitation where we aren't building and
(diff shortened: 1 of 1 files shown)


dustymabe:
untested at the moment. I'm thinking rather than set up a pipeline we roll with it and fix any issues in a followup.

jlebon (approved):
Nice comment block!

PR #430, opened 2021-11-19
plugins: configuration-as-code and pipeline-utility-steps

jmarrero:
Update configuration-as-code to 1.54 as it prevented the
Credentials Plugin as others from loading.

Adding pipeline-utility-steps 2.9.0 as is required for the current
Jenkins version to run the pipelines.

Add the coreos-ci-lib global library to the configmap.

### DIFF
--- a/jenkins/config/coreos-ci-lib.yaml
+++ b/jenkins/config/coreos-ci-lib.yaml
@@ -1,0 +1,11 @@
+unclassified:
+  globalLibraries:
+    libraries:
+      - name: "coreos-ci-lib"
+        defaultVersion: main
+        implicit: true
+        retriever:
+          modernSCM:
+            scm:
+              git:
+                remote: "https://github.com/coreos/coreos-ci-lib.git"
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,3 @@
-@Library('github.com/coreos/coreos-ci-lib@main') _
-
 import org.yaml.snakeyaml.Yaml;
 
 def pipeutils, streams, official, official_jenkins, developer_prefix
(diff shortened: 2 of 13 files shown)


dustymabe (approved):
trivial LGTM - will let jlebon have the final looksie

jenkins/config/coreos-ci-lib.yaml
```diff
@@ -0,0 +1,11 @@
+unclassified:
+  globalLibraries:
+    libraries:
+      - name: "github.com/coreos/coreos-ci-lib"
```
jlebon, line 4: Shouldn't this just be e.g. `coreos-ci-lib`?

jlebon (approved):
One comment, but LGTM overall. Thanks!

jenkins/config/coreos-ci-lib.yaml
jmarrero: The error I got said: `ERROR: Could not find any definition of libraries [github.com/coreos/coreos-ci-lib@main]` changing this to `github.com/coreos/coreos-ci-lib` made the pipeline move. I can try with `coreos-ci-lib` and report back what happens.

jmarrero: tested with coreos-ci-lib and it also works. Updated

jlebon (approved):

jlebon:
@jmarrero Mind recreating the configmap and killing the Jenkins pod in the prod environment? Otherwise I think jobs referencing coreos-ci-lib helpers will start failing.

jmarrero:
@jlebon sure thing

PR #432, opened 2021-11-22
manifest/pipeline: set cosa referencePolicy to `local`

jlebon:
That way we use the latest pulled cosa image in the internal registry,
which shouldn't be very out of date. This also avoids race conditions
during quay.io rebuilds.


dustymabe (approved):
LGTM

PR #436, opened 2021-11-23
Jenkinsfile: add rerun flag

saqibali-2k:
Reruns have been disabled by default. Lets add the flag to turn
them on in the pipeline.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -320,7 +320,7 @@ lock(resource: "build-${params.STREAM}") {
 
         stage('Kola:QEMU basic') {
             shwrap("""
-            cosa kola run --basic-qemu-scenarios --no-test-exit-error
+            cosa kola run --rerun --basic-qemu-scenarios --no-test-exit-error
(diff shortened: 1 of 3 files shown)


dustymabe:
There are a few places in this repo where we call [`fcosKola()`](https://github.com/coreos/coreos-ci-lib/blob/main/vars/fcosKola.groovy) from `coreos-ci-lib`. I'm not sure whether we should update the lib or update our calls here to make sure `--rerun` gets added.

jlebon (approved):
LGTM! As @dustymabe mentioned, there is also `fcosKola`.

> I'm not sure whether we should update the lib or update our calls here to make sure `--rerun` gets added.

I think the behaviour makes sense as a default in all CI so I'd for updating the library. Ideally there's a way to opt-out though (`reruns: false`?).

dustymabe:
ok yeah let's merge this and update the library. @saqibali-2k can you look into that?

saqibali-2k:
> ok yeah let's merge this and update the library. @saqibali-2k can you look into that?

Can do!

PR #437, opened 2021-11-23
multi-arch-pipeline: emit build.state.change fedmsg on start

jlebon:
We're emitting one on finish but were missing the start one for parity
with x86_64.


dustymabe (approved):

dustymabe:
Note that sometimes the multi-arch-pipeline can fail and we restart it with the same build ID. Would it matter if multiple messages with the same build ID went out?

jlebon:
> Note that sometimes the multi-arch-pipeline can fail and we restart it with the same build ID. Would it matter if multiple messages with the same build ID went out?

Yeah, it should be fine for `STARTED` events. I think that can happen with the main pipeline too if it fails early enough signing OSTree or building QEMU.

PR #442, opened 2021-12-03
Add kola-kubernetes job

jlebon:
This job will run the new Kubernetes e2e node tests in AWS:
https://github.com/coreos/fedora-coreos-config/pull/1297

Like the other tests, we trigger it as a followup job from the pipeline
jobs.

### DIFF
--- a/jobs/kola-kubernetes.Jenkinsfile
+++ b/jobs/kola-kubernetes.Jenkinsfile
@@ -1,0 +1,82 @@
+def streams
+node {
+    checkout scm
+    streams = load("streams.groovy")
+}
+
+properties([
+    pipelineTriggers([]),
+    parameters([
+      choice(name: 'STREAM',
+             // list devel first so that it's the default choice
+             choices: (streams.development + streams.production + streams.mechanical),
+             description: 'Fedora CoreOS stream to test'),
+      string(name: 'VERSION',
+             description: 'Fedora CoreOS Build ID to test',
+             defaultValue: '',
+             trim: true),
+      string(name: 'ARCH',
+             description: 'Target architecture',
+             defaultValue: 'x86_64',
+             trim: true),
+      string(name: 'S3_STREAM_DIR',
+             description: 'Override the Fedora CoreOS S3 stream directory',
+             defaultValue: '',
+             trim: true),
+      string(name: 'COREOS_ASSEMBLER_IMAGE',
+             description: 'Override the coreos-assembler image to use',
+             defaultValue: "coreos-assembler:main",
+             trim: true),
+      string(name: 'FCOS_CONFIG_COMMIT',
+             description: 'The exact config repo git commit to run tests against',
+             defaultValue: '',
+             trim: true),
+    ]),
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+])
+
+currentBuild.description = "[${params.STREAM}][${params.ARCH}] - ${params.VERSION}"
+
+def s3_stream_dir = params.S3_STREAM_DIR
+if (s3_stream_dir == "") {
+    s3_stream_dir = "fcos-builds/prod/streams/${params.STREAM}"
+}
+
+try { timeout(time: 60, unit: 'MINUTES') {
+    cosaPod(image: params.COREOS_ASSEMBLER_IMAGE,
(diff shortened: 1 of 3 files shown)


jobs/kola-kubernetes.Jenkinsfile
```diff
@@ -0,0 +1,80 @@
...
+        stage('Fetch Metadata') {
+            def commitopt = ''
+            if (params.FCOS_CONFIG_COMMIT != '') {
+                commitopt = "--commit=${params.FCOS_CONFIG_COMMIT}"
+            }
+            shwrap("""
+            export AWS_CONFIG_FILE=\${AWS_FCOS_BUILDS_BOT_CONFIG}/config
+            cosa init --branch ${params.STREAM} ${commitopt} https://github.com/coreos/fedora-coreos-config
+            cosa buildfetch --artifact=ostree --artifact=qemu --build=${params.VERSION} \
+                --arch=${params.ARCH} --url=s3://${s3_stream_dir}/builds
```
dustymabe, line 59: If we're launching in AWS do we need to buildfetch `--artifact=ostree --artifact=qemu` ?

jobs/multi-arch-pipeline.Jenkinsfile
```diff
@@ -480,6 +480,16 @@ EOF
...
+            stage('Kola:Kubernetes') {
+                // We consider the Kubernetes kola tests to be a followup job, so we use `wait: false` here.
+                build job: 'kola-kubernetes', wait: false, parameters: [
+                    string(name: 'STREAM', value: params.STREAM),
+                    string(name: 'VERSION', value: newBuildID),
+                    string(name: 'S3_STREAM_DIR', value: s3_stream_dir),
+                    string(name: 'ARCH', value: basearch),
+                    string(name: 'FCOS_CONFIG_COMMIT', value: params.FCOS_CONFIG_COMMIT)
+                ]
+            }
```
dustymabe, line 492: oh nice.. Does the test work against `aarch64` ?

jobs/kola-kubernetes.Jenkinsfile
```diff
@@ -0,0 +1,80 @@
...
+            """)
+        }
+
+        fcosKola(cosaDir: env.WORKSPACE,
+                 build: params.VERSION, arch: params.ARCH,
+                 extraArgs: "--tag k8s",
+                 skipBasicScenarios: true,
+                 platformArgs: """-p=aws \
```
dustymabe, line 67: might be worth adding a comment here about why we chose AWS. I assume it's because it's the only platform we have images pre-uploaded to (AWS and GCP right now) that also has `x86_64` **AND** `aarch64`.

dustymabe (commented):
mostly LGTM - some suggestions

jobs/kola-kubernetes.Jenkinsfile
jlebon: Ahh whoops, this is from what I initially was going to run this on QEMU.

jobs/multi-arch-pipeline.Jenkinsfile
jlebon: I'm not sure, but I figured we might as well find out and follow-up if it doesn't. The cri-o installer script at least accounts for it: https://github.com/cri-o/cri-o/blob/367232bb1bc7ab3a3bd5a863daf701b036966473/scripts/get#L43-L54

jobs/kola-kubernetes.Jenkinsfile
jlebon: Added comment.

jobs/kola-kubernetes.Jenkinsfile
dustymabe: do we even need `--artifact=ostree` ? I think we download that in the AWS/GCP tests because we do upgrade testing. We're not doing that here IIUC.

dustymabe (approved):
LGTM

One final open question about if `buildfetch --artifact=ostree` is needed or not.

jobs/kola-kubernetes.Jenkinsfile
jlebon: Ahh good catch, indeed.

PR #450, opened 2021-12-21
kola-aws: add test on m6i.large instances

dustymabe:
We saw a regression recently where m6i instances failed to boot on
newer kernels. Add a test here to make sure that continue to boot in
the future.

See https://github.com/coreos/fedora-coreos-tracker/issues/1004


aaradhak (approved):

jmarrero (approved):
lgtm

PR #451, opened 2022-01-05
Add and use centralized metadata for recording whether `next-devel` is in use

bgilbert:
The process for enabling/disabling `next-devel` requires multiple PRs, and there's no dashboard indicating its current state.  Add a centralized metadata source that other infrastructure can use to display or act on the current `next-devel` branch state, and use it to enable/disable that stream in the pipeline.

Previous discussion in https://github.com/coreos/fedora-coreos-tracker/pull/1052.


cgwalters (approved):

streams.groovy
```diff
@@ -1,7 +1,9 @@
 // Canonical definition of all our streams and their type.
 
+next_devel = load("next-devel/status.groovy")
```
jlebon, line 3: This is assuming that the repo is checked out exactly at the working directory. This is true for all loads of `streams.groovy` right now, but it'd be easy to regress on. A simpler Groovy approach (but trickier implementation-wise) is to have `manage.py` edit `streams.groovy` directly, but definitely good to try this!

jlebon (approved):
Thanks!

streams.groovy
bgilbert: All of the loads of `streams.groovy` also assume that, so I think the only risk is from external code using it as a library?

It seemed cleaner to keep the generated code separate, but I'm open to direct editing if we want/need.  Will merge as-is for now.

bgilbert: Turns out the `load()` doesn't work.  Followup in https://github.com/coreos/fedora-coreos-pipeline/pull/453.

PR #453, opened 2022-01-10
next-devel: modify `streams.groovy` directly

bgilbert:
Jenkins doesn't allow `load()` at file scope inside another `load()`.  Drop the nested inclusion and have `manage.py` edit `streams.groovy` directly.


jlebon (approved):
Nice! Thanks for the quick fix.

PR #454, opened 2022-01-13
streams: re-enable next-devel

jlebon:
We want to ship microdnf there only, and we'd also like to roll out the
iptables-nft migration there first.

See: https://github.com/coreos/fedora-coreos-tracker/issues/676
See: https://github.com/coreos/fedora-coreos-tracker/issues/1050

### DIFF
--- a/next-devel/badge.json
+++ b/next-devel/badge.json
@@ -2,6 +2,6 @@
     "schemaVersion": 1,
     "style": "for-the-badge",
     "label": "next-devel",
-    "message": "closed",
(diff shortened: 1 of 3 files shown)


next-devel/badge.json
```diff
@@ -2,6 +2,6 @@
     "schemaVersion": 1,
     "style": "for-the-badge",
     "label": "next-devel",
-    "message": "closed",
-    "color": "lightgrey"
```
dustymabe, line 6: where is this badge used?

Issue #455, opened 2022-01-17
Increase session timeout for web UI


bgilbert:
The login timeout seems awfully short, on the order of a couple hours?  I sometimes find myself having to log in more than twice per day.

PR #456, opened 2022-01-17
manifests/jenkins: bump session timeout

jlebon:
The default is 30 minutes, which is short enough that you sometimes have
to log back in multiple times a day. Let's just bump it to one day.

Closes: #455

### DIFF
--- a/manifests/jenkins.yaml
+++ b/manifests/jenkins.yaml
@@ -88,6 +88,12 @@ objects:
             value: >-
               -Dorg.jenkinsci.plugins.durabletask.BourneShellScript.HEARTBEAT_CHECK_INTERVAL=900
               -Dorg.jenkinsci.plugins.durabletask.BourneShellScript.LAUNCH_DIAGNOSTICS=true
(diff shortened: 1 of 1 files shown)


jlebon:
`JENKINS_OPTS` is documented [here](https://docs.openshift.com/container-platform/4.9/openshift_images/using_images/images-other-jenkins.html) (though the way I found it was by looking at https://github.com/openshift/jenkins/blob/bc2b875e82d7d09afad3326ca015719719935b0f/2/contrib/s2i/run#L658).

bgilbert (approved):
LGTM

dustymabe:
Thank you!

jlebon:
Merged, though won't redeploy Jenkins just yet. I'm sure we'll have other reasons to redeploy at some point and then this will get in.

dustymabe:
> Merged, though won't redeploy Jenkins just yet. I'm sure we'll have other reasons to redeploy at some point and then this will get in.

If it's easy to do I'd just say go ahead when things are looking idle. Otherwise it could lead to confusion along the lines of "I thought we fixed that?".

PR #457, opened 2022-01-19
release: Copy oscontainer to quay.io

cheesesashimi:
This PR is intended to supercede and incorporates elements of: https://github.com/coreos/fedora-coreos-pipeline/pull/383

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -218,6 +218,13 @@ uuidgen -r > secret
 oc create secret generic github-webhook-shared-secret --from-file=secret
 ```
 
+### [PROD] Create quay.io image push secret
+
+This secret is used to push the resulting OCI image to Quay.io
+
+1. Obtain the file `oscontainer-secret` from BitWarden.
+2. Run: `$ oc create secret generic oscontainer-secret --from-file=dockercfg=oscontainer-secret`.
+
 ### Create a Jenkins instance with a persistent volume backing store
 
 ```
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -60,6 +60,10 @@ echo "Final podspec: ${pod}"
 // use a unique label to force Kubernetes to provision a separate pod per run
 def pod_label = "cosa-${UUID.randomUUID().toString()}"
 
+// Destination for OCI image push
+// TODO: Change this to quay.io/fedora/coreos per https://fedoraproject.org/wiki/Changes/OstreeNativeContainer
+def quay_registry = "quay.io/coreos-assembler/fcos"
+
 // We just lock here out of an abundance of caution in case somehow two release
 // jobs run for the same stream, but that really shouldn't happen. Anyway, if it
 // *does*, this makes sure they're run serially.
@@ -83,9 +87,17 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
             shwrap("""
             export AWS_CONFIG_FILE=\${AWS_FCOS_BUILDS_BOT_CONFIG}
             cosa init --branch ${params.STREAM} https://github.com/coreos/fedora-coreos-config
-            cosa buildfetch --build=${params.VERSION} \
+            cosa buildfetch --artifact=ostree --build=${params.VERSION} \
                 --arch=all --url=s3://${s3_stream_dir}/builds
             """)
+        }
+
+        stage('Sync oscontainer to quay.io') {
+            withCredentials([file(credentialsId: 'oscontainer-secret', variable: 'OSCONTAINER_SECRET')]) {
+                shwrap("""
+                skopeo copy --authfile="${OSCONTAINER_SECRET}" "oci-archive://\$(cosa meta --image-path ostree)" "docker://${quay_registry}:${params.STREAM}"
+                """)
+            }
         }
 
         for (basearch in params.ARCHES.split()) {
--- a/jenkins/config/oscontainer-secret.yaml
+++ b/jenkins/config/oscontainer-secret.yaml
@@ -1,0 +1,13 @@
+credentials:
+  system:
+    domainCredentials:
+      - credentials:
+        - file:
+            scope: GLOBAL
+            fileName: oscontainer-secret
+            id: oscontainer-secret
+            # Secret must be base64-encoded
+            # See: https://github.com/jenkinsci/configuration-as-code-plugin/issues/884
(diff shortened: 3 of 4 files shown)


manifests/pod.yaml
```diff
@@ -54,6 +54,9 @@ spec:
      - name: fedora-messaging-coreos-key
        mountPath: /run/kubernetes/secrets/fedora-messaging-coreos-key
        readOnly: true
+     - name: oscontainer-registry
```
jlebon, line 57: The pod shouldn't need the secret mounted in since we're using `withCredentials`.

Jenkinsfile.release
```diff
@@ -85,9 +85,21 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
             cosa init --branch ${params.STREAM} https://github.com/coreos/fedora-coreos-config
             cosa buildfetch --build=${params.VERSION} \
                 --arch=all --url=s3://${s3_stream_dir}/builds
+            cosa buildprep --ostree --build=${params.VERSION} \
```
jlebon, line 88: You can add `--artifact ostree` to the previous `buildfetch` call instead.

Jenkinsfile.release
```diff
@@ -85,9 +85,21 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
...
+                --arch=all s3://${s3_stream_dir}/builds
             """)
         }
 
+        stage('Sync oscontainer to quay.io') {
+            withCredentials([string(credentialsId: "oscontainer-secret", variable: "OSCONTAINER_SECRET")]) {
+                shwrap("""ociarchive=\$(cosa meta --image-path ostree)
+                        case \${ociarchive} in
+                        *ociarchive) skopeo copy --authfile=<($OSCONTAINER_SECRET) oci-archive://\${ociarchive} docker://quay.io/coreos-assembler/fcos:${params.STREAM};;
```
jlebon, line 97: Might be good to const-ify the quay.io URL here at the beginning of the file for clarity. And also a comment there that says long-term we want `quay.io/fedora/coreos` as per https://fedoraproject.org/wiki/Changes/OstreeNativeContainer.

Jenkinsfile.release
```diff
@@ -85,9 +85,21 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
...
                 --arch=all --url=s3://${s3_stream_dir}/builds
+            cosa buildprep --ostree --build=${params.VERSION} \
+                --arch=all s3://${s3_stream_dir}/builds
             """)
         }
 
+        stage('Sync oscontainer to quay.io') {
+            withCredentials([string(credentialsId: "oscontainer-secret", variable: "OSCONTAINER_SECRET")]) {
+                shwrap("""ociarchive=\$(cosa meta --image-path ostree)
+                        case \${ociarchive} in
```
jlebon, line 96: All streams use ociarchive now and at some point we'll probably rip out the old support for .tar in cosa. So I think we can just simplify this logic here.

jenkins/config/oscontainer-secret.yaml
```diff
@@ -0,0 +1,9 @@
+credentials:
+  system:
+    domainCredentials:
+      - credentials:
+        - string:
```
jlebon, line 5: Cool with this, though feels like a secret file would be more appropriate for this instead.

jlebon: See https://jenkins-fedora-coreos.apps.ocp.ci.centos.org/configuration-as-code/reference#Credentials-file.

cheesesashimi: I was originally intending to do that, I just couldn't find the syntax for some reason. Thanks for the pointer!

cheesesashimi: It took a bit of trial-and-error. The short version is that you have to base64-encode the file contents to make it work.

Jenkinsfile.release
cheesesashimi: I'm not sure that works:

```console
$ cosa buildfetch --ostree
usage: cmd-buildfetch [-h] [--url URL] [--stream STREAM] [-b BUILD] [--force] [--arch ARCH] [--artifact ARTIFACT]
cmd-buildfetch: error: unrecognized arguments: --ostree
```

jlebon (approved):
Thanks, looks great!

Final bits:
- Can you squash your two commits into one?
- Can you add a section in HACKING.md about the new secret similar to the other ones?

cheesesashimi:
Done!

HACKING.md
```diff
@@ -218,6 +218,13 @@ uuidgen -r > secret
 oc create secret generic github-webhook-shared-secret --from-file=secret
 ```
 
+### [PROD] Create quay.io image push secret
+
+This secret is used to push the resulting OCI image to Quay.io
+
+1. Obtain the file `oscontainer-secret` from BitWarden.
+2. Run: `$ oc create secret generic oscontainer-secret --from-file=dockercfg=oscontainer-secret`.
```
jlebon, line 226: OK, *really* final bits so we don't break the pipeline as soon as this merges:
1. Did you add the secret to BitWarden? I don't see it.
2. Did you already create the OpenShift secret? I don't see it either.

jlebon:
This is ready for merge. Waiting until our current test runs in the new infra are over before I get it in and redeploy Jenkins.

HACKING.md
jlebon: Secret was added to BitWarden and OpenShift (both old and new cluster in case we rollback the migration temporarily).

jlebon:
I totally missed this until now but: when I said to squash your commits I meant have a single commit on top of Colin's commit. Authorship recognition is important and you did some nice cleanups and testing for this. Sorry for not noticing earlier!

cgwalters:
In this case the [Co-authored-by tag](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) seems best but for the record I don't think this is a big deal from my PoV - just something to keep in mind elsewhere!

I particularly try to preserve credit in the case of a new contributor to a project submitting a PR, and then I need to changes on top.  But that's not this case really.

Anyways - cool to see this landed!

cheesesashimi:
No worries! I thought about it almost immediately after I squashed it.

PR #460, opened 2022-01-26
release: don't use Groovy string interpolation for OSCONTAINER_SECRET

jlebon:
Jenkins blocks it by default and recommends using single quotes instead
so that the environment variable is substituted by the shell itself.


cheesesashimi (approved):
👍

PR #461, opened 2022-01-26
jenkins/config: fix oscontainer-secret formatting

jlebon:
We want to base64-encode the contents of a file, not the path of the
file itself.

PR #462, opened 2022-01-26
Revert "jenkins/config: fix oscontainer-secret formatting"

jlebon:
This reverts commit db97d5674fd82fda76826355a1be9b0a351c2173.

Lessons learned:
- `${oscontainer-secret/dockercfg}` expands to the *contents* of the
  path at `/run/secrets/oscontainer-secret/dockercfg`
- therefore, `${readFile:${oscontainer-secret/dockercfg}}` will ask
  jcasc to read the file at `$CONTENTS`, which is of course not what we
  want.

So the only issue was actually the path typo, which was fixed in #461.
As such, this is not a perfect revert. I also kept the updated
documentation link and added some details to the comment.

### DIFF
--- a/jenkins/config/oscontainer-secret.yaml
+++ b/jenkins/config/oscontainer-secret.yaml
@@ -6,7 +6,9 @@ credentials:
             scope: GLOBAL
             fileName: oscontainer-secret
             id: oscontainer-secret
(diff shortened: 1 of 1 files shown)


jlebon:
No rush on this; I've already applied it to the cluster to verify that it worked (and it did!).

dustymabe (approved):
LGTM

PR #465, opened 2022-02-03
Jenkinsfile.release: pass --build to cosa meta

dustymabe:
We hit this issue where one pipeline run kicked off immediately after
another one had finished. So we had something like:

- next-devel-x86_64#1
- next-devel-aarch64#1
- next-devel-x86_64-release#1
- next-devel-x86_64#2
- next-devel-aarch64-release#1

At the point that the release job runs for `#1` after the aarch64
pipeline is done the builds.json has already been updated with build
ID from the `#2` run. In this case `cosa meta` will look at the meta.json
from the wrong build.

To work around this we need to just be explicit and pass the build
ID directly to `cosa meta`.

### DIFF
--- a/Jenkinsfile.release
+++ b/Jenkinsfile.release
@@ -93,9 +93,11 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
         }
 
         stage('Sync oscontainer to quay.io') {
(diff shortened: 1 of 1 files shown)


jlebon (approved):
Nice catch!

PR #466, opened 2022-02-04
jobs: Add buildDiscarder property to the build pipelines

jmarrero:
Adds the buildDiscarder property to the the jobs to prevent the builds to store unlimited logs and artifacts.


dustymabe (approved):
LGTM

PR #472, opened 2022-02-08
plugins: bump workflow-api to latest

jlebon:
Needed by latest `kubernetes-plugin`.

We should update every other plugin on there, but let's do that
separately when we know we're in a good place.

### DIFF
--- a/jenkins/controller/plugins.txt
+++ b/jenkins/controller/plugins.txt
@@ -9,7 +9,7 @@ configuration-as-code-groovy:1.1
 timestamper:1.11.8
 build-token-root:1.7
 github:1.33.1
-workflow-api:2.46
+workflow-api:1136.v7f5f1759dc16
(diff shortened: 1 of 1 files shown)


jenkins/controller/plugins.txt
```diff
@@ -9,7 +9,7 @@ configuration-as-code-groovy:1.1
 timestamper:1.11.8
 build-token-root:1.7
 github:1.33.1
-workflow-api:2.46
```
jlebon, line 12: Releases changed recently to be GitHub Actions triggered: https://github.com/jenkinsci/workflow-api-plugin/releases. See also https://github.com/jenkinsci/workflow-api-plugin/pull/183.

dustymabe (approved):
LGTM

bgilbert (approved):
LGTM

PR #474, opened 2022-02-09
jobs/release: make AWS_REPLICATION a bool

jlebon:
The jobs which trigger this job were changed to pass in a boolean, but
the parameter on this job itself was still a string choice. Adapt.

Fixes 4058b24 ("Jenkinsfile.release: move to regular job `release`").


dustymabe (approved):
LGTM

PR #479, opened 2022-02-22
jobs/build: add temporary KOLA_RUN_SLEEP parameter

jlebon:
We're trying to debug
https://github.com/coreos/fedora-coreos-tracker/issues/1108, but it's
not reproducing locally, nor in CoreOS CI.

Let's add a quick and dirty checkbox that we can activate to have the
pipeline just hold at the place where it would run kola so we can `oc
exec` in and do it ourselves to try to reproduce it in more similar
conditions.


dustymabe (approved):

dustymabe:
LGTM

PR #481, opened 2022-03-08
bump-lockfile: rebase before pushing to git repo

dustymabe:
We often hit an issue where a new upstream commit has landed and
the git push will fail. It could be the case that the new upstream
commit causes an issue and it would be more correct for bump-lockfile
to just fail (as it currently does), but I think practically it's not
often that the new commits coming in would conflict with the changes
tested as part of the bump-lockfile run. Even if it did, the next
`testing-devel` pipeline run would catch it. Let's just rebase and
push.

### DIFF
--- a/jobs/bump-lockfile.Jenkinsfile
+++ b/jobs/bump-lockfile.Jenkinsfile
@@ -206,7 +206,8 @@ EOF
         withCredentials([usernamePassword(credentialsId: botCreds,
                                           usernameVariable: 'GHUSER',
                                           passwordVariable: 'GHTOKEN')]) {
-          // should gracefully handle race conditions here
+          // Sometimes other commits have come in; rebase if needed and then push.
+          sh("git -C src/config pull --rebase")
           sh("git -C src/config push https://\${GHUSER}:\${GHTOKEN}@github.com/${repo} ${branch}")
         }
     }


dustymabe:
untested - this is meant to spark discussion

cgwalters (approved):
>  I think practically it's not often that the new commits coming in would conflict with the changes tested as part of the bump-lockfile run.

My intuition agrees with you.  But, I don't think we'll have any data here until we actually *try* right?

(Alternatively, while it wouldn't be cheap I could imagine that we change this flow to retest, and then we would better know pass/fail in this scenario)

But yeah...probably not worth it.

dustymabe:
> (Alternatively, while it wouldn't be cheap I could imagine that we change this flow to retest, and then we would better know pass/fail in this scenario)

yeah. I was originally going to do this ^^, but was thinking to myself that I think it may be better just to short circuit.

dustymabe:
I think the only problem with the re-trigger approach is if you happened to be waiting on a bump-lockfile run to finish (now you have to wait two hours), but like I said I was originally going to do something similar to what you did.

If you want to open a PR with that I'll be glad to review :)

jlebon:
Threw it up in https://github.com/coreos/fedora-coreos-pipeline/pull/482!

PR #482, opened 2022-03-08
jobs/bump-lockfile: self-retrigger on races

jlebon:
Testing a lockfile bump takes time. By the time we're ready to push out
an update, a new commit might've been pushed upstream which we haven't
tested. Detect that case and instead of erroring out, retrigger
ourselves.

See also: https://github.com/coreos/fedora-coreos-pipeline/pull/481


dustymabe (approved):
LGTM

jlebon:
Let's try it! (I agree the wasted cycles are unfortunate; hmm, I think there's room to do something more sophisticated here down the line.)

PR #483, opened 2022-03-09
Move from Quay.io to app.ci

jlebon:
The latter mirrors to the former, but it's a periodic that runs every
hour. So let's just use the images directly from app.ci for faster
iteration.

Related: https://github.com/coreos/coreos-ci-lib/pull/107

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -161,6 +161,12 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}") {
                 "localhost/coreos-assembler:"
             )
         }
+        if (params.COREOS_ASSEMBLER_IMAGE.startsWith("registry.ci.openshift.org/coreos/coreos-assembler:")) {
+            image = params.COREOS_ASSEMBLER_IMAGE.replaceAll(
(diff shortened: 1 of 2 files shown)


bgilbert (approved):
LGTM

dustymabe:
I guess I'm OK with this, but to me it makes things slightly less transparent. Everyone doing development or testing is pulling from quay. If I start cross checking image digests and trying to figure out when something failed I'm looking at quay.

jlebon:
> I guess I'm OK with this, but to me it makes things slightly less transparent. Everyone doing development or testing is pulling from quay.

Note that [CoreOS CI is also using app.ci](https://github.com/coreos/coreos-ci-lib/pull/107). The reason is that it's a pain to have to wait for a rebuild *and* up to an hour after fixing something in cosa that will unblock CI. I think that reason is valid for the pipeline too, but mainly it's important that we match CoreOS CI.

> If I start cross checking image digests and trying to figure out when something failed I'm looking at quay.

We can also look at [the cosa info that the pipeline prints out](https://github.com/coreos/fedora-coreos-pipeline/blob/3e14744a65c7c73b0b1bd069a946df3a73d919af/jobs/build.Jenkinsfile#L118-L121) to easily check which commits it's running with.

PR #484, opened 2022-03-09
jobs/build: also early archive QEMU image

jlebon:
It's already in `meta.json` so we can't *not* upload it. Otherwise if
there's a build failure before the full upload, it can cause issues.
E.g. a broken link in the build browser and throwing off upgrade tests.
We could remove it from the JSON but that seems unnecessary.

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -360,16 +360,16 @@ lock(resource: "build-${params.STREAM}") {
             return
         }
 
-        // Do an Early Archive of just the OSTree. This has the
-        // desired side effect of reserving our build ID before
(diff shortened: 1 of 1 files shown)


cgwalters:
(I think https://github.com/coreos/coreos-assembler/issues/2685 will fix this because the "reserved build id" will be in the registry, not in S3)

cgwalters:
Ah sorry, I only read the first thing and not the later comments before merging.

jlebon:
No worries, revert in https://github.com/coreos/fedora-coreos-pipeline/pull/485.
Will open an issue against cosa for the higher-level issue.

PR #485, opened 2022-03-09
Revert "jobs/build: also early archive QEMU image"

jlebon:
This reverts commit 29e179bcfc3f52a110d867d06c50446de3ff7a99.

This uploads the QEMU image uncompressed, which is not what we want.
See discussions in
https://github.com/coreos/fedora-coreos-pipeline/pull/484.


dustymabe (approved):
LGTM

PR #487, opened 2022-03-09
jobs: only run release job once per build

jlebon:
Before, we were running the release job multiple times per build: once
for each architecture. This was done by triggering it at the end of both
the x86_64 build and multi-arch builds.

However, it's possible that the x86_64 build fails while the multi-arch
build succeeds. In that case, we don't want to release just for e.g.
aarch64 if the x86_64 build is missing or broken.

Fix this by making only the x86_64 build trigger the release job. This
way, it'll only be triggered once and only if the x86_64 build
succeeded. Add locking in the release job and the multi-arch job to
ensure that the release job cannot execute until the multi-arch job has
completed, whether successfully or not.

This also allows us to hack around issues in the tools used by the
release job which don't fully support releasing for a single
architecture only. These are gaps that we should still fix regardless.

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -77,7 +77,7 @@ properties([
              trim: true),
       string(name: 'ARCH',
              description: 'The target architecture',
-             defaultValue: 'aarch64',
+             choices: streams.additional_arches,
              trim: true),
       string(name: 'FCOS_CONFIG_COMMIT',
              description: 'The exact config repo git commit to build against',
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -127,7 +127,10 @@ def pod_label = "cosa-${UUID.randomUUID().toString()}"
 echo "Waiting for build-${params.STREAM}-${params.ARCH} lock"
 currentBuild.description = "[${params.STREAM}][${params.ARCH}] Waiting"
 
-lock(resource: "build-${params.STREAM}-${params.ARCH}") {
+// build lock: we don't want multiple concurrent builds for the same stream and
+// arch (though this should work fine in theory)
+// release lock: we want to block the release job until we're done
+lock(resource: "build-${params.STREAM}-${params.ARCH}", extra: [[resource: "release-${params.VERSION}-${params.ARCH}"]]) {
     currentBuild.description = "[${params.STREAM}][${params.ARCH}] Running"
 
     podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
@@ -503,24 +506,6 @@ EOF
                     string(name: 'S3_STREAM_DIR', value: s3_stream_dir),
                     string(name: 'ARCH', value: basearch),
                     string(name: 'FCOS_CONFIG_COMMIT', value: params.FCOS_CONFIG_COMMIT)
-                ]
-            }
-        }
-
-        // For now, we auto-release all non-production streams builds. That
-        // way, we can e.g. test testing-devel AMIs easily.
-        //
-        // Since we are only running this stage for non-production (i.e. mechanical
-        // and development) builds we'll default to not doing AWS AMI replication.
-        // That can be overridden by the user setting the AWS_REPLICATION parameter
-        // to true, overriding the default (false).
-        if (official && !(params.STREAM in streams.production)) {
-            stage('Publish') {
-                build job: 'release', wait: false, parameters: [
-                    string(name: 'STREAM', value: params.STREAM),
-                    string(name: 'ARCHES', value: basearch),
-                    string(name: 'VERSION', value: newBuildID),
-                    booleanParam(name: 'AWS_REPLICATION', value: params.AWS_REPLICATION)
                 ]
             }
         }
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -56,7 +56,7 @@ properties([
              choices: (streams.development + streams.production + streams.mechanical),
              description: 'Fedora CoreOS stream to build'),
       string(name: 'VERSION',
-             description: 'Override default versioning mechanism',
+             description: 'Build version',
              defaultValue: '',
              trim: true),
       booleanParam(name: 'FORCE',
@@ -139,6 +139,7 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}") {
         """)
 
         // declare these early so we can use them in `finally` block
+        assert params.VERSION != ""
         def newBuildID = params.VERSION
         def basearch = params.ARCH
 
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -21,7 +21,7 @@ properties([
              trim: true),
       string(name: 'ARCHES',
              description: 'Space-separated list of target architectures',
-             defaultValue: 'x86_64 aarch64',
+             defaultValue: '',
              trim: true),
       // Default to true for AWS_REPLICATION because the only case
       // where we are running the job by hand is when we're doing a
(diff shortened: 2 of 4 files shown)


jlebon:
Independently, we also need to enhance the upgrade tests so it knows to keep going backwards in the release index until it finds a release with a matching architecture *and* a QEMU image (after we stop [being misleading](https://github.com/coreos/coreos-assembler/pull/2746) about that).

dustymabe:
> But anyway, I think we don't want to release any !x86_64 builds if
> x86_64 itself wasn't released. So let's add a check for this.

hmm - the problem is that depending on load and other factors `aarch64` can finish before `x86_64`. `x86_64` could still be running (not failed) at the point we'd have to make that decision. 

How tricky is it to make the release job only affect the arches it was told to?

jlebon:
> > But anyway, I think we don't want to release any !x86_64 builds if
> > x86_64 itself wasn't released. So let's add a check for this.
> 
> hmm - the problem is that depending on load and other factors `aarch64` can finish before `x86_64`. `x86_64` could still be running (not failed) at the point we'd have to make that decision.

OK, pushed a fix for that!

> How tricky is it to make the release job only affect the arches it was told to?

I'm not sure, I think it may require something similar to what we do for `builds.json`. To clarify though, do you disagree with this:

> I think we don't want to release any !x86_64 builds if x86_64 itself wasn't released

?

jlebon:
This is not quite right, actually. I also think we need something stronger here.

jlebon:
@dustymabe and I discussed this OOB and we came up with a more elegant approach.
Updated this now!

Requires: https://github.com/coreos/coreos-assembler/pull/2749

jobs/build-arch.Jenkinsfile
```diff
@@ -139,6 +142,7 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}") {
         """)
 
         // declare these early so we can use them in `finally` block
+        assert params.VERSION != ""
```
dustymabe, line 145: Is there no way to mark a parameter as required when declaring the parameter above?

jobs/release.Jenkinsfile
```diff
@@ -89,20 +96,30 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
...
-                withEnv(["SRC_IMAGE=${image_path}",
-                         "DEST_IMAGE=docker://${quay_registry}:${params.STREAM}"]) {
-                    sh('skopeo copy --authfile=${OSCONTAINER_SECRET} oci-archive://${SRC_IMAGE} ${DEST_IMAGE}')
-                }
-            }
+        def basearches = ""
+        if (params.ARCHES == '') {
+            basearches = shwrapCapture("jq -r '.builds | map(select(.id == \"${params.VERSION}\"))[].arches[]' builds/builds.json").split()
+        } else {
+            params.ARCHES.split()
```
dustymabe, line 103:
```suggestion
            basearches = params.ARCHES.split()
```

dustymabe (changes requested):
Thanks for digging into this!

I have a few small suggestions (comments inline) and one large suggestion:

I'd like to be able to drive the alt arches at runtime from the toplevel `build` job and have everything else flow from there as passed parameters. This means less directly consuming the variables in streams.groovy, but rather using the streams.groovy variable as an input (and also as a means to sanity check).. Something like this:

```diff
diff --git a/jobs/build.Jenkinsfile b/jobs/build.Jenkinsfile
index 1e8a2aa..357d7a3 100644
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -57,6 +57,10 @@ properties([
              description: 'Override default versioning mechanism',
              defaultValue: '',
              trim: true),
+      string(name: 'OTHER_ARCHES',
+             description: 'Space-separated list of other target architectures',
+             defaultValue: streams.secondary_arches.join(" "),
+             trim: true),
       booleanParam(name: 'FORCE',
                    defaultValue: false,
                    description: 'Whether to force a rebuild'),
@@ -387,7 +391,7 @@ lock(resource: "build-${params.STREAM}") {
         }
 
         stage('Fork Multi-Arch Builds') {
-            for (arch in streams.secondary_arches) {
+            for (arch in params.OTHER_ARCHES.split()) {
                 build job: 'build-arch', wait: false, parameters: [
                     booleanParam(name: 'FORCE', value: params.FORCE),
                     booleanParam(name: 'MINIMAL', value: params.MINIMAL),
@@ -635,7 +639,7 @@ lock(resource: "build-${params.STREAM}") {
             stage('Publish') {
                 build job: 'release', wait: false, parameters: [
                     string(name: 'STREAM', value: params.STREAM),
-                    string(name: 'ARCHES', value: basearch),
+                    string(name: 'ARCHES', value: basearch + " " + params.OTHER_ARCHES,
                     string(name: 'VERSION', value: newBuildID),
                     booleanParam(name: 'AWS_REPLICATION', value: params.AWS_REPLICATION)
                 ]
diff --git a/jobs/release.Jenkinsfile b/jobs/release.Jenkinsfile
index 268905a..7c26888 100644
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -21,7 +21,7 @@ properties([
              trim: true),
       string(name: 'ARCHES',
              description: 'Space-separated list of target architectures',
-             defaultValue: '',
+             defaultValue: "x86_64" + " " + streams.secondary_arches.join(" "),
              trim: true),
       // Default to true for AWS_REPLICATION because the only case
       // where we are running the job by hand is when we're doing a
@@ -65,7 +65,7 @@ def quay_registry = "quay.io/coreos-assembler/fcos"
 // jobs run for the same stream, but that really shouldn't happen. Anyway, if it
 // *does*, this makes sure they're run serially.
 // Also lock version-arch-specific locks to make sure these builds are finished.
-def locks = streams.secondary_arches.each {[resource: "release-${params.VERSION}-${it}"]}
+def locks = params.ARCHES.split().each {[resource: "release-${params.VERSION}-${it}"]}
 lock(resource: "release-${params.STREAM}", extra: locks) {
 podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
     node(pod_label) { container('coreos-assembler') { try {
@@ -96,11 +96,12 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
             """)
         }
 
-        def basearches = ""
-        if (params.ARCHES == '') {
-            basearches = shwrapCapture("jq -r '.builds | map(select(.id == \"${params.VERSION}\"))[].arches[]' builds/builds.json").split()
-        } else {
-            params.ARCHES.split()
+        def basearches = params.ARCHES.split()
+        def builtarches = ""
+        builtarches = shwrapCapture("jq -r '.builds | map(select(.id == \"${params.VERSION}\"))[].arches[]' builds/builds.json").split()
+        if builtarches != basearches {
+            "Some requested architectures did not successfully build.
+            basearches = builtarches
         }
 
         for (basearch in basearches) {

```

streams.groovy
```diff
@@ -8,6 +8,9 @@ production = ['testing', 'stable', 'next']
 development = ['testing-devel'] + next_devel
 mechanical = ['rawhide', 'branched' /* 'bodhi-updates', 'bodhi-updates-testing' */]
 
+// list of secondary architectures we support
+secondary_arches = ['aarch64']
```
dustymabe, line 12: any ideas for another term than "secondary". I think that's a loaded term in Fedora from the time where secondary arches were built in an entirely different build env etc..

`alternative_arches` maybe?

jlebon:
> I'd like to be able to drive the alt arches at runtime from the toplevel build job and have everything else flow from there as passed parameters.

Hmm, that means that we wouldn't release anything if even one !x86_64 arch fails, which isn't what we want right?

streams.groovy
jlebon: Went with `additional_arches`!

jobs/release.Jenkinsfile
jlebon: Good catch! Too much Rust, heh.

jobs/build-arch.Jenkinsfile
jlebon: Not AFAIK.

dustymabe:
> Hmm, that means that we wouldn't release anything if even one !x86_64 arch fails, which isn't what we want right?

I don't think so. At least that's not my understanding. The diff I proposed prints a warning but then continues.

dustymabe (approved):
LGTM - i'll open a followup PR with my patch and we'll discuss the merits there.

PR #490, opened 2022-03-14
build/release: allow passing arches from toplevel build job

dustymabe:
This will allow us to specify ADDITIONAL_ARCHES to the toplevel build
job and have it filter all the way through the cascading secondary jobs.
It also means that we're using a single source of truth (the parameter)
for each job rather than mixing the paramater and the additional_arches
definition from `streams.groovy`.

### DIFF
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -25,6 +25,9 @@ properties([
              description: 'Space-separated list of target architectures',
              defaultValue: "x86_64" + " " + streams.additional_arches.join(" "),
              trim: true),
+      booleanParam(name: 'ALLOW_MISSING_ARCHES',
+                   defaultValue: false,
+                   description: 'Allow release to continue even with missing architectures'),
       // Default to true for AWS_REPLICATION because the only case
       // where we are running the job by hand is when we're doing a
       // production release and we want to replicate there. Defaulting
@@ -104,8 +107,15 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
         def builtarches = shwrapCapture("jq -r '.builds | map(select(.id == \"${params.VERSION}\"))[].arches[]' builds/builds.json").split()
         assert builtarches.contains("x86_64"): "The x86_64 architecture was not in builtarches."
         if (!builtarches.containsAll(basearches)) {
-            echo "Some requested architectures did not successfully build! Continuing."
-            basearches = builtarches.intersect(basearches)
+            if (params.ALLOW_MISSING_ARCHES) {
+                echo "Some requested architectures did not successfully build! Continuing."
+                basearches = builtarches.intersect(basearches)
+            } else {
+                echo "ERROR: Some requested architectures did not successfully build"
+                echo "ERROR: Detected built architectures: $builtarches"
+                echo "ERROR: Requested base architectures: $basearches"
+                return
+            }
         }
 
         for (basearch in basearches) {
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -23,7 +23,7 @@ properties([
              trim: true),
       string(name: 'ARCHES',
              description: 'Space-separated list of target architectures',
-             defaultValue: '',
+             defaultValue: "x86_64" + " " + streams.additional_arches.join(" "),
              trim: true),
       // Default to true for AWS_REPLICATION because the only case
       // where we are running the job by hand is when we're doing a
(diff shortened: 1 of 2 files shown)


jobs/release.Jenkinsfile
```diff
@@ -96,11 +96,11 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
...
         }
 
-        def basearches = ""
-        if (params.ARCHES == '') {
-            basearches = shwrapCapture("jq -r '.builds | map(select(.id == \"${params.VERSION}\"))[].arches[]' builds/builds.json").split()
-        } else {
-            basearches = params.ARCHES.split()
+        def basearches = params.ARCHES.split()
+        def builtarches = shwrapCapture("jq -r '.builds | map(select(.id == \"${params.VERSION}\"))[].arches[]' builds/builds.json").split()
+        if builtarches != basearches {
```
dustymabe, line 101: This comparison might need to be improved to make it not depend on ordering.

jlebon: I think if we convert it to a set instead it should work.

jobs/release.Jenkinsfile
```diff
@@ -96,11 +96,11 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
...
+            echo "Some requested architectures did not successfully build!"
```
jlebon, line 102: Hmm, I think we should error out here instead if `basearches` isn't a subset of `builtarches`.

But the issue then is that we won't release at all if aarch64 failed but x86_64 didn't. Which is why the special empty string value for `ARCHES` was added.

jobs/release.Jenkinsfile
dustymabe: I converted the logic to use sets using the `.toSet()` function which I think should work???

jobs/release.Jenkinsfile
dustymabe: Ehh. I think maybe we should make "special empty value" not so special and be more explicit. I added a second commit to add a `ALLOW_MISSING_ARCHES` parameter.

jobs/release.Jenkinsfile
```diff
@@ -96,11 +99,16 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
...
         }
 
-        def basearches = ""
-        if (params.ARCHES == '') {
-            basearches = shwrapCapture("jq -r '.builds | map(select(.id == \"${params.VERSION}\"))[].arches[]' builds/builds.json").split()
-        } else {
-            basearches = params.ARCHES.split()
+        def basearches = params.ARCHES.split()
+        def builtarches = shwrapCapture("jq -r '.builds | map(select(.id == \"${params.VERSION}\"))[].arches[]' builds/builds.json").split()
+        if builtarches.toSet() != basearches.toSet() {
```
jlebon, line 104: Hmm, shouldn't this be a subset check though?

Ahh, looks like `containsAll()` exists on any Iterable: https://docs.groovy-lang.org/latest/html/groovy-jdk/java/lang/Iterable.html#containsAll(java.lang.Object[]).

So I think this could be

```suggestion
        if !builtarches.containsAll(basearches) {
```

(Double-check in the script console first.)

jobs/release.Jenkinsfile
```diff
@@ -96,11 +99,16 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
...
+            if params.ALLOW_MISSING_ARCHES {
+                echo "Some requested architectures did not successfully build! Continuing."
+                basearches = builtarches
```
jlebon, line 107: And then here

```suggestion
                basearches = builtarches.intersect(basearches)
```

jobs/release.Jenkinsfile
```diff
@@ -96,11 +99,16 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
...
+            else {
+                echo "ERROR: Some requested architectures did not successfully build"
```
jlebon, line 109: Could we print here `basearches` and `builtarches` to make it easier to debug?

jobs/release.Jenkinsfile
jlebon: And also maybe an `assert` here to sanity-check that `x86_64` is there.

jobs/release.Jenkinsfile
dustymabe: will do

jobs/release.Jenkinsfile
jlebon: I think we should keep the basearches in the parameter as the source of truth. I.e. we shouldn't release for an arch just because it's built. So then all `ALLOW_MISSING_ARCHES` does is to allow releasing a subset, not a different set.

dustymabe: ok I think I get it now. I needed to work through an example:

```
builtarches = ["1", "2", "3"]
basearches = ["1", "2", "4"]

basearches = builtarches.intersect(basearches) -> ["1", "2"]
```

Wheras my code would have `basearches` as `["1", "2", "3"]`

dustymabe:
ok new changes pushed. I'm going to test this in staging and see how it does.

jobs/release.Jenkinsfile
```diff
@@ -96,11 +99,19 @@ podTemplate(cloud: 'openshift', label: pod_label, yaml: pod) {
             """)
         }
 
-        def basearches = ""
-        if (params.ARCHES == '') {
-            basearches = shwrapCapture("jq -r '.builds | map(select(.id == \"${params.VERSION}\"))[].arches[]' builds/builds.json").split()
-        } else {
-            basearches = params.ARCHES.split()
+        def basearches = params.ARCHES.split()
```
jlebon, line 102: Minor: we could move this up to before the locks so that we only split once.

jlebon (approved):

jobs/release.Jenkinsfile
dustymabe: done

jlebon (approved):

jlebon (approved):

PR #491, opened 2022-03-22
Add `notify-slack` knob to configmap

jlebon:
This makes it easier for developers to turn off Slack notifications. In
addition, it allows us to drop the try/catch blocks since we know it
should always succeed if attempted.


dustymabe (approved):
LGTM

PR #494, opened 2022-03-22
jobs: load pipeline configmap once upfront

jlebon:
Rather than shelling out to `oc` multiple times to read configmap knobs,
just read the whole configmap once upfront and parse it into a variable.

This is not only more efficient but also avoids issues with the current
execution environment not necessarily having `oc` available.

I left `developer_prefix` alone for now since we're going to nuke it
anyway.

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -1,6 +1,7 @@
 import org.yaml.snakeyaml.Yaml;
 
-def pipeutils, streams, official, developer_prefix, repo, src_config_url, src_config_ref, s3_bucket, gp
+def pipeutils, streams, official, developer_prefix, repo, gp
+def src_config_url, src_config_ref, s3_bucket, notify_slack
 node {
     checkout scm
     pipeutils = load("utils.groovy")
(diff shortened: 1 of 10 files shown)


jlebon:
While there, I did a few minor cleanups, like deleting unused variables and moving the declaration of the `fcos_config_commit` variable lower down to where it's actually set.

jobs/build-arch.Jenkinsfile
```diff
@@ -20,11 +20,8 @@ node {
         echo "Running in developer mode on ${env.JENKINS_URL}."
     }
 
-    developer_prefix = pipeutils.get_config('developer-prefix')
-    src_config_url = pipeutils.get_config('source-config-url')
-    src_config_ref = pipeutils.get_config('source-config-ref')
-    s3_bucket = pipeutils.get_config('s3-bucket')
-    gcp_gs_bucket = pipeutils.get_config('gcp-gs-bucket')
+    pipecfg = pipeutils.load_config()
+    developer_prefix = pipecfg['developer-prefix']
```
dustymabe, line 24: Any chance we could use `pipecfg`, but still declare/define these variables up front? i.e. we'd add a new var for `notify_slack` and here we would `notify_slack = pipecfg['notify_slack']`.

I think it makes later parts of the code easier to read if you don't worry about where the definition of the variable comes from, but just what the variable means. The extra `pipecfg[]` in each use now makes you think about that.

PR #496, opened 2022-03-23
jobs: fix comparison for `src_config_ref`

jlebon:
I erroneously thought in 72406a1 that we should change those comparisons
to `null` since `pipecfg['nonexistent']` returns `null`. But actually
the configmap template will always define the key anyway and the default
value is the empty string.


dustymabe (approved):
LGTM

PR #497, opened 2022-03-23
Drop logic for developer_prefix in pipelines

dustymabe:
The whole concept is a bit confusing and there is a decent amount of
conditional logic in there. If someone needs to do some hacking for
development we'll just give them their own project in openshift and
a fresh s3 bucket in another account. That way there's no risk of
messing with the official releases, but they are exercising similar
code paths.

### DIFF
--- a/deploy
+++ b/deploy
@@ -12,8 +12,6 @@
             # OR
             --cosa https://github.com/jlebon/coreos-assembler
 
-    Deleting developer pipeline:
-        ./deploy --delete-devel
 '''
 
 import os
@@ -33,19 +31,12 @@ def main():
     args = parse_args()
 
     if targeting_official_namespace(args) and not args.official:
-        eprint("Refusing to create developer resource in official namespace.")
-        eprint("Use --official to create official resources.")
+        eprint("Refusing to create resources in official namespace.")
+        eprint("Use --official to create resources there.")
         return 1
 
     resources = process_template(args)
-
-    if args.update:
-        update_resources(args, resources)
-    else:
-        assert args.delete_devel
-
-        delete_developer_resources(args, resources)
-
+    update_resources(args, resources)
 
 def targeting_official_namespace(args):
     ctx_name = subprocess.check_output([args.oc_cmd, 'config', 'current-context'],
@@ -70,20 +61,10 @@ def targeting_official_namespace(args):
 
 def parse_args():
     parser = argparse.ArgumentParser()
-    action = parser.add_mutually_exclusive_group(required=True)
-    action.add_argument("--update", action='store_true',
-                        help="Create or update resources")
-    action.add_argument("--delete-devel", action='store_true',
-                        help="Delete developer resources")
     parser.add_argument("--official", action='store_true',
-                        help="Whether to update official resources "
-                             "(implies --all)")
-    parser.add_argument("--all", action='store_true',
-                        help="Whether to update all resources")
+                        help="Allows updating resources in official pipeline")
     parser.add_argument("--dry-run", action='store_true',
                         help="Only print what would happen")
-    parser.add_argument("--prefix", help="Developer prefix for resources",
-                        default=get_username())
     parser.add_argument("--pipeline", metavar='<URL>[@REF]',
                         help="Repo and ref to use for pipeline code")
     parser.add_argument("--config", metavar='<URL>[@REF]',
@@ -103,15 +84,6 @@ def parse_args():
 
     args = parser.parse_args()
 
-    if args.official:
-        args.all = True
-
-    # just sanity check we have a prefix
-    assert len(args.prefix)
-    assert not args.prefix.endswith('-'), "Prefix must not end with dash"
-
-    # e.g. jlebon --> jlebon-fedora-coreos-pipeline
-    args.prefix += '-'
 
     return args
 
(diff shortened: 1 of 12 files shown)


jobs/build.Jenkinsfile
```diff
@@ -267,7 +221,7 @@ lock(resource: "build-${params.STREAM}") {
             def version = ""
             if (params.VERSION) {
                 version = "--version ${params.VERSION}"
-            } else if (official) {
+            } else {
                 def new_version = shwrapCapture("/var/tmp/fcos-releng/scripts/versionary.py")
```
bgilbert, line 231: Hmm, I don't think we should encourage anyone to produce development artifacts versioned like official ones.  It seems likely to create confusion.

dustymabe: That's fair. Added a new commit to address this concern.

dustymabe:
rebased and added some more fixes

deploy
```diff
@@ -23,67 +21,15 @@ import yaml
...
-}
-
-
 def main():
     args = parse_args()
-
-    if targeting_official_namespace(args) and not args.official:
-        eprint("Refusing to create developer resource in official namespace.")
-        eprint("Use --official to create official resources.")
-        return 1
```
jlebon, line 38: Can we keep this check but just delete the "developer" word in there? It's a simple fail-safe to ensure that you're not deploying against the official namespace by mistake.

jobs/build-arch.Jenkinsfile
```diff
@@ -190,26 +170,16 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}", extra: [[resource: "rele
...
-            // towards testing a specific combination of (cosa, pipeline, fcos config),
-            // not a full duplication of all the prod streams. One can always instantiate
-            // a second prefix to test a separate combination if more than 1 concurrent
-            // devel pipeline is needed.
-            s3_stream_dir = "${s3_bucket}/devel/streams/${developer_prefix}"
-          }
         }
 
-        def developer_builddir = "/srv/devel/${developer_prefix}/build"
+        def local_builddir = "/srv/devel/streams/${params.STREAM}"
```
jlebon, line 177: Kinda leaning towards just not supporting this at all.

jobs/release.Jenkinsfile
```diff
@@ -67,7 +67,7 @@ def pod_label = "cosa-${UUID.randomUUID().toString()}"
 def quay_registry = "quay.io/coreos-assembler/fcos"
 
 // Get the list of requested architectures to release
-def basearches = params.ARCHES.split() as List
+def basearches = params.ARCHES.split() as Set
```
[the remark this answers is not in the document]
dustymabe: +1 - I needed to `git pull` - will update

deploy
dustymabe: I was kind of trying to get away from hardcoding the URL in various places but I think maybe it's OK to keep this.

jobs/build-arch.Jenkinsfile
```diff
@@ -177,7 +157,7 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}", extra: [[resource: "rele
         git clone --depth=1 https://github.com/coreos/fedora-coreos-releng-automation /var/tmp/fcos-releng
         """)
 
-        if (official) {
+        if (utils.pathExists("/etc/fedora-messaging-cfg/fedmsg.toml")) {
```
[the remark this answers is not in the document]
dustymabe: this somewhat negates the simplification if we have to add back logic for if the pipeline is "official" or not.

Maybe instead of littering these official checks throughout the code we could have a single check (pipeutils function) that runs up front and does sanity checks based on if it's "official" or not. WDYT?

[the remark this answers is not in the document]
jlebon: > I guess the alternative is that we just keep the notion of "official" but drop just the developer prefix bits.

That sounds good to me. I think it'd make the intent clearer. I definitely agree we should be using it sparingly though.

jobs/build-arch.Jenkinsfile
jlebon: Sure, SGTM!

dustymabe:
ok I reworked this a little bit - focus on the content for now and I'll squash the commits once we agree this is the right path forward.

jobs/build-arch.Jenkinsfile
```diff
@@ -539,7 +510,7 @@ EOF
             }
 
             echo message
-            if (notify_slack == "yes") {
+            if (official) {
```
jlebon, line 513: Meh FWIW, in the spirit of keeping the minimum amount of things gated on `official`, I'd just leave the `notify_slack` knob. It's also not entirely unreasonable to have someone testing changes to Slack messages and targeting their own private channel.

But also OK with this. In that case though, we should also delete the `notify_slack` variable.

jlebon (approved):

jobs/build-arch.Jenkinsfile
dustymabe: yeah I went and cleaned up the extraneous vars. Thanks for catching them.

jlebon (approved):
For the commit title of the second commit, maybe:

> Remove `notify_slack` knob from configmap

Otherwise, LGTM.

dustymabe:
> Remove `notify_slack` knob from configmap

done

PR #501, opened 2022-03-24
jobs/sync-stream-metadata: also sync release notes

jlebon:
Now that we have release notes in the streams repo, also start syncing
it to S3 from which the website will fetch it just like stream metadata.

This will trigger a `stream.metadata.update` message even if only
release notes changed. I'm not aware of any consumers of this today, but
still eventually we should probably split stream/updates/release notes
into separate messages.

### DIFF
--- a/jobs/sync-stream-metadata.Jenkinsfile
+++ b/jobs/sync-stream-metadata.Jenkinsfile
@@ -45,6 +45,16 @@ cosaPod(configMaps: ["fedora-messaging-cfg"], secrets: ["fedora-messaging-coreos
                     stream.metadata.update --stream ${stream}
                 """)
             }
+            // Currently, we always re-upload release notes. We don't want to
+            // falsely emit a stream.metadata.update message when only release
+            // notes changed, and also the way change detection works above
+            // doesn't mesh well with freshly regenerated data.
+            shwrap("""
+                python -c 'import sys, yaml, json; json.dump(yaml.safe_load(sys.stdin.read()), sys.stdout)' \
+                    < release-notes/${stream}.yaml > release-notes/${stream}.json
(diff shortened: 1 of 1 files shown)


jlebon:
Requires: https://github.com/coreos/fedora-coreos-streams/pull/477

travier:
We now have https://github.com/coreos/fedora-coreos-streams/pull/477 in so we should get this one in

travier:
This needs to convert the YAML to JSON for the release notes.

travier:
Ping @dustymabe @jlebon

dustymabe (approved):
LGTM

jobs/sync-stream-metadata.Jenkinsfile
```diff
@@ -29,12 +29,16 @@ cosaPod(configMaps: ["fedora-messaging-cfg"], secrets: ["fedora-messaging-coreos
         // they're different from the checkout, and copy out the new versions
         // if so
         streams.production.each{stream ->
-            for (subdir in ["streams", "updates"]) {
+            shwrap("""
+                python -c 'import sys, yaml, json; json.dump(yaml.safe_load(sys.stdin.read()), sys.stdout)' \
+                    < release-notes/${stream}.yaml > release-notes/${stream}.json
+            """)
+            for (subdir in ["streams", "updates", "release-notes"]) {
```
jlebon, line 36: This will overwrite the newly converted JSON files above.

OK this is messy because it interacts with the naive change detection we try to do. I think rather than trying to get change detection right for release notes, I'd say let's just have it always re-upload them.

Let me try tweaking this.

dustymabe (approved):
LGTM

travier:
Thanks!

PR #504, opened 2022-03-25
jobs/release: convert basearches and builtarches to a Set

dustymabe:
The .intersect() method isn't allowed by default on a List but it is
on a Set [1] so let's just use a Set here instead so we don't get this
ugly error:

```
12:20:49  Scripts not permitted to use staticMethod
          org.codehaus.groovy.runtime.DefaultGroovyMethods
          intersect java.util.List java.lang.Iterable.
```

[1] https://github.com/jenkinsci/script-security-plugin/blob/bcf6cf6ed9603db7e4c10d71ec9c15e604107ec1/src/main/resources/org/jenkinsci/plugins/scriptsecurity/sandbox/whitelists/generic-whitelist#L1011


jlebon (approved):
Nice, this is better anyway!

PR #507, opened 2022-04-02
manifests/pod.yaml: add Azure secrets to pod definition

dustymabe:
This is needed for jobs/build.Jenkinsfile so we can detect when
the Azure secret exists and behave accordingly. I guess this wasn't
needed for jobs/kola-azure.Jenkinsfile because we define the pod
differently with cosaPod() and specify the secret there [1].

This is a fixup to 72f2177.

[1] https://github.com/coreos/fedora-coreos-pipeline/blob/7dde485c8706afbf06ca7da5a6f64a4d2910b63b/jobs/kola-azure.Jenkinsfile#L71-L73

### DIFF
--- a/manifests/pod.yaml
+++ b/manifests/pod.yaml
@@ -24,6 +24,9 @@ spec:
          value: /.aws-fcos-builds-bot-config/config
        - name: AWS_FCOS_KOLA_BOT_CONFIG
          value: /.aws-fcos-kola-bot-config/config
+       # This one doesn't have a config subdirectory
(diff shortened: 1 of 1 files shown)


jlebon (approved):
LGTM, but we should really be moving in the other direction IMO (draining secrets from this manifest).

jlebon:
Maybe the main build job could check the configmap instead for whether Azure was set up?

dustymabe:
> Maybe the main build job could check the configmap instead for whether Azure was set up?

do you have an example of that?

PR #508, opened 2022-04-02
kola-azure: bump timeout to 60 minutes

dustymabe:
Even at 45 we're still getting close. Let's just move it to 60.


bgilbert (approved):
LGTM

PR #509, opened 2022-04-05
jobs/bump-lockfile: drop cost=500 hack

jlebon:
I don't quite recall the details of this hack, but I think it had to do
with RPM digests and signing. Though it's been a while now since we've
started stripping digests from lockfiles.

Anyway, let's just drop it for now. If we hit any errors from it, we'll
at least have something to debug.


bgilbert (approved):
LGTM(?)

dustymabe (approved):
LGTM

dustymabe (approved):

PR #510, opened 2022-04-06
jobs/bump-lockfile: send lockfile updates to aarch64 builder

dustymabe:
It turns out that we were properly updating the aarch64 lockfile
in the fetch but we weren't actually sending that to the remote
builder when doing the build/test pieces. This meant the aarch64
tests in bump-lockfile were essentially always testing what was
in git and not the updated lockfiles. Let's forward the changes
to the remote builder.

I had to use patch here because there isn't an easy way to forward
arbitrary files via gangplank so this was just a way to do it.

Also note that originally I tried a different way to apply the patch
like `git apply <(base64 -d <<< 'abcdefg..')` but that fails to render
properly because `<` and `>` get turned into `&lt `and `&rt `because
we are using html templating on the backend in gangplank.

I also took this as an opportunity to break out and call `cosa fetch`
and `cosa build` explicitly in the spec rather than using
`build_artifacts: [base]`, which is a little more opaque to me.

### DIFF
--- a/jobs/bump-lockfile.Jenkinsfile
+++ b/jobs/bump-lockfile.Jenkinsfile
@@ -123,6 +123,14 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
     // The bulk of the work (build, test, etc) is done in the following.
     // We only need to do that work if we have changes.
     if (haveChanges) {
+        // Get the local diff to the manifest lockfiles and capture them in a variable
+        // that we will later apply in a prep stage. This is a huge hack. We need to
+        // convey to the multi-arch builder(s) the updated manifest lockfiles but we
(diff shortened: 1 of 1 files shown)


dustymabe:
Unfortunately I can't test this change fully in the pipeline because staging doesn't allow SSH out (for now). I tested locally using gangplank but I'll probably have to post a followup with a fix for groovy sins that I've made.

jlebon (approved):
LGTM! As a follow-up, can you file an RFE against gangplank so we can circle back and clean this up?

dustymabe:
> LGTM! As a follow-up, can you file an RFE against gangplank so we can circle back and clean this up?

Here's the RFE: https://github.com/coreos/coreos-assembler/issues/2799

PR #511, opened 2022-04-11
Enable `OVERRIDE_PV_PLUGINS_WITH_IMAGE_PLUGINS` and stop pinning plugin versions

jlebon:
```
commit e6b1368f8b5cd6478ca4c6c34b45a9de6e3338c2
Date:   Mon Apr 11 10:32:26 2022 -0400

    manifests/jenkins: enable OVERRIDE_PV_PLUGINS_WITH_IMAGE_PLUGINS

    Jenkins stores configs and plugins in a single directory, which is on a
    PVC mount. When updating the Jenkins image, by default new plugins in
    the image are not copied over into the PVC. If we want to make the image
    plugins authoritative (i.e. if we actually want plugin updates), we need
    to opt into it via an env var.
```
---
```
commit e7d572dbf4b36094bb1ed75755d324ab1bb26a4c
Date:   Mon Apr 11 10:32:27 2022 -0400

    jenkins/controller/plugins: drop version specifications

    A lot of those version strings are stale. Let's default to just
    following the latest releases because it's important to stay on top of
    plugin updates.
```


manifests/jenkins.yaml
```diff
@@ -96,6 +96,10 @@ objects:
             value: >-
               --sessionEviction=86400
               --sessionTimeout=1440
+          # DELTA: This ensures that plugins are actually updated when using a
+          # PVC.
```
dustymabe, line 100:
```suggestion
          # DELTA: When using a PVC this ensures that plugins are
          # actually updated when the image gets updated.
```

dustymabe (approved):
LGTM - one suggestion

PR #514, opened 2022-04-13
Build and publish the KubeVirt containerdisk on S3

rmohr:
Build the containerdisk ociarchive so that it gets uploaded to s3.

xref https://github.com/coreos/fedora-coreos-tracker/issues/1126

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -420,7 +420,7 @@ lock(resource: "build-${params.STREAM}") {
 
             // parallel build these artifacts
             def pbuilds = [:]
-            ["Aliyun", "AWS", "Azure", "AzureStack", "DigitalOcean", "Exoscale", "GCP", "IBMCloud", "Nutanix", "OpenStack", "VirtualBox", "VMware", "Vultr"].each {
+            ["Aliyun", "AWS", "Azure", "AzureStack", "DigitalOcean", "Exoscale", "GCP", "IBMCloud", "KubeVirt", "Nutanix", "OpenStack", "VirtualBox", "VMware", "Vultr"].each {
                 pbuilds[it] = {
                     def cmd = it.toLowerCase()
                     shwrap("""


rmohr:
Here is how I would imagine that a possible upload to a registry may look like:

```
        stage("Push KubeVirt image") {
            withCredentials([file(credentialsId: 'oscontainer-secret', variable: 'OSCONTAINER_SECRET')]) {
                    // push the KubeVirt containerdisk to quay.io/coreos-assembler/fedora-coreos
                    // FIXME: let's find a final location
                    shwrap("""
                    cosa buildextend-kubevirt \
                        --log-level=INFO \
                        --build=${newBuildID} \
                        --upload \
                        --registry quay.io/coreos-assembler
                    """)
            }
        }
```

I am however not entirely sure what the desired way from the coreos perspective is to get the final image for each stream.

jlebon (approved):
LGTM!

For uploading, I think we want something similar to [this](https://github.com/coreos/fedora-coreos-pipeline/blob/03f736071905620148cc0faa08ba9baea62b00c0/jobs/release.Jenkinsfile#L127-L135). It would be in the release job as well (could just be right after that block of code, or in `parallel`).

We could reuse the same secret too.

Maybe let's do that in a separate PR instead where we can bikeshed on the new container name? Edit: but fine to do it here too if you prefer!

Also, can you confirm how exactly the image would be used by users? Would it go in the `VirtualMachineInstance` spec like [the one in the last example here](https://kubevirt.io/user-guide/virtual_machines/disks_and_volumes/#containerdisk-workflow-example) which has `image: vmidisks/fedora25:latest`?

rmohr:
Added documentation steps:

 * https://github.com/coreos/fedora-coreos-docs/pull/394

I guess to test the documentation end-to-end, ignition and afterburn need to be released this PR needs to be merged so that kubevirt is actually built? Let me know what you think about the docs. I tested all individual steps with custom builds and so forth, but I am sure that I still have small mistakes there :)

rmohr:
> Maybe let's do that in a separate PR instead where we can bikeshed on the new container name? Edit: but fine to do it here too if you prefer!

Only building and releasing the s3 image with this PR sounds good to me.

rmohr:
> Maybe let's do that in a separate PR instead where we can bikeshed on the new container name? Edit: but fine to do it here too if you prefer!

Separate PR is here: https://github.com/coreos/fedora-coreos-pipeline/pull/515

rmohr:
Closing this one since I am not actively working on it. Doing some cleanup. Thanks everyone for your help.

PR #519, opened 2022-04-27
jobs/build: cut image building parallelism in half

jlebon:
We're seeing issues where the pipeline sometimes hits EAGAIN on `fork`:

https://github.com/coreos/fedora-coreos-tracker/issues/1186

This may have started happening more frequently after the recent
addition of VirtualBox.

Anyway, it's kinda silly to build that many images in parallel. There's
diminishing returns and eventually a performance cost. So let's do it in
two rounds instead. (Ideally instead it'd be a kind of task pool, but
let's not try to implement that in Groovy.)

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -418,17 +418,19 @@ lock(resource: "build-${params.STREAM}") {
                 }
             }
 
-            // parallel build these artifacts
-            def pbuilds = [:]
-            ["Aliyun", "AWS", "Azure", "AzureStack", "DigitalOcean", "Exoscale", "GCP", "IBMCloud", "Nutanix", "OpenStack", "VirtualBox", "VMware", "Vultr"].each {
-                pbuilds[it] = {
+            // parallel build these artifacts (over two runs to stay in the sweet spot and avoid hitting PID limits)
+            def platforms = ["Aliyun", "AWS", "Azure", "AzureStack", "DigitalOcean", "Exoscale", "GCP", "IBMCloud", "Nutanix", "OpenStack", "VirtualBox", "VMware", "Vultr"]
+            def pbuilds = platforms.collectEntries {
+                [it, {
                     def cmd = it.toLowerCase()
                     shwrap("""
                     cosa buildextend-${cmd}
                     """)
-                }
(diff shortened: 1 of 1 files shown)


cgwalters (approved):
Looks sane, but I think what we really want is to have `cosa buildextend -j N` or so...

dustymabe:
Hmm. Let's be a bit smarter here and put the ones that are long running together. Let me look at the run history and make a suggestion.

jlebon:
> Hmm. Let's be a bit smarter here and put the ones that are long running together. Let me look at the run history and make a suggestion.

I'm not sure it's worth trying to optimize this further vs e.g. pooling as suggested. The step itself only takes ~5 minutes which is tiny in comparison to a full pipeline run (~1h30m). It'd make adding new platforms trickier too.

dustymabe:
> I'm not sure it's worth trying to optimize this further vs e.g. pooling as suggested. The step itself only takes ~5 minutes which is tiny in comparison to a full pipeline run (~1h30m). It'd make adding new platforms trickier too.

Yeah OK. I vaguely remembered the DigitalOcean one taking more than 10 minutes at one point in the past and one other one being long too and was going to try to put them in the same parallel bucket. Looking at the most recent runs it seems like the times are really too bad.

jobs/build.Jenkinsfile
```diff
@@ -418,17 +418,19 @@ lock(resource: "build-${params.STREAM}") {
                 }
             }
 
-            // parallel build these artifacts
-            def pbuilds = [:]
-            ["Aliyun", "AWS", "Azure", "AzureStack", "DigitalOcean", "Exoscale", "GCP", "IBMCloud", "Nutanix", "OpenStack", "VirtualBox", "VMware", "Vultr"].each {
-                pbuilds[it] = {
+            // parallel build these artifacts over two runs
```
dustymabe, line 421: The first question I have in the future when I read this code is "why?". Why split it over two runs when no steps depend on each other?

We know the answer now.. Let's add a comment explaining?

dustymabe (approved):
LGTM

dustymabe:
Thanks for working on this.

PR #523, opened 2022-05-12
kola-openstack.Jenkinsfile: Allow rerun success on openstack platform

gursewak1997:
Allow rerun success on openstack platform. The rerun success tag was implemented in https://github.com/coreos/coreos-assembler/pull/2857.
 
 Issue: https://github.com/coreos/coreos-assembler/issues/2853


dustymabe (approved):
LGTM

PR #524, opened 2022-05-18
jobs/bump-lockfile: fix retrigger logic

dustymabe:
The git repo is under src/config. Without this we get an error
about not being in a git repo. Fixes 237a03b.


bgilbert (approved):
LGTM

PR #527, opened 2022-05-26
jobs/sync-stream-metadata: fix release notes extension

jlebon:
It's `yml`, not `yaml`.


dustymabe:
hmm. I typically prefer the `yaml` form.. Do we have a convention here? Should we change the source?

dustymabe (approved):

jlebon:
> hmm. I typically prefer the `yaml` form.. Do we have a convention here? Should we change the source?

I also do, but not a strong preference.

PR #387, opened 2021-09-06
multi-arch-pipeline: final enablement bits for aarch64

dustymabe:
```
commit 3cdd1f846019e2883989a0557454aaa6fb207db9
Author: Dusty Mabe <dusty@dustymabe.com>
Date:   Mon Sep 6 00:09:59 2021 -0400

    Revert "Jenkinsfile: limit multi-arch-pipeline to non production builds"
    
    This reverts commit 431f7e8ad33ed3e56d9f8c9fd740c9c0d79f9272.
    We want them for production builds now too.

commit 869a78cc49ca232a2dff160870d90aad0e5319f0
Author: Dusty Mabe <dusty@dustymabe.com>
Date:   Sun Sep 5 16:28:20 2021 -0400

    multi-arch-pipeline: re-enable aws upload/testing
    
    Now that we've hacked in console=ttyS0 (see [1]) now we can
    start to upload/test there again.
    
    [1] https://github.com/coreos/fedora-coreos-tracker/issues/920

commit 2307ae760222e8c3d1cca369e1ded8beaf851f73
Author: Dusty Mabe <dusty@dustymabe.com>
Date:   Sun Sep 5 16:23:20 2021 -0400

    multi-arch-pipeline: hack in ttyS0 for aarch64 aws image
    
    This is a workaround to get console=ttyS0,115200n8 into the
    aarch64 AWS image. It does so by applying the following patch
    to gf-platformid:
    
    ```diff
    diff --git a/usr/lib/coreos-assembler/gf-platformid b/usr/lib/coreos-assembler/gf-platformid
    index 2912b322c..36d089651 100755
    --- a/usr/lib/coreos-assembler/gf-platformid
    +++ b/usr/lib/coreos-assembler/gf-platformid
    @@ -46,7 +46,11 @@ blscfg_path=$(coreos_gf glob-expand /boot/loader/entries/ostree-*.conf)
     coreos_gf download "${blscfg_path}" "${tmpd}"/bls.conf
     # Remove any platformid currently there
     sed -i -e 's, ignition.platform.id=[a-zA-Z0-9]*,,g' "${tmpd}"/bls.conf
    -sed -i -e 's,^\(options .*\),\1 ignition.platform.id='"${platformid}"',' "${tmpd}"/bls.conf
    +if [ "${platformid}" == 'aws' ]; then
[...]

PR #532, opened 2022-05-27
RFC: Update aarch64 aws console hack to use platforms.yaml

bgilbert:
Switch to using the platforms.yaml mechanism from https://github.com/coreos/coreos-assembler/pull/2400 to configure the aarch64 AWS console until the long-term configuration in https://github.com/coreos/fedora-coreos-config/pull/1181 lands, then use that config instead.

No functional change to build artifacts intended.

Follow-up to #387.  Needs to merge alongside https://github.com/coreos/coreos-assembler/pull/2400.

Untested.

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -322,7 +322,34 @@ stages:
     - cosa buildextend-openstack
     # Hack for serial console on aarch64 aws images
     # see https://github.com/coreos/fedora-coreos-tracker/issues/920#issuecomment-914334988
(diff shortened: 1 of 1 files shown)


jobs/build-arch.Jenkinsfile
```diff
@@ -322,7 +322,34 @@ stages:
...
+        openstack:
+          grub_commands:
+            - serial --speed=115200
+            - terminal_input serial console
+            - terminal_output serial console
+        qemu:
+          grub_commands:
+            - serial --speed=115200
+            - terminal_input serial console
+            - terminal_output serial console
```
dustymabe, line 350: for this surgical piece, are these lines required?

jobs/build-arch.Jenkinsfile
```diff
@@ -322,7 +322,34 @@ stages:
...
+    - |
+      if [ "$(uname -m)" = aarch64 ] && [ ! -f src/config/platforms.yaml ]; then
+          cat > src/config/platforms.yaml <<EOG
+      aarch64:
+        aws:
+          grub_commands:
+            - serial --speed=115200
+            - terminal_input serial console
+            - terminal_output serial console
```
dustymabe, line 333: do we need the grub_commands part here? I think we only want to affect the karg here (like what we were doing previously with the hack) if we can.

bgilbert: We need the GRUB commands if we want to maintain 100% compatibility with the existing configuration until https://github.com/coreos/fedora-coreos-config/pull/1181 lands.

jobs/build-arch.Jenkinsfile
bgilbert: xref https://github.com/coreos/fedora-coreos-pipeline/pull/532#discussion_r885166415

dustymabe:
I think we might need to use a different approach here. For the platforms.yaml (or platforms.json) new flow (see https://github.com/coreos/coreos-assembler/pull/2400) that information gets embedded into the previous disk image (i.e. the qemu qcow2 that the AWS image is derived from). So we'd have to bake it into the qemu artifact too if we wanted it to affect the AWS one. Maybe we should just update the hack/patch to apply to only the case where platforms.json doesn't exist.

dustymabe:
alternative in https://github.com/coreos/fedora-coreos-pipeline/pull/533

bgilbert:
Closing in favor of https://github.com/coreos/fedora-coreos-pipeline/pull/533.

PR #533, opened 2022-05-31
jobs/build-arch: update hack for aarch64/aws serial console

dustymabe:
This is an update to the serial console hack added in ddd9da9.
This needed updating because of [1] that will update and
rename gf-platformid to gf-set-platform. Eventually once
platforms.yaml exists everywhere we will finally be able
to drop this patch.

[1] https://github.com/coreos/coreos-assembler/pull/2400

The decoded version of this patch looks like:

```
diff --git a/src/gf-set-platform b/src/gf-set-platform
index 3b1c5ae31..df5e0f9d7 100755
--- a/src/gf-set-platform
+++ b/src/gf-set-platform
@@ -59,7 +59,13 @@ blscfg_path=$(coreos_gf glob-expand /boot/loader/entries/ostree-*.conf)
 coreos_gf download "${blscfg_path}" "${tmpd}"/bls.conf
 # Remove any platformid currently there
 sed -i -e 's, ignition.platform.id=[a-zA-Z0-9]*,,g' "${tmpd}"/bls.conf
-sed -i -e '/^options / s,$, ignition.platform.id='"${platformid}"',' "${tmpd}"/bls.conf
+if [ "$(coreos_gf exists /boot/coreos/platforms.json)" != "true" -a "${platformid}" == 'aws' ]; then
+    # Our platform is AWS and we still need the console=ttyS0 hack for the legacy
+    # (no platforms.yaml) path.
+    sed -i -e 's|^\(options .*\)|\1 ignition.platform.id='"${platformid}"' console=ttyS0,115200n8|' "${tmpd}"/bls.conf
+else
+    sed -i -e '/^options / s,$, ignition.platform.id='"${platformid}"',' "${tmpd}"/bls.conf
+fi
 if [ -n "$remove_kargs" ]; then
     # Remove existing qemu-specific kargs
     sed -i -e '/^options / s@ '"${remove_kargs}"'@@' "${tmpd}"/bls.conf

```


dustymabe:
needs testing

bgilbert (approved):
LGTM

PR #537, opened 2022-06-04
jobs/build-arch: drop aarch64 aws serial console hack

bgilbert:
Once https://github.com/coreos/fedora-coreos-config/pull/1181 has shipped to all streams, drop the aarch64 aws console hack here because it'll no longer be doing anything.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -54,6 +54,7 @@ default_artifacts:
     - openstack
     - hashlist-experimental
   aarch64:
+    - aws
     - azure
   ppc64le:
     - powervs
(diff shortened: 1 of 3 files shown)


dustymabe (approved):

dustymabe:
This needs to be updated now do delete https://github.com/coreos/fedora-coreos-pipeline/blob/92749fe725f3949b34b7c58cda8954c35235ad1f/jobs/build-arch.Jenkinsfile#L470-L481 and add AWS to the list of artifacts to build in parallel.

dustymabe:
rebased this. This can merge in a month after the platforms.yaml change has made it to `stable`.

jlebon (commented):
Don't we also want to drop `aws_aarch64_serial_console_hack` from `config.yaml` and `docs/config.yaml`?

dustymabe:
> Don't we also want to drop `aws_aarch64_serial_console_hack` from `config.yaml` and `docs/config.yaml`?

fixed! good catch

PR #540, opened 2022-06-09
jobs/build-arch: only run AWS/OpenStack tests for aarch64

dustymabe:
AWS doesn't support s390x/ppc64le and we don't currently have an
OpenStack environment that supports either of those arches either.

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -495,18 +495,6 @@ EOF
                     string(name: 'FCOS_CONFIG_COMMIT', value: params.FCOS_CONFIG_COMMIT)
                 ]
             }
-            // XXX: The Kubernetes node e2e tests are currently not supported on !x86_64:
-            // https://github.com/coreos/fedora-coreos-tracker/issues/1044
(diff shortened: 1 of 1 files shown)


jobs/build-arch.Jenkinsfile
```diff
@@ -483,30 +483,32 @@ EOF
...
-            stage('Kola:AWS') {
-                // We consider the AWS kola tests to be a followup job, so we use `wait: false` here.
-                build job: 'kola-aws', wait: false, parameters: [
-                    string(name: 'STREAM', value: params.STREAM),
-                    string(name: 'VERSION', value: newBuildID),
-                    string(name: 'S3_STREAM_DIR', value: s3_stream_dir),
-                    string(name: 'ARCH', value: basearch),
-                    string(name: 'FCOS_CONFIG_COMMIT', value: params.FCOS_CONFIG_COMMIT)
-                ]
+        if (params.ARCH == "aarch64") {
```
jlebon, line 486: Aside: I feel like this knowledge should live outside the pipeline, just not sure exactly how.

jlebon (approved):

PR #541, opened 2022-06-09
Reopen next-devel

bgilbert:
We want to bake the console changes in next before they go to testing.

https://github.com/coreos/fedora-coreos-tracker/issues/567#issuecomment-1150275044

Hold until after the next release.


bgilbert:
Obsoleted by #625.

PR #543, opened 2022-06-14
jobs/release: drop use of `withEnv`

jlebon:
Let's simplify this even further and drop the `withEnv` wrapper.

Note we're switching from single to double quotes here. As a result, we
need to escape the `OSCONTAINER_SECRET` variable since it's in the env
only.


dustymabe (approved):
LGTM

PR #544, opened 2022-06-16
verify s390x builder instructions

LakshmiRavichandran1:


multi-arch-builders/README.md
```diff
@@ -127,22 +127,6 @@ ibmcloud is instance-create $NAME $VPC $ZONE $PROFILE $SUBNET --output json --im
      --user-data @fcos-s390x-builder.ign > out.json
```
LakshmiRavichandran1, line 127: The argument to ssh keys '--keys $SSHKEY' (can be obtained using 'ibmcloud is keys') is missing. is this expected?

dustymabe: In this case we are passing SSH keys via Ignition, so no need to use the cloud keys. See 

https://github.com/coreos/fedora-coreos-pipeline/blob/f3f6c5ebf7688b1f91ff3e29e1d215bceea2d727/multi-arch-builders/fcos-s390x-builder.bu#L19-L26

We can update that list as needed.

multi-arch-builders/README.md
```diff
@@ -127,22 +127,6 @@ ibmcloud is instance-create $NAME $VPC $ZONE $PROFILE $SUBNET --output json --im
...
-and then log in:
-
-```bash
-NIC=$(jq --raw-output .primary_network_interface.id out.json)
-ibmcloud is floating-ip-update fcos-s390x-builder-backup --nic $NIC
-
-INSTANCE=$(jq --raw-output .id out.json)
-IP=$(ibmcloud is instance $INSTANCE --output json \
-     | jq -r '.primary_network_interface.floating_ips[0].address')
-echo "You can now SSH to core@${IP}"
```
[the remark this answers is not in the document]
dustymabe: In this case we use two different floating IPs. One is the backup IP used for us to just verify the new machine is up and functioning before we take down the old machine and move the main IP to it.

So we run floating IP commands twice.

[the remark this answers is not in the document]
LakshmiRavichandran1: thank you for explaining this to me. I am able to get a picture of it now.

multi-arch-builders/README.md
LakshmiRavichandran1: thank you for pointing me to this.

LakshmiRavichandran1:
Closing this PR as the changes are no more valid.

PR #545, opened 2022-06-16
jobs/build: parallelize some stages

dustymabe:
- QEMU and QEMU upgrade tests
- keylime hashlist generation and building the qemu image
- Uploading to Clouds (AWS/GCP)
- Kicking off Cloud Provider test jobs

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -281,11 +281,28 @@ lock(resource: "build-${params.STREAM}") {
             }
         }
 
-        stage('Build QEMU') {
+        // A few independent tasks that can be run in parallel
+        def parallelruns = [:]
+
+        // Generate KeyLime hashes for attestation on builds
+        // This is a POC setup and will be modified over time
+        // See: https://github.com/keylime/enhancements/blob/master/16_remote_allowlist_retrieval.md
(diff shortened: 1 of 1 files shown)


dustymabe:
currently untested.

jobs/build.Jenkinsfile
```diff
@@ -305,36 +322,44 @@ lock(resource: "build-${params.STREAM}") {
...
         }
 
-        stage('Kola:QEMU') {
-            // leave 512M for overhead; VMs are 1G each
-            def parallel = ((cosa_memory_request_mb - 512) / 1024) as Integer
+        // reset for the next batch of independent tasks
+        parallelruns = [:]
+
+        // Kola QEMU tests
+        parallelruns['Kola:QEMU'] = {
```
jlebon, line 329: We should really look at moving this over to `fcosKola` which already knows to parallelize with upgrade tests.

jlebon (approved):
I didn't review the syntax in details, but the high-level looks sane to me!

jobs/build.Jenkinsfile
dustymabe: yeah. I'm torn here. I like that all of the actual commands that we know and love are in this file without a level of indirection, but there is a lot of value in the library too.

PR #546, opened 2022-06-16
jobs/build: only early archive if we have other architectures

dustymabe:
Otherwise there is no point. My goal with this change is to enable
us to run a pipeline with `KOLA_RUN_SLEEP=1` (i.e. a run for debugging
that we don't intend to use the results of) without creating a bunch
of new entries in our build history.


dustymabe:
Talked with @jlebon. Going to explore adding a parameter to the pipeline instead of doing this.

PR #547, opened 2022-06-16
jobs/build: Add NO_UPLOAD parameter to allow for debugging

dustymabe:
This allows for in pipeline debugging without worrying about pushing
artifacts over multiple debug runs.

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -1,6 +1,6 @@
 import org.yaml.snakeyaml.Yaml;
 
-def pipeutils, streams, official
+def pipeutils, streams, official, uploading
 def src_config_url, src_config_ref, s3_bucket, gcp_gs_bucket
 node {
     checkout scm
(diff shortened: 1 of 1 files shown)


jobs/build.Jenkinsfile
```diff
@@ -63,6 +63,9 @@ properties([
       booleanParam(name: 'KOLA_RUN_SLEEP',
                    defaultValue: false,
                    description: 'Just wait forever when we get to running kola tests (temporary)'),
```
jlebon, line 65: Maybe let's add a "Implies NO_UPLOAD." here?

jlebon (commented):
We need to also gate multi-arch triggering (or better, add a `NO_UPLOAD` parameter to that job too and pass it through).

We need to also gate publishing.

jobs/build.Jenkinsfile
dustymabe: done

jobs/build.Jenkinsfile
```diff
@@ -360,11 +371,11 @@ lock(resource: "build-${params.STREAM}") {
...
 
-        // Do an Early Archive of just the OSTree. This has the
-        // desired side effect of reserving our build ID before
-        // we fork off multi-arch builds.
+        // If we are uploading results then let's do an early archive
+        // of just the OSTree. This has the desired side effect of
+        // reserving our build ID before we fork off multi-arch builds.
         stage('Archive OSTree') {
-            if (s3_stream_dir) {
+            if (uploading) {
```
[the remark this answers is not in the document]
dustymabe: yeah, there are enough of these where I'd prefer a separate PR if we go that route.

dustymabe:
> We need to also gate multi-arch triggering (or better, add a `NO_UPLOAD` parameter to that job too and pass it through).

For now I'll just gate it here and maybe add the `NO_UPLOAD` parameter to the multi-arch job when we unify `build` and `build-arch` here soon.

> We need to also gate publishing.

done

jlebon (approved):

jlebon:
I like the replacement of `s3_stream_dir` by `uploading`. Much clearer!

PR #548, opened 2022-06-17
jobs/build: collapse Metal/Metal4k/Live ISO builds into builds for other platforms

dustymabe:
This allows us to take the (Metal{,4k})->Live ISO serial run and
collapse it into the two runs of parallel builds we were doing for all
the other platforms, which will save us time.

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -416,23 +416,31 @@ lock(resource: "build-${params.STREAM}") {
 
         if (!params.MINIMAL) {
 
-            stage("Metal") {
-                parallel metal: {
+            // We will parallel build all the different artifacts for this architecture. We split this up
+            // into two separate parallel runs to stay in the sweet spot and avoid hitting PID limits in
+            // our pipeline environment.
+            //
+            // First, define a list of all the derivative artifacts for this architecture.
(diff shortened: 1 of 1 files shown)


jobs/build.Jenkinsfile
```diff
@@ -416,23 +416,31 @@ lock(resource: "build-${params.STREAM}") {
...
+            // First, define a list of all the derivative artifacts for this architecture.
+            def artifacts = ["Aliyun", "AWS", "Azure", "AzureStack", "DigitalOcean", "Exoscale",
+                             "GCP", "IBMCloud", "Nutanix", "OpenStack", "VirtualBox", "VMware", "Vultr"]
+            // For all architectures we need to build the metal/metal4k artifacts and the Live ISO. Since the
+            // ISO depends on the Metal/Metal4k images we'll make sure to put the Metal* ones in the first run
+            // and the Live ISO in the second run.
+            artifacts.add(0, "Metal")
+            artifacts.add(1, "Metal4k")
+            artifacts.add("Live")
+            // Run the two runs of parallel builds
```
jlebon, line 432: Minor/optional: this comment seems redundant.

jlebon (approved):
Looks sane to me! I think there are other tweaks we could make to save more time but it's good to do it in increments and benchmark.

PR #556, opened 2022-06-21
jobs/build: use parallel for 'Sign Images' and 'OSTree Import' stages

dustymabe:
These two don't depend on one another and don't modify the metadata
files so we should fine running them in parallel.

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -558,9 +558,11 @@ lock(resource: "build-${params.STREAM}") {
         // These steps interact with Fedora Infrastructure/Releng for
         // signing of artifacts and importing of OSTree commits. They
         // must be run after the archive stage because the artifacts
-        // are pulled from their S3 locations.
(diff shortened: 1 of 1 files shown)


jobs/build.Jenkinsfile
```diff
@@ -564,13 +566,15 @@ lock(resource: "build-${params.STREAM}") {
                     --fedmsg-conf /etc/fedora-messaging-cfg/fedmsg.toml
                 """)
             }
-            stage("OSTree Import: Compose Repo") {
+            parallelruns['OSTree Import: Compose Repo') {
```
jlebon, line 569:
```suggestion
            parallelruns['OSTree Import: Compose Repo') = {
```

jlebon (approved):
One bug, LGTM otherwise.

jobs/build.Jenkinsfile
dustymabe: fixed!

jlebon (approved):

PR #557, opened 2022-06-22
jobs/build: add temporary iso-live-login UEFI test with rd.debug

jlebon:
We're trying to debug
https://github.com/coreos/fedora-coreos-tracker/issues/1233 but logs
aren't helping much. We think adding `rd.debug` might help in case it's
e.g. a dracut initqueue issue.

Add a separate test for this on a separate ISO so that we don't affect
the official artifacts.

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -452,6 +452,12 @@ lock(resource: "build-${params.STREAM}") {
                         shwrap("kola testiso -SP --qemu-native-4k --output-dir tmp/kola-metal4k")
                     }, uefi: {
                         shwrap("mkdir -p tmp/kola-uefi")
+                        shwrap("""
+                        mkdir tmp/iso-live-login-with-rd-debug
+                        iso=tmp/iso-live-login-with-rd-debug/test.iso
+                        coreos-installer iso kargs modify --append rd.debug builds/${newBuildID}/${basearch}/*.iso -o \$iso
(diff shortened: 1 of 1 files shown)


jobs/build.Jenkinsfile
```diff
@@ -452,6 +452,12 @@ lock(resource: "build-${params.STREAM}") {
                         shwrap("kola testiso -SP --qemu-native-4k --output-dir tmp/kola-metal4k")
                     }, uefi: {
                         shwrap("mkdir -p tmp/kola-uefi")
+                        shwrap("""
+                        mkdir tmp/iso-live-login-with-rd-debug
+                        iso=tmp/iso-live-login-with-rd-debug/test.iso
+                        coreos-installer iso kargs modify --append rd.debug builds/${newBuildID}/${basearch}/*.iso -o \$iso
+                        kola testiso -S --qemu-firmware=uefi --scenarios iso-live-login,iso-as-disk --qemu-iso \$iso --output-dir tmp/kola-uefi/rd-debug")
+                        """)
                         shwrap("kola testiso -S --qemu-firmware=uefi --scenarios iso-live-login,iso-as-disk --output-dir tmp/kola-uefi/insecure")
```
dustymabe, line 461: shouldn't we drop this invocation so we don't run it twice?

dustymabe (approved):
LGTM

jobs/build.Jenkinsfile
jlebon: I wanted to keep that run so that if it keeps happening with the original one but not the `rd.debug` one, we'll know that the bug is still present, but that `rd.debug` prevents it from triggering.

PR #560, opened 2022-07-06
multi-arch-builders: create a common butane config

dustymabe:
90% of the butane config is shared amongst the builders. Let's create
a builder-common.bu to share the common configuration.


jlebon (approved):
Cool, much better!

PR #561, opened 2022-07-07
jobs: use the 4K testiso tests to test multipath

miabbott:
We follow this pattern in the RHCOS pipeline, so let's do it here too.

Testing the multipath case may have caught something like
coreos/fedora-coreos-config#1827 before it made it to RHCOS.


cgwalters (approved):
(Hmm ultimately I think we need to share the test flows somewhere, no real reason to have them be different I think)

miabbott:
> (Hmm ultimately I think we need to share the test flows somewhere, no real reason to have them be different I think)

Long term, RHCOS pipeline should be using nearly the same code as FCOS pipeline - https://issues.redhat.com/browse/COS-1537

Issue #562, opened 2022-07-13
write test to validate jenkins readiness


dustymabe:
Periodically the latest plugins and the version of jenkins offered by https://github.com/openshift/jenkins can get out of sync where jenkins won't come up. Let's run tests on some time schedule (weekly?) that verify the current configuration in this repo is able to at least bring jenkins up without errors.

This will mean that if we ever need to redeploy in hast we'll be ready.

jlebon:
Let's repurpose this to go a step further and lock down all our plugins and the base Jenkins version itself and e.g. have a job that opens a PR to bump them with a link to documentation on how to test the PR in staging. (And maybe eventually it directly instruments the test in staging itself... or maybe better we request a separate namespace for this.)

This will require some investigations to figure out how we can run depsolving ourselves.

marmijo:
> Let's repurpose this to go a step further and lock down all our plugins and the base Jenkins version itself and e.g. have a job that opens a PR to bump them with a link to documentation on how to test the PR in staging.

We've been discussing possible ways to implement this. One way is to have a job that removes the versions altogether, which would just pull all the latest versions. We then attempt a build like that and "see if it works" (loaded statement). 

I tried to manually build the Jenkins instance without any versions and it failed to load because there seems to be an issue with the openshift/jenkins code. I opened https://github.com/openshift/jenkins/issues/1709 regarding it.

In the mean time, we might have to find an alternative way to automate this.

jlebon:
Was chatting with @marmijo about this today. One suggestion is to have the job on the prod Jenkins instrument the Jenkins on the stage cluster. Something like:
1. Clone repo, edit lockfile, push to coreosbot-releng branch
2. Start `jenkins-s2i` build on stage cluster, pointing at the ref
3. Monitor build and deployment of new Jenkins
4. Do a test FCOS build in the new Jenkins (see https://github.com/coreos/fedora-coreos-pipeline/blob/main/HACKING.md#triggering-builds-remotely)
5. If it succeeds, open PR with branch

We'd need to figure out how to set up authentication across the clusters.

Another approach is to instrument the Jenkins from which the job is running itself, but if the plugins break so hard that the Jenkins never comes back up, we'll never hear back from that job.

And yet another approach is to bring up a second Jenkins in the same namespace. I think for this we'd need to make sure we get the resource naming right so there's no conflicts between the two, but it could get messy. I think it's more conventional to use a separate namespace/cluster for tests like this.

[...]

PR #563, opened 2022-07-13
plugins.txt: freeze plugins to working versions

saqibali-2k:
We ran into a version mismatch between jenkins and the plugins
while setting up jenkins in the staging cluster. Lets freeze the
package versions to avoid the mismatch.


jlebon (approved):
Hah, I meant only freezing the plugins that we need by looking at which ones exactly Jenkins is complaining about. But this works too for now!

jlebon:
Did you verify that the latest Jenkins comes up successfully with this?

dustymabe:
> Hah, I meant only freezing the plugins that we need by looking at which ones exactly Jenkins is complaining about. But this works too for now!

Yeah I told him to just freeze them all. Maybe we can work on https://github.com/coreos/fedora-coreos-pipeline/issues/562 soon?

saqibali-2k:
> > Hah, I meant only freezing the plugins that we need by looking at which ones exactly Jenkins is complaining about. But this works too for now!
> 
> Yeah I told him to just freeze them all. Maybe we can work on #562 soon?

The issue was with  `configuration-as-code`, `ssh-credentials`, and `github`, but yep I just ended up freezing them all.

saqibali-2k:
> Did you verify that the latest Jenkins comes up successfully with this?

⚠️ Currently running into some issues. Lets hold off for now ⚠️

cgwalters:
Saqib is the Elsa of Jenkins plugins.

PR #564, opened 2022-07-13
jobs: set cpu request for pods

miabbott:
It's good practice to be explicit with the requests used by pods, so
let's configure the CPU requests.

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -86,6 +86,10 @@ def strict_build_param = is_mechanical ? "" : "--strict"
 // workflow, only the qemu image is built).
 def cosa_memory_request_mb = 6.5 * 1024 as Integer
 
+// the build pod runs most frequently and does the majority of the computation
+// so give it some healthy CPU shares
+pod = pod.replace("COREOS_ASSEMBLER_CPU_REQUEST", "4")
+
 // substitute the right COSA image and mem request into the pod definition before spawning it
 pod = pod.replace("COREOS_ASSEMBLER_MEMORY_REQUEST", "${cosa_memory_request_mb}Mi")
 pod = pod.replace("COREOS_ASSEMBLER_IMAGE", params.COREOS_ASSEMBLER_IMAGE)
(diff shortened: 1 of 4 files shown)


manifests/pod.yaml
```diff
@@ -64,9 +64,11 @@ spec:
        privileged: false
      resources:
        requests:
+         cpu: COREOS_ASSEMBLER_CPU_REQUEST
          memory: COREOS_ASSEMBLER_MEMORY_REQUEST
          devices.kubevirt.io/kvm: '1'
        limits:
+         cpu: '4'
```
dustymabe, line 71: We can bump the requests but I think we should leave the limit alone. If the cluster has excess CPU capacity then I don't think we should artificially limit ourselves, right?

dustymabe:
All of your `pod.replace` lines have a `*` in there that I don't think should be there.

manifests/pod.yaml
miabbott: I was reading https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/cpu-default-namespace/#what-if-you-specify-a-container-s-request-but-not-its-limit

...and was attempting to avoid a situation where the cluster/namespace defaults would arbitrarily limit the amount of CPUs we could request.  Though I cannot determine if there are default CPU limits in place.

Anyways...I'll drop the limit definition.

dustymabe: yeah, since we request 4 then we won't get scheduled unless we get 4. If the cluster has a default limit then we'll get that, but it will be at least >4 since that's what we requested.

dustymabe:
trivial LGTM

jobs/build.Jenkinsfile
```diff
@@ -86,6 +86,10 @@ def strict_build_param = is_mechanical ? "" : "--strict"
 // workflow, only the qemu image is built).
 def cosa_memory_request_mb = 6.5 * 1024 as Integer
 
+// the build pod runs most frequently and does the majority of the computation
+// so give it some healthy CPU shares
+pod = pod.replace("COREOS_ASSEMBLER_CPU_REQUEST", "4")
```
jlebon, line 91: I think we could be more generous there, but cool also to start with.

jlebon (approved):
Cool, thanks for writing this!

PR #567, opened 2022-07-18
refactor pipelines to use `cosa remote-session`; drop gangplank

dustymabe:
This includes quite a bit of re-organization of the build
job and then a wholesale replacement of the build-arch and
bump-lockfile jobs. It's probably best to review commits
individually up until you get to the build-arch update, then
it may be best to diff the build.Jenkinsfile and the
build-arch.Jenkinsfile directly.

I've done a fair amount of testing on this but there will
inevitably be followups where something small was missed.

Similarly there are a lot of (new) opportunities for
refactoring and/or improvements, which I've decided to leave
for followups in order not to take forever on this. For
example, a clear followup is now modifying the fcosKola in
coreos-ci-lib to be able to handle transparently being run
via `remote-session`.


cgwalters (commented):
Awesome!  I only read/reviewed up to https://github.com/coreos/fedora-coreos-pipeline/pull/567/commits/c20e7e548358f5b0d2d329a49e9f7b493f1fef9f

so far - i.e. all the prep patches.  They look good; do you think it's worth trying to land those separately?  I have no strong opinion myself.

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -69,25 +76,74 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
...
+            "quay.io/coreos-assembler/coreos-assembler:",
+            "localhost/coreos-assembler:"
+        )
+    }
+
+    // Initialize the sessions on the remote builders
+    stage("Initialize Remotes") {
+        parallel aarch64: {
+            pipeutils.withPodmanRemoteArchBuilder(arch: "aarch64") {
+                sessionaarch64 = shwrapCapture("cosa remote-session create --image ${image} --expiration 4h")
```
cgwalters, line 103: Let's hoist the `4h` to a handy constant somewhere? 

Hmmm more generally I think we can deduplicate more code across architectures.  But, not a blocker.

cgwalters (approved):

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -130,116 +186,228 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
...
+                archiveArtifacts allowEmptyArchive: true, artifacts: 'kola-testiso*${arch}.tar.xz'
+            }
+        }
+    }
+
+    // Destroy the remote sessions. We don't need them anymore
+    stage("Destroy Remotes") {
+        parallel aarch64: {
+            pipeutils.withExistingCOSARemoteSession(arch: "aarch64",
+                                                 session: sessionaarch64) {
```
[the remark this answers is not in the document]
dustymabe: I'll try to address this in a followup. I'm planning more re-org anyway.

jobs/bump-lockfile.Jenkinsfile
dustymabe: Yeah. I'll probably hoist the remote session create code into a shared function soon. Will address in a followup.

dustymabe:
> so far - i.e. all the prep patches. They look good; do you think it's worth trying to land those separately? I have no strong opinion myself.

I think it would be good to just land it all and work out any broken pieces. I've tested in staging but that doesn't quite 100% test everything.

jobs/build.Jenkinsfile
```diff
@@ -178,24 +178,28 @@ lock(resource: "build-${params.STREAM}") {
...
+                        parent_commit = commit_obj["checksum"]
+                        parent_version = release["version"]
+                        break
                     }
                 }
+            }
+        }
 
+        // buildfetch previous build info
+        stage('BuildFetch') {
```
jlebon, line 202: Bikeshed: "BuildFetch" looks weird... maybe just "Fetch Metadata" ?

jobs/build.Jenkinsfile
```diff
@@ -178,24 +178,28 @@ lock(resource: "build-${params.STREAM}") {
...
+                for (release in releases["releases"].reverse()) {
+                    def commit_obj = release["commits"].find{ commit -> commit["architecture"] == basearch }
+                    if (commit_obj != null) {
+                        parent_commit = commit_obj["checksum"]
+                        parent_version = release["version"]
+                        break
                     }
                 }
+            }
+        }
```
jlebon, line 199: Aside: I know I wrote this way back, but probably about time we factor that out into a script in cosa or fedora-coreos-releng-automation. Or maybe it could be a `cosa buildfetch --parent` which fetches the parent build info and creates a `parent` symlink in `builds/`.

jobs/build.Jenkinsfile
```diff
@@ -253,36 +253,36 @@ lock(resource: "build-${params.STREAM}") {
...
             cosa build ostree ${strict_build_param} --skip-prune ${force} ${version} ${parent_arg}
             """)
+
+            // Insert the parent info into meta.json so we can display it in
+            // the release browser and for sanity checking
+            if (parent_commit && parent_version) {
+                shwrap("""
+                cosa meta \
+                    --set fedora-coreos.parent-commit=${parent_commit} \
+                    --set fedora-coreos.parent-version=${parent_version}
```
jlebon, line 264: Nice!

jobs/build.Jenkinsfile
```diff
@@ -308,8 +308,9 @@ lock(resource: "build-${params.STREAM}") {
             shwrap("""
             cosa generate-hashlist --arch=${basearch} --release=${newBuildID} \
                 --output=builds/${newBuildID}/${basearch}/exp-hash.json
-            sha256sum builds/${newBuildID}/${basearch}/exp-hash.json \
-                > builds/${newBuildID}/${basearch}/exp-hash.json-CHECKSUM
+            source="builds/${newBuildID}/${basearch}/exp-hash.json"
+            target="builds/${newBuildID}/${basearch}/exp-hash.json-CHECKSUM"
+            cosa shell -- bash -c "sha256sum \$source > \$target"
```
jlebon, line 314: I can imagine this `cosa shell` trap being something we easily forget and regress on. An alternative is introduce a new e.g. `cosa_shwrap()` which is like `shwrap` but automatically runs on the remote host via `cosa shell` in the remote invocation case.

utils.groovy
```diff
@@ -82,4 +82,36 @@ def bump_builds_json(stream, buildid, arch, s3_stream_dir) {
...
+    }
+}
+
+// Run in a cosa remote session context on a builder of the given
+// architecture.
+//
+// Available parameters:
+// session:  string -- The session ID of the already created session
+//    arch:  string -- The architecture of the desired host
+def withExistingCOSARemoteSession(params = [:], Closure body) {
```
jlebon, line 105: Bikeshed: "existing" seems wordily redundant here with "with". Also maybe `s/COSA/Cosa/`?

utils.groovy
```diff
@@ -82,4 +82,36 @@ def bump_builds_json(stream, buildid, arch, s3_stream_dir) {
...
+    arch = params['arch']
+    session = params['session']
+    withPodmanRemote(remoteHost: "fcos-${arch}-builder-host-string",
+                     remoteUid:  "fcos-${arch}-builder-uid-string",
+                     sshKey:     "fcos-${arch}-builder-sshkey-key") {
+        withEnv(["COREOS_ASSEMBLER_REMOTE_SESSION=${session}"]) {
+            body()
+        }
+    }
```
jlebon, line 114: Isn't this

```groovy
session = params['session']
withPodmanRemoteArchBuilder(params) {
    withEnv(["COREOS_ASSEMBLER_REMOTE_SESSION=${session}"]) {
        body()
    }
}
```

?

jobs/build-arch.Jenkinsfile
```diff
@@ -1,14 +1,13 @@
 import org.yaml.snakeyaml.Yaml;
 
-def pipeutils, streams, official, repo, gp
+def pipeutils, streams, official, uploading
 def src_config_url, src_config_ref, s3_bucket
+def session
```
jlebon, line 5: Does this need to be global? Can we declare it where we define it instead?

jobs/build-arch.Jenkinsfile
```diff
@@ -245,200 +302,244 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}", extra: [[resource: "rele
...
-    - cosa buildextend-metal4k
-    - cosa buildextend-live
-    - kola testiso -S --output-dir tmp/kola-metal
-    - cosa buildextend-ibmcloud
-  post_commands:
-    - cosa compress --compressor xz
-    - tar --xz -cf tmp/kola.tar.xz tmp/kola{-basic,,-metal,-metal4k} || true
-  post_always: true
-delay_meta_merge: false
-EOF
```
jlebon, line 366: :tada:

jobs/build-arch.Jenkinsfile
jlebon, line 5: In the commit message introducing this hunk: `s/no/now/`

jlebon (approved):
A lot of nits, but looks great overall! This is going to be much nicer to work with! I'm OK addressing everything in a follow-up (well, the commit message typo of course can't :) ).

jobs/bump-lockfile.Jenkinsfile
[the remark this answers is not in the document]
cgwalters: Yep totally fine by me!

jobs/build.Jenkinsfile
dustymabe: > Aside: I know I wrote this way back, but probably about time we factor that out into a script in cosa or fedora-coreos-releng-automation.

I was actually looking at `fedora-coreos-releng-automation`. I almost think the scripts there that we only use in the pipeline we should just pull into COSA itself rather than `git clone`ing that repo for every pipeline run.

Or maybe it could be a `cosa buildfetch --parent` which fetches the parent build info and creates a `parent` symlink in `builds/`.

Yeah I think I prefer this ^^

jobs/build.Jenkinsfile
dustymabe: i.e. we easily forget when adding new code to do the right thing and appropriately use `cosa shell -- cmd` when needed? That is a valid concern. I'm not in love with the idea of a `cosa_shwrap()` as I don't really think it alleviates the concern; we can still regress easily. I think it's just something we need to be aware of.. and possibly more aggressively refactor and deduplicate similar code in the future.

utils.groovy
dustymabe: I'm toying with the idea of adding a `withCOSARemoteSession()` in a followup that will create and destroy the remote session with a `body()` in between. This function will still have it's uses though so it might be good to keep the `Existing` here in `withExistingCOSARemoteSession()`.

I'm +1 to make it `Cosa` instead of `COSA`.

jobs/build.Jenkinsfile
dustymabe: Because I'm intimately familiar with the code here "BuildFetch" is more meaningful to me because it runs a `cosa buildfetch` and I know what that does.

I agree that to untrained eyes "Fetch Metadata" might be more obvious.

I'm ok changing it if you want to in a followup.

jobs/build-arch.Jenkinsfile
dustymabe: I think you are right. Will fixe

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -1,9 +1,9 @@
-def pipeutils, streams, official, gp
+def pipeutils, streams, official
+def sessionaarch64, sessions390x
```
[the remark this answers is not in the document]
dustymabe: fixed

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -130,116 +184,228 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
...
             } finally {
-                shwrap("tar -cf - tmp/kola-testiso-metal/ | xz -c9 > ${env.WORKSPACE}/kola-testiso-metal.tar.xz")
-                shwrap("tar -cf - tmp/kola-testiso-metal4k/ | xz -c9 > ${env.WORKSPACE}/kola-testiso-metal4k.tar.xz")
-                shwrap("tar -cf - tmp/kola-testiso-uefi/ | xz -c9 > ${env.WORKSPACE}/kola-testiso-uefi.tar.xz")
-                archiveArtifacts allowEmptyArchive: true, artifacts: 'kola-testiso*.tar.xz'
+                shwrap("""
+                cosa shell -- tar -c --xz tmp/kola-testiso-metal/ > kola-testiso-metal.${arch}.tar.xz
+                cosa shell -- tar -c --xz tmp/kola-testiso-metal4k/ > kola-testiso-metal4k.${arch}.tar.xz
+                cosa shell -- tar -c --xz tmp/kola-testiso-uefi/ > kola-testiso-uefi.${arch}x86_64.tar.xz
+				""")
```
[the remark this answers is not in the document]
dustymabe: removed

jobs/build-arch.Jenkinsfile
dustymabe: fixed

utils.groovy
dustymabe: yes. it can be simplified that way.

dustymabe:
ok new upload should have taken care of most of code review comments

jlebon (approved):

PR #569, opened 2022-07-20
jobs/build-arch: take build-specific release lock first

jlebon:
We need to take the release lock first before the build lock so that
even if a build for the given stream is pending, we still make sure the
release job waits until we finish.

Add some details about the theoretical race condition that exists while
we're here.

Reported-by: Dusty Mabe <dusty@dustymabe.com>


dustymabe (approved):
LGTM

PR #570, opened 2022-07-20
utils: add missing `def`

jlebon:
Otherwise it's a global binding.

I think this is what's causing what appears to be races where one arch
seem like it's using another arch's session in `bump-lockfile`.


dustymabe (approved):
LGTM

PR #572, opened 2022-07-22
jobs/{build,build-arch}: don't run metal4k tests on s390x

dustymabe:
Also rework a bit the structure of the "parallel" invocation
for testiso.

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -482,17 +482,24 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}", extra: [[resource: "rele
                 cosa compress --compressor xz --artifact metal --artifact metal4k
                 """)
                 try {
-                    parallel metal: {
+                    parallelruns = [:]
+                    parallelruns['metal'] = {
                         shwrap("cosa kola testiso -S --output-dir tmp/kola-testiso-metal")
-                    }, metal4k: {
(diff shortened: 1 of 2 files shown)


jobs/build-arch.Jenkinsfile
```diff
@@ -482,17 +482,23 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}", extra: [[resource: "rele
...
                 """)
                 try {
-                    parallel metal: {
+                    parallelruns = [:]
+                    parallelruns['metal'] = {
                         shwrap("cosa kola testiso -S --output-dir tmp/kola-testiso-metal")
-                    }, metal4k: {
-                        shwrap("cosa kola testiso -SP --qemu-native-4k --qemu-multipath --output-dir tmp/kola-testiso-metal4k")
                     }
+                    if (basearch != "s390x") {
```
cgwalters, line 489: LGTM as is but: I think the right long term direction is to encode this knowledge in coreos-assembler.  (It's what gangplank was also trying to do to a degree)

cgwalters (approved):

jobs/build-arch.Jenkinsfile
dustymabe: yeah I think we reached out to someone recently to find out if we should stop building the 4k metal images on s390x at all.

Tracking in https://github.com/coreos/fedora-coreos-tracker/issues/1261 - which I should probably also link to here in the code. Let add that.

PR #578, opened 2022-07-23
jobs/bump-lockfile: use locally scoped arch var

dustymabe:
I was seeing some issues where the `arch` variable was getting
evalated lazily and all three of my parallel runs were all executing
against the same architecture. Adding a locally scoped arch variable
seems to fix it.


cgwalters (approved):
Hmm.  Groovy may have a semantic here similar to https://go.dev/doc/faq#closures_and_goroutines

PR #579, opened 2022-07-25
jobs/build: Set Force=true for forked multi-arch builds

dustymabe:
We've seen it happen a few times recently where the x86_64 build
detects a change and does a build but the multi-arch builds don't
detect a change and report "No apparent changes since previous commit".

This is probably because of our COSA builds being out of sync, which
we'll fix soon, but either way we probably want to set Force=true
for our multi-arch builds.

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -416,8 +416,11 @@ lock(resource: "build-${params.STREAM}") {
         stage('Fork Multi-Arch Builds') {
             if (uploading) {
                 for (arch in params.ADDITIONAL_ARCHES.split()) {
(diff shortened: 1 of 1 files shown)


jlebon (approved):
Makes sense to me!

cgwalters:
(Eventually we want it the other way around too right?  i.e. an aarch64-only change also needs to force x86_64)

dustymabe:
> (Eventually we want it the other way around too right? i.e. an aarch64-only change also needs to force x86_64)

Yes. That's a gap right now. The good news is that enough changes are coming in all the time that even if we starve a change in a non x86_64 pipeline today more changes will come in soon enough and the new run will pick up the changes too.

PR #582, opened 2022-07-29
Add support for additional root CA certs

jlebon:
We're working on getting this pipeline working internally. The first
thing we hit is that Red Hat has its own internal root CA.

We'll need to handle this both for cosa and the Jenkins agent which
wants to fetch `remoting.jar` from the controller as soon as it starts.

This patch handles just the latter for now. We do this by optionally
supporting building our own Jenkins agent with the certificates in it.

For cosa, I think we will be able to work around this by directly
pointing e.g. git and libdnf at the certificate.

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -1,6 +1,6 @@
 import org.yaml.snakeyaml.Yaml;
 
-def pipeutils, streams, official, uploading
+def pipeutils, streams, official, uploading, jenkins_agent_image_tag
 def src_config_url, src_config_ref, s3_bucket
 def gcp_gs_bucket
 node {
@@ -14,6 +14,7 @@ node {
     src_config_ref = pipecfg['source-config-ref']
     s3_bucket = pipecfg['s3-bucket']
     gcp_gs_bucket = pipecfg['gcp-gs-bucket']
+    jenkins_agent_image_tag = pipecfg['jenkins-agent-image-tag']
 
     official = pipeutils.isOfficial()
     if (official) {
@@ -95,6 +96,7 @@ pod = pod.replace("COREOS_ASSEMBLER_CPU_LIMIT", "4")
 // substitute the right COSA image and mem request into the pod definition before spawning it
 pod = pod.replace("COREOS_ASSEMBLER_MEMORY_REQUEST", "${cosa_memory_request_mb}Mi")
 pod = pod.replace("COREOS_ASSEMBLER_IMAGE", params.COREOS_ASSEMBLER_IMAGE)
+pod = pod.replace("JENKINS_AGENT_IMAGE_TAG", jenkins_agent_image_tag)
 
 def podYaml = readYaml(text: pod);
 
--- a/manifests/jenkins-agent.yaml
+++ b/manifests/jenkins-agent.yaml
@@ -1,0 +1,36 @@
+apiVersion: v1
+kind: Template
+metadata:
+  name: jenkins-agent
+parameters:
+  - description: Secret name containing any additional root CA certificates
+    name: ADDITIONAL_ROOT_CA_CERTS_SECRET
+    value: additional-root-ca-certs
+objects:
+  - kind: BuildConfig
+    apiVersion: v1
+    metadata:
+      name: jenkins-agent
+    spec:
+      source:
+        dockerfile: |
+          FROM overridden
+          COPY certs/* /etc/pki/ca-trust/source/anchors/
+          RUN update-ca-trust
+        secrets:
+          - destinationDir: certs
+            secret:
+              name: ${ADDITIONAL_ROOT_CA_CERTS_SECRET}
+      strategy:
+        dockerStrategy:
+          from:
+            kind: ImageStreamTag
+            name: jenkins-agent-base:latest
+            namespace: openshift
+          forcePull: true
+      output:
+        to:
+          kind: ImageStreamTag
+          name: jenkins-agent:derived
+      successfulBuildsHistoryLimit: 2
+      failedBuildsHistoryLimit: 2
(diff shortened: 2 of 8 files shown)


jlebon:
Of course, this does not work because the pod runs fully unprivileged. Reworking this now to just support rebuilding the jnlp image.

jlebon:
> For cosa, I think we will be able to work around this by directly
> pointing e.g. git and libdnf at the certificate.

Basically, I want the API for "additional CA certs" to just be creating the secret and passing `--additional-root-ca-certs-secret` to `deploy`. Which is why the naming around this is generic and not e.g. "jenkins-agent-ca-certs".

saqibali-2k:
Getting `BuildConfig in version "v1" cannot be handled as a BuildConfig: json: cannot unmarshal string into Go struct field SecretBuildSource.spec.source.secrets.secret of type v1.LocalObjectReference` when running the script with the new argument.

manifests/jenkins-agent.yaml
```diff
@@ -0,0 +1,40 @@
...
+      name: jenkins-agent
+    spec:
+      source:
+        dockerfile: |
+          FROM overridden
+          COPY certs/* /etc/pki/ca-trust/source/anchors/
+          RUN update-ca-trust
+        secrets:
+          - destinationDir: certs
+            secret: additional-root-ca-certs
```
saqibali-2k, line 26: I think this needs to be changed to:
```yaml
        secrets:
          - secret:
              name: additional-root-ca-certs
            destinationDir: certs
```

jlebon: Yes, looking at https://docs.openshift.com/container-platform/4.10/rest_api/workloads_apis/buildconfig-build-openshift-io-v1.html#spec-source-secrets, you're right. Fixed!

jobs/build.Jenkinsfile
```diff
@@ -94,6 +95,7 @@ pod = pod.replace("COREOS_ASSEMBLER_CPU_REQUEST", "4")
 // substitute the right COSA image and mem request into the pod definition before spawning it
 pod = pod.replace("COREOS_ASSEMBLER_MEMORY_REQUEST", "${cosa_memory_request_mb}Mi")
 pod = pod.replace("COREOS_ASSEMBLER_IMAGE", params.COREOS_ASSEMBLER_IMAGE)
+pod = pod.replace("JENKINS_AGENT_IMAGE", params.JENKINS_AGENT_IMAGE)
```
saqibali-2k, line 98: Do we mean to use `jenkins_agent_image`? `params.JENKINS_AGENT_IMAGE` is `null`

jlebon: Fixed!

jlebon: OK, let's switch to a higher bandwidth channel to debug this.

cgwalters (commented):
Not super familiar with this code, but looks generally sane.

One thing though; any reason not to use https://docs.openshift.com/container-platform/4.10/security/certificates/updating-ca-bundle.html at the cluster level?

jlebon:
> One thing though; any reason not to use [docs.openshift.com/container-platform/4.10/security/certificates/updating-ca-bundle.html](https://docs.openshift.com/container-platform/4.10/security/certificates/updating-ca-bundle.html) at the cluster level?

Thanks for pointing me at this. I'm trying to understand its scope. It says:

> This bundle is merged with the Red Hat Enterprise Linux CoreOS (RHCOS) trust bundle and injected into the trust store of platform components that make egress HTTPS calls.

The reference of "platform components" here makes me think this is something OCP components opt into?

OK yup, digging into the mentioned example `image-registry-operator`, I see:
- https://github.com/openshift/cluster-image-registry-operator/blob/39a9ada4d4a3519a7e078d0f29263cccac32b954/manifests/07-operator-ibm-cloud-managed.yaml#L79-L85
- https://github.com/openshift/cluster-image-registry-operator/blob/6e88375a583645f65179836027954021eb5fdd30/images/bin/entrypoint.sh

Hmm, though from that `entrypoint.sh`, it's copying into `/etc/pki/ca-trust/` so I think that code runs privileged, which is not the case for the Jenkins agent container. But we could submit an RFE to have [the one `curl` operation we need](https://github.com/openshift/jenkins/blob/8b75b11f7dbcaefbed6161965c17da4d13c69ab2/slave-base/contrib/bin/run-jnlp-client#L131) point to use the mounted bundle if available.

Another approach, which is what I think we were doing internally too is to create the bundle ourselves. Then we could directly mount it at its `/etc` location and avoid having to rebuild the agent. The downside is we then have to manage the bundle.

Anyway, I think for now let's get this in. There's definitely more follow-ups coming related to this, and we might have to revisit the approach as a result.

dustymabe (approved):
LGTM

manifests/pod.yaml
```diff
@@ -12,7 +12,7 @@ spec:
   serviceAccountName: jenkins
   containers:
    - name: jnlp
-     image: quay.io/openshift/origin-jenkins-agent-base:latest
+     image: JENKINS_AGENT_IMAGE
```
jlebon, line 15: I forgot to mention this, but this actually also fixes a mismatch issue. Before we were using an unversioned agent image from Quay. Now, we're following the in-cluster imagestream, just like for the controller itself, so we know they're always matching.

jlebon:
Now tested working in stage!

jlebon:
The main change in the latest push is that we now *always* define a `jenkins-agent` ImageStream (I've moved its definition to `jenkins-s2i.yaml`). The BuildConfig though is still only defined if additional root CA certs are necessary. The `base` tag follows the built-in OpenShift `jenkins-agent-base` ImageStream and the `derived` tag is present if we're deriving. The configmap now just holds whether the tag to use is `base` or `derived`.

dustymabe (approved):

PR #584, opened 2022-08-01
convert build/build-arch jobs to use cosaPod

dustymabe:
With the recent changes in https://github.com/coreos/coreos-assembler/pull/2975
we now need to set a cpu request limit. cosaPod() does this for us
now that https://github.com/coreos/coreos-ci-lib/pull/112 is merged.
Let's convert our jobs to use cosaPod().


jlebon:
I think this is definitely the goal. The primary problem is secrets. Right now they're defined in `pod.yaml` but we have to move over to using Jenkins credentials and then we'd just use `withCredentials`.

jlebon:
OK wow, can't believe I missed this before: https://plugins.jenkins.io/kubernetes-credentials-provider/. This plugin can automatically import Kubernetes secrets into Jenkins optionally filtered by a specific label, which means secrets don't need to be individually mounted into the Jenkins pod either.

jlebon:
> OK wow, can't believe I missed this before: [plugins.jenkins.io/kubernetes-credentials-provider](https://plugins.jenkins.io/kubernetes-credentials-provider/). This plugin can automatically import Kubernetes secrets into Jenkins optionally filtered by a specific label, which means secrets don't need to be individually mounted into the Jenkins pod either.

Added in https://github.com/coreos/fedora-coreos-pipeline/pull/607.

jlebon:
https://github.com/coreos/fedora-coreos-pipeline/pull/671 should unblock this!

dustymabe:
ok I updated this PR. I should probably run it through a staging run before deploying.

jlebon (approved):
Nice!

PR #585, opened 2022-08-01
jobs/{build,build-arch,release}: set cpu limit

dustymabe:
With the recent changes in https://github.com/coreos/coreos-assembler/pull/2975
we now need to set a cpu request limit.


bgilbert (approved):
LGTM

PR #583, opened 2022-07-29
Revert "jenkins/config: bump kubernetes-plugin maxRequestsPerHost"

jlebon:
This reverts commit fc3532704a0c3082810d34ca461d52747a53150a.

The intention is good, but we can't do this via JCASC because it
clobbers the existing `openshift` cloud configuration instead of merging
it in.

PR #587, opened 2022-08-03
manifests/jenkins.yaml: set OVERRIDE_PV_CONFIG_WITH_IMAGE_CONFIG

jlebon:
By default, when using a PVC, the OpenShift S2I Jenkins image only
instantiates templates on the first run. This however means that
configuration state can accumulate in the PVC.

Set the `OVERRIDE_PV_CONFIG_WITH_IMAGE_CONFIG` env var which tells the
S2I run script to always instantiate and copy in templates. This
*should* preserve build logs.

An example of this is the Kubernetes cloud config, which was lost in a
recent change (see #583).


dustymabe (approved):

dustymabe:
LGTM

PR #591, opened 2022-08-05
Revert "manifests/jenkins.yaml: set OVERRIDE_PV_CONFIG_WITH_IMAGE_CONFIG"

jlebon:
This reverts commit 086c56ea2320d0f711ba5f32ed55caa9294ed1e9.

This helped us recover access to the Kubernetes API server, but it's not
a great default because it means we lose everything anytime Jenkins is
restarted for any reason (e.g. pipeline or cluster updates). We do want
to reprovision from scratch more often, but we need to first move off of
the Jenkins SCM webhook handling, which is not friendly to frequent
reprovisioning.


dustymabe (approved):
LGTM

PR #593, opened 2022-08-05
jobs/build: request as many CPUs as parallel VMs and run reprovision tests serially

jlebon:
```
commit c9fc772f7c17d77b94b6254cbab0888d63c7e736
Date:   Thu Aug 4 19:23:35 2022 -0400

    jobs/build: request as many CPUs as parallel VMs

    Right now we're running with `--parallel 5` and run the upgrade tests in
    parallel, but only have 4 CPUs. This causes some heavier tests to fail.
    Let's request as many CPUs as we run VM tests in parallel. Since the
    latter is calculated from the memory request, that's what we do.

    In effect, this bumps up our CPU requests to 6.
```
```
commit 04cb8c388688cfb4a11f1db1381aca2a37998df6
Author: Jonathan Lebon <jonathan@jlebon.com>
Date:   Fri Aug 5 11:21:12 2022 -0400

    jobs/build: run reprovision tests not in parallel

    Those tests take 4G of RAM each and are heavy in I/O. Run them serially
    to stay within our declared memory request and minimize the chances of
    timeouts.
```

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -88,10 +88,16 @@ def strict_build_param = is_mechanical ? "" : "--strict"
 // workflow, only the qemu image is built).
 def cosa_memory_request_mb = 6.5 * 1024 as Integer
 
+// Now that we've established the memory constraint based on xz above, derive
+// kola parallelism from that. We leave 512M for overhead and VMs are 1G each
+// (in general; root reprovisioning tests require 4G which is partly why we run
+// them separately).
+def ncpus = ((cosa_memory_request_mb - 512) / 1024) as Integer
+
 // the build pod runs most frequently and does the majority of the computation
 // so give it some healthy CPU shares
-pod = pod.replace("COREOS_ASSEMBLER_CPU_REQUEST", "4")
-pod = pod.replace("COREOS_ASSEMBLER_CPU_LIMIT", "4")
+pod = pod.replace("COREOS_ASSEMBLER_CPU_REQUEST", "${ncpus}")
+pod = pod.replace("COREOS_ASSEMBLER_CPU_LIMIT", "${ncpus}")
 
 // substitute the right COSA image and mem request into the pod definition before spawning it
 pod = pod.replace("COREOS_ASSEMBLER_MEMORY_REQUEST", "${cosa_memory_request_mb}Mi")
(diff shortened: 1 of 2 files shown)


jlebon:
The background for this is that we're working on moving the RHCOS pipeline to a new cluster (and basing it on the FCOS pipeline), and the reprovisioning tests there are much flakier (slower/more contended disk I/O?).

I think eventually we should move our x86_64 build to also use `podman --remote` to be consistent with the other arches. We can then use a VM with actual IOPS guarantees, which Kubernetes can't offer us AFAIK.

jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/3028

jobs/build.Jenkinsfile
```diff
@@ -88,10 +88,13 @@ def strict_build_param = is_mechanical ? "" : "--strict"
 // workflow, only the qemu image is built).
 def cosa_memory_request_mb = 6.5 * 1024 as Integer
 
+// leave 512M for overhead; VMs are 1G each
+def ncpus = ((cosa_memory_request_mb - 512) / 1024) as Integer
+
 // the build pod runs most frequently and does the majority of the computation
 // so give it some healthy CPU shares
```
dustymabe, line 95: maybe let's add a better comment here on why we're basing CPU requests on memory (which seems odd to me). basically what we're doing here is saying our limiting factor is memory and our VMs are approximately 1G memory per VM so request that many CPUs.

A more direct way of doing this may be defining first here how "parallel" we want our kola tests to run and then deriving both CPU and memory from that.

i.e. if we want a parallel of 6 (5 kola + 1 kola upgrade) then we define parallel of 6 and we need 6*1G+512M of memory and 6 cpus.

jlebon: Though technically currently we're defining the amount of memory we need based on xz, not based on kola parallelism. The latter is being derived from that. I've added a comment to expand on that.

I think we could flip it around and say kola tests is our most memory intensive factor though it's mostly playing with words unless we actually want to go past 6.

dustymabe (approved):
LGTM

jlebon:
Let's wait until releases are done before merging this.

jlebon:
We're past the critical point in the releases. Merging.

PR #596, opened 2022-08-08
jobs/build-cosa: add `FORCE` parameter

jlebon:
We need to be able to rebuild the same cosa commit e.g. if a package
changed. (This would be a nice benefit of using lockfiles for cosa too.)


dustymabe (approved):
LGTM

PR #598, opened 2022-08-10
jobs/build-cosa: push in v2s2 format

jlebon:
The cosa image we build is used to trigger internal RHCOS multi-arch
rebuilds (and in the near future will be used directly once we have all
the arches built). But the stack on the target cluster is too old to
understand OCI, so we get:

    Internal error occurred: missing signature key

Add `--v2s2` to push in v2s2 format.

Once we've rebased RHCOS to the FCOS pipeline, we should be able to
revert this.

Related: https://github.com/coreos/coreos-assembler/pull/2726

### DIFF
--- a/jobs/build-cosa.Jenkinsfile
+++ b/jobs/build-cosa.Jenkinsfile
@@ -118,7 +118,7 @@ try {
                     images += " --image=docker://${params.CONTAINER_REGISTRY_STAGING_REPO}:${arch}-${shortcommit}"
                 }
                 shwrap("""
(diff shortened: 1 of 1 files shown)


travier:
Does this mean that we will have to "force push" a tag for all arches & versions?

jlebon:
> Does this mean that we will have to "force push" a tag for all arches & versions?

Yeah, we'll have to rerun it once per version. It should be able to avoid rebuilding the images I think.

dustymabe (approved):
LGTM

PR #599, opened 2022-08-10
Start adding Brew necessary secrets and configs

ravanelli:


manifests/pipeline.yaml
```diff
@@ -35,7 +35,13 @@ parameters:
     value: quay.io/coreos-assembler/coreos-assembler:main
   - description: AWS S3 bucket in which to store builds (or blank for none)
     name: S3_BUCKET
-    value: fcos-builds
+    value: ''
```
jlebon, line 38: Was this intended?

manifests/pipeline.yaml
```diff
@@ -35,7 +35,13 @@ parameters:
...
+  - description: AWS S3 principle to be used (or blank for none)
+    name: S3_PRINCIPLE
+    value: ''
+  - description: AWS S3 profile to be used (or blank for none)
+    name: S3_PROFILE
+    value: ''
```
jlebon, line 44: Shouldn't those be e.g. `BREW_PRINCIPAL` and `BREW_PROFILE`?

manifests/pod.yaml
```diff
@@ -83,6 +95,23 @@ spec:
     secret:
       secretName: aws-fcos-kola-bot-config
       optional: true
+  # Necessary secrets for Brew
```
jlebon, line 98: For secrets, we've been trying to go in the reverse direction actually and drain the secrets from there. The latest in that area is https://github.com/coreos/fedora-coreos-pipeline/pull/607. You can see an example of how it can be used in https://github.com/coreos/fedora-coreos-pipeline/pull/614. So concretely, the idea is:
- you create the secrets in OpenShift in a format compatible with kubernetes-credentials-provider (see examples here: https://github.com/jenkinsci/kubernetes-credentials-provider-plugin/tree/master/docs/examples)
- they get automatically synced as Jenkins credentials
- in jobs which have to tolerate them not existing, you use `tryWithCredentials()`
- in jobs which can assume they always exist, you use `withCredentials()` (like I think the new dedicated Brew job would)

manifests/pipeline.yaml
ravanelli: Yes, you have to submit it.  The S3 bucket will be different for fcos/rhcos

manifests/pipeline.yaml
ravanelli: Yeah, fixing

manifests/pod.yaml
ravanelli: Great! I will remove this part here so. And add necessary stuff in  https://github.com/coreos/fedora-coreos-pipeline/pull/602

ravanelli:
Closing in favor of https://github.com/coreos/fedora-coreos-pipeline/pull/836

PR #602, opened 2022-08-10
jobs: Add Brew upload job

ravanelli:
- Create a separate job to upload the metadata
- We are not required to upload it for every single build as we
did before. Only ART will be responsible for it, we are required to
make Brew uploads for GAs or images available to customers.


ravanelli:
I still need to test it with the new FCOS/RHCOS CI

PR #606, opened 2022-08-15
jobs/bump-lockfile: add some updates from build job

dustymabe:
Let's give the cosaPod more CPUs by default. It changed recently
to limiting the job to just 2 by default and we need more than that
to cut down on flakes.

Also, let's stitch in the separation of reprovision tests so that
they run serially.

### DIFF
--- a/jobs/bump-lockfile.Jenkinsfile
+++ b/jobs/bump-lockfile.Jenkinsfile
@@ -51,8 +51,13 @@ def getLockfileInfo(lockfile) {
     return [pkgChecksum, pkgTimestamp]
 }
 
+// Keep in sync with build.Jenkinsfile
+def cosa_memory_request_mb = 6.5 * 1024 as Integer
+def ncpus = ((cosa_memory_request_mb - 512) / 1024) as Integer
+
 try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTES') { 
-    cosaPod(image: params.COREOS_ASSEMBLER_IMAGE) {
+    cosaPod(image: params.COREOS_ASSEMBLER_IMAGE
+            cpu: "${ncpus}", memory: "${cosa_memory_request_mb}Mi") {
     currentBuild.description = "[${params.STREAM}] Running"
 
     // set up git user upfront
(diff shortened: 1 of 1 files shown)


jobs/bump-lockfile.Jenkinsfile
```diff
@@ -202,14 +206,23 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
                     }
                     parallelruns["${arch}:Kola"] = {
                         shwrap("""
-                        cosa kola run --rerun --parallel 5 --no-test-exit-error
+                        cosa kola run --rerun --parallel 5 --no-test-exit-error --tag '!reprovision'
```
jlebon, line 209: I think we want

```diff
@@ -201,8 +210,9 @@ try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTE
                         }
                     }
                     parallelruns["${arch}:Kola"] = {
+                        def n = ncpus - 1 // remove 1 for upgrade test
                         shwrap("""
-                        cosa kola run --rerun --parallel 5 --no-test-exit-error
+                        cosa kola run --rerun --parallel ${n} --no-test-exit-error --tag '!reprovision'
                         cosa shell -- tar -c --xz tmp/kola/ > kola-run.${arch}.tar.xz
                         cosa shell -- cat tmp/kola/reports/report.json > report-kola.${arch}.json
                         """)
```

jlebon:
Eventually though, we need to move these jobs to use the `kola` step from coreos-ci-lib instead. We can then hardcode the reprovision stuff there only.

jobs/bump-lockfile.Jenkinsfile
dustymabe: since this runs on all architectures the plan was to leave it hardcoded at `5`. I guess I could do something more fancy here or just assume the value for `ncpus` we are using is valid across all architectures.

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -51,8 +51,12 @@ def getLockfileInfo(lockfile) {
     return [pkgChecksum, pkgTimestamp]
 }
 
+// Keep in sync with build.Jenkinsfile
+def cosa_memory_request_mb = 6.5 * 1024 as Integer
+def ncpus = ((cosa_memory_request_mb - 512) / 1024) as Integer
+
 try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTES') { 
-    cosaPod(image: params.COREOS_ASSEMBLER_IMAGE) {
+    cosaPod(cpu: ncpus, image: params.COREOS_ASSEMBLER_IMAGE) {
```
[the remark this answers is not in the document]
dustymabe: Will fix

dustymabe:
> Eventually though, we need to move these jobs to use the `kola` step from coreos-ci-lib instead. We can then hardcode the reprovision stuff there only.

+1 to refactor. Let's discuss this more soon.

jobs/bump-lockfile.Jenkinsfile
jlebon: Yeah, this is what I had locally:

```diff
diff --git a/jobs/bump-lockfile.Jenkinsfile b/jobs/bump-lockfile.Jenkinsfile
index b744e73..df94e83 100644
--- a/jobs/bump-lockfile.Jenkinsfile
+++ b/jobs/bump-lockfile.Jenkinsfile
@@ -51,8 +51,17 @@ def getLockfileInfo(lockfile) {
     return [pkgChecksum, pkgTimestamp]
 }

-try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTES') {
-    cosaPod(image: params.COREOS_ASSEMBLER_IMAGE) {
+// On x86_64, we run up to 6 VMs in parallel (--parallel 5 + upgrade test).
+// Note this also controls parallelism on !x86_64 where tests run outside any
+// cgroups control.
+def ncpus = 6
+
+// Each VM is 1G (in general; root reprovisioning tests require 4G which is
+// partly why we run them separately). Add 512M for overhead.
+def memory = (ncpus * 1024) + 512
+
+try { lock(resource: "bump-${params.STREAM}") { timeout(time: 120, unit: 'MINUTES') {
+    cosaPod(image: params.COREOS_ASSEMBLER_IMAGE, memory: "${memory}Gi", cpu: "${ncpus}") {
     currentBuild.description = "[${params.STREAM}] Running"

     // set up git user upfront
```

dustymabe: also.. one thing that I learned recently is that the default memory for VMs on x86_64 is 1G while for other architectures it is 2G.

jlebon (approved):

PR #607, opened 2022-08-15
plugins: add kubernetes-credentials-provider

jlebon:
This plugin can automatically import Kubernetes secrets with a certain
label into Jenkins. This will allow us to drop secret definitions
defined in a bunch of places (Jenkins template, JCASC dropins, and cosa
pod template).


jlebon:
I've tested this plugin manually in the staging pipeline (e.g. creating a secret and adding the label immediately created the Jenkins secret!). I have work in a follow-up that will make use of this.

dustymabe (approved):
LGTM

PR #608, opened 2022-08-15
don't run kola basic tests twice

dustymabe:
Since we have a kola basic tests stage where we separately run the
basic tests, let's denylist them when we run the rest of the kola
tests.


jlebon:
Don't we also need it in the `--tag reprovision` case?

dustymabe:
> Don't we also need it in the `--tag reprovision` case?

Does the `basic` test have the `reprovision` tag? Maybe there's something I don't understand.

jlebon (approved):
Oh right, sorry makes sense!

PR #609, opened 2022-08-15
enable ppc64le in streams.groovy

dustymabe:
Things should be in place now for this to start working. Let's find
what's broken!


dustymabe:
requires: https://github.com/coreos/coreos-assembler/pull/3033

jlebon (approved):
:tada:

PR #611, opened 2022-08-16
jobs/bump-lockfile: make arch list based on streams.additional_arches

dustymabe:
This way we don't need to maintain two separate lists.


jlebon (approved):
LGTM, though feels like we should formalize this more. Maybe we can have an `all_arches` in `streams.groovy` which is `x86_64` + `additional_arches`?

dustymabe:
> LGTM, though feels like we should formalize this more. Maybe we can have an `all_arches` in `streams.groovy` which is `x86_64` + `additional_arches`?

Yeah. I'll make a followup for that.

PR #612, opened 2022-08-16
manifests/sleep.yaml: drop AWS secrets from pod, add KVM, and an `/srv` emptyDir

jlebon:
See individual commits.


dustymabe:
Are we able to drop the AWS secrets from the pod without doing something else to make sure the secret is mapped in?

jlebon:
This is removing it from `sleep.yaml`, not `pod.yaml`. Hoping to start draining the latter soon.

dustymabe (approved):

PR #613, opened 2022-08-16
utils: add `tryWithCredentials` function

jlebon:
This new helper will be useful together with the new
`kubernetes-credentials-provider` plugin to only run a block of code if
a given Jenkins credential exists.


dustymabe (approved):
Nice! LGTM

PR #614, opened 2022-08-16
Rely on new kubernetes-credentials-provider plugin for root CA cert and inject it into cosa container

jlebon:
Roughly, this series documents how to create a secret for the root CA usable with kubernetes-credentials-provider, and then uses the auto-created secret in the pipeline to optionally have it added to the cosa container (relying on https://github.com/coreos/coreos-assembler/pull/3032).

See individual commit for more details.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -295,6 +295,31 @@ echo $coreosbot_token > token
 oc create secret generic github-coreosbot-token --from-file=token
 ```
 
+### [PROD, OPTIONAL] Create additional root CA certificate secret
+
+If an additional root CA certificate is needed, create it as
+a secret:
+
+```
+apiVersion: v1
+kind: Secret
+metadata:
+  name: additional-root-ca-cert
+  labels:
+    jenkins.io/credentials-type: "secretFile"
+  annotations:
+    jenkins.io/credentials-description: "Root certificate for XXX"
+type: Opaque
+stringData:
+  filename: ca.crt
+data:
+  data: |
+    # output of $(base64 < ca.crt)
+```
+
+The description can be customized. All other fields must be
+exactly as is.
+
 ### Create a Jenkins instance with a persistent volume backing store
 
 ```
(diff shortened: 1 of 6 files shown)


jlebon:
Requires: https://github.com/coreos/fedora-coreos-pipeline/pull/613

jobs/build.Jenkinsfile
```diff
@@ -137,6 +137,14 @@ lock(resource: "build-${params.STREAM}") {
         cat /cosa/coreos-assembler-git.json
         """)
 
+        // add any additional root CA cert before we do anything that fetches
+        pipeutils.tryWithCredentials([file(credentialsId: 'additional-root-ca-cert', variable: 'ROOT_CA')]) {
+            shwrap('''
+                cp $ROOT_CA /etc/pki/ca-trust/source/anchors/
+                /usr/lib/coreos-assembler/update-ca-trust-unpriv
+            ''')
+        }
```
dustymabe, line 146: Is this only needed for the `build` job?

dustymabe (approved):
LGTM - a few comments

jobs/build.Jenkinsfile
jlebon: Good catch, it was needed in the `bump-lockfile` job too! I made it into a helper function. It's tempting to put some of this stuff into coreos-ci-lib. Probably makes sense to do that eventually.

HACKING.md
```diff
@@ -360,15 +383,22 @@ This will create:
...
 ```
 oc start-build --follow jenkins
 ```
 
+If a root CA cert was provided, also start a build of the Jenkins agent:
+
+```
+oc start-build --follow jenkins-agent
+```
+
```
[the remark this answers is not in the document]
jlebon: Possibly. I do like that we only pay for it if it's needed, but I see the appeal of keeping it consistent. Though in the no-cert case, there's nothing to actually build. We could make it be a no-op build I guess (the Dockerfile templating there would get hacky, or maybe just have two separate manifest files), though it might be more complex than it's worth.

jobs/build.Jenkinsfile
dustymabe: I feel like we would need this also in `build-arch.Jenkinsfile`? and possibly need to plumb that through to the "remote-session" in that job?

Maybe we don't need it in that job for some reason that I don't fully understand?

jlebon: I punted on that because I wasn't sure what the plan was there. If all the worker nodes we're planning to use will have access to the internal network, then I can wire through the root CA cert to the remote sessions. But if not, it might not be needed depending on how we'll proxy access.

Right now, I'm just trying to get at least x86_64 working though. Let's get this in and then work on the multi-arch bits in follow-ups?

dustymabe (approved):
LGTM

jlebon:
Just removed a line I had forgotten to remove. Merging.

PR #615, opened 2022-08-23
deploy: fix `has_additional_root_ca()` return value

jlebon:
Too much Rust.

Fixes 88a769a ("Simplify root CA cert provisioning").


dustymabe (approved):
LGTM

PR #616, opened 2022-08-23
streams: enable branched stream

dustymabe:
Now that F37 has branched from rawhide we'll start building the
branched stream again.


jlebon (approved):
Aside: I'm thinking we should delete `bodhi-updates`-related references.

Then for `branched`, maybe we can have something a little more formal/automated like we do for `next-devel`.

Issue #618, opened 2022-08-26
Design pipeline configuration strategy


jlebon:
We're working on making this pipeline usable to build RHCOS too. At a high-level, the biggest difference between RHCOS and FCOS in that respect is that the RHCOS streams have very different lifecycles.

E.g. OCP 4.6 was released [Oct 27, 2020](https://access.redhat.com/support/policy/updates/openshift/#dates). Maintenance will end Oct 27 2022. Many features and platforms were added since then, but we don't ship them for 4.6. Contrast with FCOS, where `stable` is at most 4 weeks apart from `testing-devel`. (This is also why branching cosa is much more needed for RHCOS than FCOS, where we're only recently getting around to doing this.)

As we add support for building RHCOS with this pipeline, we need to have some things be configurable. Here's a (likely non-comprehensive) list of possible configuration knobs:
- whether to build all bootimages
- whether to upload to AWS GovCloud or not (usually, dev streams are not until they GA)
- whether to build artifacts for certain platforms (e.g. we shouldn't build Nutanix images for streams older than 4.10 where enablement was done)
- sharing AMIs with specific OpenShift CI accounts
- uploading to some AWS/Aliyun zones

The way the existing RHCOS pipeline does this is using a "jobspec", which is a YAML file full of knobs. Though it also includes pipeline details like secret names and pod resource requests.

Another big one is the actual list of streams. Currently, this pipeline hardcodes `stable`, `testing`, ... The existing RHCOS pipeline avoids this by branching the pipeline code itself per releases, which has its own set of tradeoffs.

cgwalters:
This relates to https://github.com/coreos/coreos-assembler/issues/3008

cgwalters:
https://github.com/jlebon/fedora-coreos-pipecfg is a proposal that should be highlighted here

cgwalters:
https://github.com/coreos/coreos-assembler/pull/3094

dustymabe:
With https://github.com/coreos/fedora-coreos-pipeline/pull/631 I think we can mark the initial version of this as done. We're currently iterating on adding new stuff to the config.yaml (documentation PR in https://github.com/coreos/fedora-coreos-pipeline/pull/682).

Let's open new issues for further improvements here.

PR #619, opened 2022-08-30
Add concept of a pipeline config (pipecfg) repo and use it to remove all FCOS-specific knobs from this repo

jlebon:
This is a proposal for solving the tackling the configuration issue laid out in https://github.com/coreos/fedora-coreos-pipeline/issues/618.

The high-level is:
- we convert `streams.groovy` to a YAML config file
- we move more knobs from the configmap to this YAML config file
- we support the YAML config file living in a separate repo
- we remove the YAML config from this repo

See the individual commits for details.


travier:
Duplication is less important here than maintainability. It's OK if the configs are 99% equal as long deploying a new one is copy/pasting an existing one. We can remove/refactor the configs progressively after.

travier:
https://github.com/jlebon/fedora-coreos-pipecfg: This vaguely looks like what I'm saying so seems that I've missed something here.

travier:
The architectures are defined per version in RHCOS.

dustymabe:
I think this looks good to me as a start. The only thing I don't necessarily like is removing the `config.yaml` and the support for `in-tree` config in the last commit. I think until we worth out some kinks it could be good to have them coupled together.

I propose that I take this tomorrow and start to rebase it on top of latest `main` and then deploy it to our staging cluster to work out any problems.

Any objections?

travier:
If we can make a single line change + rebase of a different file to have an in repo downstream config then that's good enough for now and does simplify things a lot

travier:
This fails when the file is in tree with:
```
18:05:07  org.jenkinsci.plugins.workflow.cps.CpsCompilationErrorsException: startup failed:
18:05:07  Script1.groovy: 19: expecting '}', found 'file' @ line 19, column 25.
18:05:07             return readYaml file: "config.yaml"
18:05:07                             ^
```

dustymabe:
I've got a fix for that in my fork. Local testing seems to be going well. I'll post up a PR here soon and see if we can get the FCOS pipeline migrated.

dustymabe:
> I've got a fix for that in my fork. Local testing seems to be going well. I'll post up a PR here soon and see if we can get the FCOS pipeline migrated.

https://github.com/coreos/fedora-coreos-pipeline/pull/631

jlebon:
This was mostly merged as part of https://github.com/coreos/fedora-coreos-pipeline/pull/631. We can revisit the last commit once things stabilize.

PR #620, opened 2022-09-01
pipeline/bump-lockfile: fetch parent build for multi-arch upgrade tests

LakshmiRavichandran1:
Multi-arch upgrade tests fail when there is no parent build available for that arch. Adding functionality to fetch the parent build and incase there is no parent build, default to testing-devel

Fixes: https://github.com/coreos/coreos-assembler/issues/3017

### DIFF
--- a/jobs/bump-lockfile.Jenkinsfile
+++ b/jobs/bump-lockfile.Jenkinsfile
@@ -5,6 +5,7 @@ node {
     streams = load("streams.groovy")
     arches = streams.additional_arches.plus("x86_64")
     def pipecfg = pipeutils.load_config()
+    def s3_bucket = pipecfg['s3-bucket']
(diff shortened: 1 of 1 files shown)


dustymabe:
hey @LakshmiRavichandran1 - thank you for working on this.

I think the goal from https://github.com/coreos/coreos-assembler/issues/3017 is to modify the kola code in coreos-assembler (see linked kola.go file from the description of that issue) to perform this logic so it doesn't live in the pipeline. Does that make sense?

LakshmiRavichandran1:
sure, I will work around and come back to you on this.

LakshmiRavichandran1:
moving to [PR - 3080 ](https://github.com/coreos/coreos-assembler/pull/3080) for the fix and closing this.

PR #621, opened 2022-09-02
jobs/bump-lockfile: add option for skipping tests on arches

dustymabe:
Sometimes we have flakes. If 3 architectures pass tests and one flakes
we shouldn't need to re-run the 3 passing arches on a re-run of the
job. Allow the operator to selectively skip running tests on
architectures where they know the tests have already passed.


dustymabe:
Also have a second commit for parallelizing some more stuff in the job.

travier (approved):
LGTM but I have limited experience here

dustymabe:
ok did some testing on this and pushed up a fixup - merging since the change was trivial.

PR #623, opened 2022-09-02
jobs/build-fcos-buildroot: fix slack message topic

dustymabe:
This isn't the build-cosa job so we shouldn't report errors
that way.


bgilbert (approved):
LGTM

PR #624, opened 2022-09-07
jobs/build: warn when workaround for iso-live-login is detected

dustymabe:
```
commit fe2634dafe8202fdd01161c3ac8c8b5f564bb044
Date:   Tue Sep 6 23:38:48 2022 -0400

    jobs/build: warn when workaround for iso-live-login is detected
    
    For now we want to notify ourselves when this workaround is observed. It won't
    fail the build, just give us information. See
    https://github.com/coreos/fedora-coreos-tracker/issues/1233

commit 8f4ada8c3c9774c1ad105e55179e3bb362e3032f
Date:   Tue Sep 6 22:42:43 2022 -0400

    jobs/build: drop the extra testiso with rd.debug test
    
    It turns out that passing in `--qemu-iso` to `kola testiso` doesn't
    work. See https://github.com/coreos/coreos-assembler/issues/3071
    
    This effectively reverts 34c0d2c.

```


marmijo (approved):
lgtm!

PR #627, opened 2022-09-11
Add and use generic-webhook-trigger plugin

dustymabe:
```
commit 8cbed1420b2192549bb298754015105c97c578cf
Date:   Sun Sep 11 09:06:10 2022 -0400

    HACKING: document generic webhook setup in GitHub
    
    This explains what values to use when setting up a webhook in GitHub
    for the jobs that are using the Generic Webhook Trigger plugin.

commit 9101e2b66d8ab797a5ede43796550112da6f25cd
Date:   Sat Sep 10 11:05:43 2022 -0400

    jobs/build-cosa: use the Generic Webhook Trigger plugin
    
    This plugin allows us to get the exact branch name from the webhook
    payload itself and use that when doing a build. The githubPush()
    method we were using previously would only trigger the job to happen
    but then the job could select any possible branch that hadn't been
    built recently to build. This gives us tighter control.

commit ccff24bd27099a969e4954faf4b94a7f7b1c1df4
Date:   Mon Sep 5 22:08:29 2022 -0400

    jobs/build-fcos-buildroot: use Generic Webhook Trigger plugin
    
    This plugin allows us to only trigger the job when something in
    the ci/buildroot/ subdirectory has changed, which will cut down
    on wasted resources.

commit 6ca0ee5b59ca2e07793919f704a432b7f7098877
Date:   Sat Sep 10 14:17:56 2022 -0400

    plugins: add header with some comments
    
    It explains what the file is and how to generate
    a similar verion of it from a running cluster.

commit c575e8e69f4298ea5a9fff546b7c79c52b3b538d
Date:   Sat Sep 10 14:09:33 2022 -0400

    plugins: add generic-webhook-trigger plugin
    
    This plugin will allow us tighter grain control over what branches
    get built against. It also allows us more filtering. For example,
    if we only want to trigger a job when a subdirectory in a repo is
    changed.
```

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -252,6 +252,25 @@ uuidgen -r > secret
 oc create secret generic github-webhook-shared-secret --from-file=secret
 ```
 
+NOTE: the secret will be used again when setting up the generic webhooks in the next section.
+
+### [PROD] Setup webhooks for the Generic webhook plugin
+
+Add a new webhook to the following repos:
+
+- [`coreos/fedora-coreos-config`](https://github.com/coreos/fedora-coreos-config.git) for the `build-fcos-buildroot` job.
+- [`coreos/coreos-assembler`](https://github.com/coreos/coreos-assembler.git) for the `build-cosa` job.
+
+In this case we'll re-use the GitHub webhook shared secret text that
+was created in the previous section. Add a webhook for each repo that
+follows:
+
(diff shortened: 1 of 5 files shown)


marmijo (approved):
LGTM, though I'm no expert here.

PR #630, opened 2022-09-14
jobs/release: move oscontainer to quay.io/fedora/fedora-coreos

dustymabe:
Also keep old location active for now.

Fixes https://github.com/coreos/fedora-coreos-tracker/issues/1171

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -271,12 +271,12 @@ follows:
 - Content Type: `application/json`
 - Secret: Use the secret text from the GitHub webhook shared secret above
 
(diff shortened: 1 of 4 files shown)


cgwalters (approved):

cgwalters:
So this will take effect on the next releases, right?

dustymabe:
> So this will take effect on the next releases, right?

Right. The next run of any release job will push content to the new and old locations.

`next-devel` just finished: https://quay.io/repository/fedora/fedora-coreos?tab=tags

PR #632, opened 2022-09-15
jobs/{build,build-arch}: drop src_config_ref

dustymabe:
Since it was set it was overriding the stream every time the `build` job would run. This is kind of confusing behavior. Let's get rid of it and just force people who are hacking to use the real branch names, but in their fork.


bgilbert (approved):
I don't fully understand the implications of this, but LGTM.

PR #634, opened 2022-09-15
jobs/{build,build-arch}: drain AWS test account ID into config

dustymabe:
No need to hard code this into the jobs any longer. Let's drain this value into the config itself.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -28,6 +28,10 @@ source_config:
 s3_bucket: fcos-builds
 
 clouds:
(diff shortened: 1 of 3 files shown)


travier:
Nit (can be for later, but likely needed soon): can we make that an array/list? Thanks!

dustymabe:
ok I tested this now so we should be good for review/merge

cgwalters (approved):

PR #638, opened 2022-09-16
jobs/kola-azure: fix variable reference

dustymabe:
In the yaml config it's under clouds.azure.test_*.

Fixes e66dec6


cgwalters (approved):
*mumbles something about dynamic programming languages being fast to write and also fast to fail at runtime*

LGTM!  Hopefully will make Azure go green.  (or should it be blue?)

dustymabe:
> LGTM! Hopefully will make Azure go green. (or should it be blue?)

🤣

PR #639, opened 2022-09-16
support older COSA versions for multi-arch building

dustymabe:
```
commit a75cdfdac686da8e1058c4e0d3d75edbcf336064
Date:   Fri Sep 16 09:51:18 2022 -0400

    jobs/build-arch: only patch gf-set-platform in the FCOS pipeline
    
    This patch is a temporary hack, but it's still around for now and
    we need to start supporting the RHCOS pipeline. Let's conditionalize
    the patch on if we are operating on the FCOS pipeline.

commit d5563b7cf57bfb1b29b49717fca3f672ca303708
Date:   Fri Sep 16 09:46:08 2022 -0400

    config: add toplevel type: config setting
    
    This will allow us to use this information in the pipeline to perform
    certain actions or not.

commit adc6f865e1a7348773c0b541888da4c18fbd081d
Date:   Thu Sep 15 15:39:35 2022 -0400

    jobs/build-arch: bridge `cosa remote-session` to older RHCOS releases
    
    We recently did some golang migration [1] and happened to add the
    `cosa remote-session` pieces [2] after that migration. This makes
    it not easy to backport to ther branches. For a period of time
    let's use the latest COSA for our COSA pod for multi-arch builds
    instead of the versioned COSA. We will still use the versioned COSA
    on the remote so the actual building will happened with the correctly
    versioned COSA (substituted below and passed as an argument to
    `cosa remote-session create --image=`).
    
    [1] https://github.com/coreos/coreos-assembler/pull/2919
    [2] https://github.com/coreos/coreos-assembler/pull/2979

commit 6fad0526e315296556e7e542ff271ee1134d2780
Date:   Thu Sep 15 15:35:56 2022 -0400

    jobs/build-arch: support tagged COSA image streams
    
    We need to support more than just the "coreos-assembler:main" image stream.
    Add in a translation layer for image streams that are from the tagged
    rhcos-4.X branches.

```

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -144,10 +144,12 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}") {
         def basearch = params.ARCH
 
         // If we are using the image stream (the default) then just translate
-        // that into 'quay.io/coreos-assembler/coreos-assembler:main'.
+        // that into a quay registry equivalent (the multi-arch builders can't
+        // run containers from image streams).
         def image = params.COREOS_ASSEMBLER_IMAGE
-        if (image == "coreos-assembler:main") {
-            image = "quay.io/coreos-assembler/coreos-assembler:main"
+        if (image.startsWith("coreos-assembler:")) {
+            image = image.replaceAll("coreos-assembler:",
+                                     "quay.io/coreos-assembler/coreos-assembler:")
         }
 
         try { timeout(time: 240, unit: 'MINUTES') {
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -2,10 +2,12 @@ import org.yaml.snakeyaml.Yaml;
 
 def pipeutils, pipecfg, official, uploading, jenkins_agent_image_tag
 def src_config_url, s3_bucket, aws_test_accounts
+def pipetype
 node {
     checkout scm
     pipeutils = load("utils.groovy")
     pipecfg = pipeutils.load_pipecfg()
+    pipetype = pipecfg.type
     pod = readFile(file: "manifests/pod.yaml")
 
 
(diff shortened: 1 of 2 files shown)


jobs/build-arch.Jenkinsfile
```diff
@@ -92,9 +94,24 @@ def cosa_memory_request_mb = 2.5 * 1024 as Integer
...
+// it not easy to backport to ther branches. For a period of time
+// let's use the latest COSA for our COSA pod for multi-arch builds
+// instead of the versioned COSA. We will still use the versioned COSA
+// on the remote so the actual building will happened with the correctly
+// versioned COSA (substituted below and passed as an argument to
+// `cosa remote-session create --image=`).
+// [1] https://github.com/coreos/coreos-assembler/pull/2919
+// [2] https://github.com/coreos/coreos-assembler/pull/2979
+def cosa_pod_image = params.COREOS_ASSEMBLER_IMAGE
+if (cosa_pod_image =~ '^coreos-assembler:rhcos-4.(6|7|8|9|10|11)$') {
```
travier, line 108: Can we make that more generic to not have to edit it when 4.12 gets out?

dustymabe: We won't - 4.12 supports `cosa remote-session` so we won't need to do any special handling there. When this list is exhausted we'll remove this code.

travier: nice

travier (approved):

ravanelli (approved):

jlebon:
Hmm, I think I would've preferred having a e.g. `force_aws_serial_console_hack: true` temporary knob instead. We should try to avoid a generic `type` key like this IMO in favour of more targeted config knobs.

dustymabe:
> Hmm, I think I would've preferred having a e.g. `force_aws_serial_console_hack: true` temporary knob instead. We should try to avoid a generic `type` key like this IMO in favour of more targeted config knobs.

I'm not opposed to a different approach. This seemed the easiest and would have other uses. We already have a lot of `if rhcos; do this; if fcos; do that;` in other places in our stack. Might be hard to get away from it here too. But we can start with that as a goal.

PR #642, opened 2022-09-26
Start using artifacts from config.yaml

ravanelli:


cgwalters (commented):
We've made this more declarative, but at the cost of duplicating the list across all streams.  

It seems like a simple form of inheritance would help a lot here where there's support for extensions per architecture/stream.

ravanelli:
> We've made this more declarative, but at the cost of duplicating the list across all streams.
> 
> It seems like a simple form of inheritance would help a lot here where there's support for extensions per architecture/stream.

We had this discuss [here](https://github.com/travier/fedora-coreos-pipeline/pull/1#issuecomment-1258027179) , where I added `anchors` and `alias` for yaml

dustymabe:
Let me think about this a bit tomorrow. Obviously we want to keep this maintainable and easy to see differences from the defaults. I agree with Colin what we have in this PR with each set of information duplicated isn't ideal. I also agree with Timothée that the advanced yaml features appear to make thinks more complex.

This also intersects a bit with https://github.com/coreos/coreos-assembler/pull/3094, which we should probably discuss up front or agree to discuss soon in the context of this change.

travier:
I'd say we start with a verbose list of everything we need and then we move the ones that are common to all versions directly into the code and use an "additional" artifacts lists instead for artifacts only supported for new versions.
As we have versions reach EOL, we should be able to move artifacts out the additional list and into the code directly.

travier:
Or do what Colin suggests too in COSA but that won't work for previous versions.

travier:
So we'll likely have to do both, and progressively migrate to a config stored in COSA only.

ravanelli:
> So we'll likely have to do both, and progressively migrate to a config stored mostly in COSA.

The rhcos/fcos part in the artifacts would have to change. For `rhcos` we don't build the full list of artifacts all the time, specially for the old versions. We need a way to disable/enable bootimages if we want too, without affecting fcos if we move this part to cosa

dustymabe:
> Or do what Colin suggests too in COSA but that won't work for previous versions.

> So we'll likely have to do both, and progressively migrate to a config stored mostly in COSA.

Note that his proposal has the list of artifacts stored in the `src/config` so the definition is not in COSA but in the config repo. COSA just needs to support it.

dustymabe:
ok. Here is my suggestion here for the short term:

1. Forget the word `bootimages`. I don't think it adds much value either upstream or downstream.
2. Create a list of default artifacts to produce for an architecture.
3. Allow that list to be overridden for a given stream.


```yaml
# this is an FCOS pipeline
type: fcos

streams:
    stable:
      type: production
    testing:
      type: production
    next:
      type: production
      artifacts:
        aarch64:
          - AWS
          - GCP
          - Live
          - Metal
          - Metal4k
          - OpenStack
    testing-devel:
      type: development
      default: true
    next-devel:         # do not touch; line managed by `next-devel/manage.py`
      type: development # do not touch; line managed by `next-devel/manage.py`
    rawhide:
      type: mechanical
    branched:
      type: mechanical
    # bodhi-updates:
    #   type: mechanical
    # bodhi-updates-testing:
    #   type: mechanical

additional_arches: [aarch64, ppc64le, s390x]

source_config:

# remove to disable S3 uploads
s3_bucket: fcos-builds

default_artifacts:
  x86_64:
    - AWS
    - Aliyun
    - Azure
    - AzureStack
    - DigitalOcean
    - Exoscale
    - GCP
    - IBMCloud
    - Live
    - Metal
    - Metal4k
    - Nutanix
    - OpenStack
    - VMware
    - VirtualBox
    - Vultr
  aarch64:
    - AWS
    - Live
    - Metal
    - Metal4k
    - OpenStack

clouds:
  aws:
    # Accounts to share newly created AMIs with
    test_accounts:
      - "013116697141"
  azure:
    test_resource_group: fedora-coreos-testing
    test_storage_account: fedoracoreostesting
    test_storage_container: fedora-coreos-testing-image-blobs
  gcp:
    bucket: fedora-coreos-cloud-image-uploads

```

This example shows the default artifact list for `x86_64` and `aarch64` and shows aarch64 overridden for the `next` stream to include `GCP`.

jlebon:
> ok. Here is my suggestion here for the short term:

Seems reasonable to me. I think we'll probably want support for something like:

```yaml
streams:
    next:
      type: production
      additional_artifacts:
        aarch64:
          - coolplatform
```

where the list in that case is just appended to the default so that one can avoid repeating the common items in the case of adding new platforms, but that can come later.

The only modification I'd make is to have the artifact names be lowercase, i.e. matching the cosa command name. We did it the other way in the code to have the pretty names show up in the BlueOcean view, but in a config file it feels like we want to have actual valid identifiers and not pretty names.

ravanelli:
+1 @jlebon. Appending would make more sense, the full list is only used for the latest versions. The default list, at least for `rhcos` would be a small list, that is increased as new artifacts are added in the new versions.
+1 for lowercase too

jlebon:
> > The only modification I'd make is to have the artifact names be lowercase, i.e. matching the cosa command name. We did it the other way in the code to have the pretty names show up in the BlueOcean view, but in a config file it feels like we want to have actual valid identifiers and not pretty names.
> 
> I'm OK either way on this. Ultimately it doesn't matter if we use upper case letters here or not because the code will convert it all to lowercase before calling `buildextend` with it.

In my head, I was thinking we wouldn't lowercase anymore in the code. We'd assume that the artifacts in the config file were valid names and otherwise let `cosa` error out. We could I guess pre-validate them against a known list, but lowercasing before using them feels a bit odd.

dustymabe:
> We could I guess pre-validate them against a known list, but lowercasing before using them feels a bit odd.

My thinking is that if we know buildextend only takes lower case names then why not convert it and save the user a cycle if they made a mistake.

utils.groovy
```diff
@@ -199,4 +199,18 @@ def get_push_trigger() {
     ]
 }
 
+def get_artifacts(pipecfg, stream, basearch) {
+    // Get artifacts from config.yaml
+    def artifacts
+    if  (pipecfg.streams."${stream}".artifacts?."${basearch}" != null) {
```
jlebon, line 205: Minor: I think another syntax that'd be shorter is: `pipecfg.streams[stream]...`

utils.groovy
```diff
@@ -199,4 +199,18 @@ def get_push_trigger() {
...
+        artifacts = pipecfg.streams."${stream}".artifacts."${basearch}"
+        return artifact
+    }
+    else { // Merge default with additional artifacts
+        def default_artifacts =  pipecfg.default_artifacts."${basearch}"
+        def additional_atifacts = pipecfg.streams."${stream}".additional_artifacts?."${basearch}"
+        artifacts = default_artifacts + additional_atifacts
+    }
+    return artifacts
```
jlebon, line 225: Let's use [`artifacts.unique()`](https://docs.groovy-lang.org/latest/html/groovy-jdk/java/util/List.html#unique()) here to be defensive against duplicates?

config.yaml
```diff
@@ -27,6 +27,24 @@ source_config:
...
+        - GCP
+        - IBMCloud
+        - Live
+        - Nutanix
+        - Metal
+        - Metal4k
+        - OpenStack
+        - VirtualBox
+        - VMware
+        - Vultr
```
jlebon, line 47: Minor: I still think it'd be more appropriate in this config file to have the artifact names be lowercase. And ideally we also stop lowercasing in the code too.

We're defining a new config format, which may live outside this repo in the future, so I'd lean towards starting strict and simple and loosening later on if we really perceive a need.

This also makes the deduplication mentioned earlier safer.

utils.groovy
```diff
@@ -204,13 +204,24 @@ def get_artifacts(pipecfg, stream, basearch) {
     def artifacts
     if  (pipecfg.streams."${stream}".artifacts?."${basearch}" != null) {
         artifacts = pipecfg.streams."${stream}".artifacts."${basearch}"
-        return artifact
```
jlebon: This hunk should be part of the commit that introduces the function.

utils.groovy
```diff
@@ -199,4 +199,18 @@ def get_push_trigger() {
...
 }
 
+def get_artifacts(pipecfg, stream, basearch) {
+    // Get artifacts from config.yaml
+    def artifacts
+    if  (pipecfg.streams."${stream}".artifacts?."${basearch}" != null) {
+        artifacts = pipecfg.streams."${stream}".artifacts."${basearch}"
+        return artifact
+    }
+    else { // Merge default with additional artifacts
```
jlebon, line 208: Minor: why not on one line?

utils.groovy
```diff
@@ -199,4 +199,29 @@ def get_push_trigger() {
...
+        def default_artifacts =  pipecfg.default_artifacts."${basearch}"
+        def additional_atifacts = pipecfg.streams."${stream}".additional_artifacts?."${basearch}"
+        artifacts = default_artifacts + additional_atifacts
+    }
+    // For all architectures we need to build the metal/metal4k artifacts and the Live ISO. Since the
+    // ISO depends on the Metal/Metal4k images we'll make sure to put the Metal* ones in the first run
+    // and the Live ISO in the second run for parallel runs where the list is split
+    if (artifacts.contains('Metal') && artifacts.contains('Live')) {
```
jlebon, line 216: I think this can just be

```suggestion
    if (artifacts.contains('Live')) {
```

jobs/build.Jenkinsfile
```diff
@@ -458,20 +458,13 @@ lock(resource: "build-${params.STREAM}") {
...
-            artifacts.add(1, "Metal4k")
-            artifacts.add("Live")
+            def artifacts = pipeutils.get_artifacts(pipecfg, params.STREAM, basearch)
             // Run the two runs of parallel builds
             parallelruns = artifacts.collectEntries {
                 [it, {
                     def cmd = it.toLowerCase()
                     shwrap("""
-                    cosa buildextend-${cmd}
+                    cosa buildextend-"${cmd}"
```
jlebon, line 467: Do we actually need the quotes here?

utils.groovy
```diff
@@ -199,4 +199,29 @@ def get_push_trigger() {
...
+        artifacts = pipecfg.streams."${stream}".artifacts."${basearch}"
+    }
+    else { // Merge default with additional artifacts
+        def default_artifacts =  pipecfg.default_artifacts."${basearch}"
+        def additional_atifacts = pipecfg.streams."${stream}".additional_artifacts?."${basearch}"
+        artifacts = default_artifacts + additional_atifacts
+    }
+    // For all architectures we need to build the metal/metal4k artifacts and the Live ISO. Since the
+    // ISO depends on the Metal/Metal4k images we'll make sure to put the Metal* ones in the first run
+    // and the Live ISO in the second run for parallel runs where the list is split
```
jlebon, line 215: It's a bit awkward to do this in this function because it's an implementation detail of the caller that the list then gets split into two lists. I guess it's OK, but let's maybe reword the comment so it says something like: "...put the metal* ones at the start of the list and the live ISO at the end".

jobs/build.Jenkinsfile
dustymabe: I wouldn't think so. If we do I'd put it around the entire arg: `"buildextend-${cmd}"

utils.groovy
dustymabe: I agree. I almost think we could have a separate function that does this that is documented appropriately. Is it possible to return multiple values in java/groovy? If so we could just parallel what gets returned by the new function.

utils.groovy
dustymabe: True, since Live depends on Metal we can assume it exists?

utils.groovy
```diff
@@ -199,4 +199,18 @@ def get_push_trigger() {
     ]
 }
 
+def get_artifacts(pipecfg, stream, basearch) {
+    // Get artifacts from config.yaml
+    def artifacts
+    if  (pipecfg.streams."${stream}".artifacts?."${basearch}" != null) {
+        artifacts = pipecfg.streams."${stream}".artifacts."${basearch}"
+        return artifact
```
dustymabe: we should remove `return artifact` from this commit. It's wrong (should be `artifacts` and gets later cleaned up anyway).

utils.groovy
```diff
@@ -199,4 +199,29 @@ def get_push_trigger() {
     ]
 }
 
+def get_artifacts(pipecfg, stream, basearch) {
```
dustymabe, line 202: bikeshed: `get_desired_artifacts()` or `get_artifacts_to_build()`

config.yaml
```diff
@@ -27,6 +27,24 @@ source_config:
 # remove to disable S3 uploads
 s3_bucket: fcos-builds
 
+default_artifacts:
+  x86_64:
```
dustymabe, line 31: let's just go ahead and add all architectures up front. WDYT?

jobs/build.Jenkinsfile
ravanelli: I know what happened. 
I was getting the following error, when I changed it for testing, but the really issue is: I skipped  `buildextend-qemu`. Seems it should have a better error about the missing qemu or it could be even a bug?
```
21:32:48  + set -xeuo pipefail
21:32:48  ++ umask
21:32:48  + '[' 0022 = 0000 ']'
21:32:48  + cosa buildextend-aws
21:32:48  [Pipeline] sh
21:32:48  2022-09-29 00:32:48,709 INFO - symlink is for a build and publish command
21:32:48  2022-09-29 00:32:48,709 INFO - ore target aws found via symlink
21:32:48  2022-09-29 00:32:48,711 INFO - Targeting architecture: x86_64
21:32:48  2022-09-29 00:32:48,711 INFO - Targeting build: 412.86.202209290026-0
21:32:48  2022-09-29 00:32:48,715 INFO - Processed build for: OpenShift 4 (RHCOS-x86_64) 412.86.202209290026-0
21:32:48  2022-09-29 00:32:48,715 INFO - operating on rhcos-412.86.202209290026-0-aws.x86_64.vmdk
21:32:48  2022-09-29 00:32:48,715 INFO - Processing the build artifacts
21:32:48  2022-09-29 00:32:48,739 INFO - Staging temp image: /home/jenkins/agent/workspace/build-artifacts-Renata/tmp/tmpv_edgaee/rhcos-412.86.202209290026-0-aws.x86_64.vmdk
21:32:48  Traceback (most recent call last):
21:32:48    File "/usr/lib/coreos-assembler/cmd-buildextend-aws", line 115, in <module>
21:32:48      build.ensure_built()
21:32:48    File "/usr/lib/coreos-assembler/cosalib/build.py", line 263, in ensure_built
21:32:48      self.build_artifacts()
21:32:48    File "/usr/lib/coreos-assembler/cosalib/build.py", line 414, in build_artifacts
21:32:48      self._build_artifacts(*args, **kwargs)
21:32:48    File "/usr/lib/coreos-assembler/cosalib/qemuvariants.py", line 335, in _build_artifacts
21:32:48      meta_patch = self.mutate_image()
21:32:48    File "/usr/lib/coreos-assembler/cosalib/qemuvariants.py", line 251, in mutate_image
21:32:48      self.set_platform()
21:32:48    File "/usr/lib/coreos-assembler/cosalib/qemuvariants.py", line 226, in set_platform
21:32:48      self.image_qemu, self.tmp_image, self.platform])
21:32:48    File "/usr/lib/coreos-assembler/cosalib/qemuvariants.py", line 198, in image_qemu
21:32:48      qimage = os.path.join(
21:32:48    File "/usr/lib64/python3.10/posixpath.py", line 90, in join
21:32:48      genericpath._check_arg_types('join', a, *p)
21:32:48    File "/usr/lib64/python3.10/genericpath.py", line 152, in _check_arg_types
21:32:48      raise TypeError(f'{funcname}() argument must be str, bytes, or '
21:32:48  TypeError: join() argument must be str, bytes, or os.PathLike object, not 'NoneType'
21:32:48  error: failed to execute cmd-buildextend-aws: exit status 1
```

utils.groovy
```diff
@@ -199,4 +199,28 @@ def get_push_trigger() {
...
+def get_artifacts_to_build(pipecfg, stream, basearch) {
+    def artifacts
+    if  (pipecfg.streams[stream].artifacts?."${basearch}" != null) {
+        artifacts = pipecfg.streams[stream]['artifacts'][basearch]
+    } else { // Merge default with additional artifacts
+        def default_artifacts =  pipecfg['default_artifacts'][basearch]
+        def additional_atifacts = pipecfg.streams[stream].additional_artifacts?."${basearch}"
+        artifacts = default_artifacts + additional_atifacts
+    }
+    return artifacts.unique()
```
dustymabe, line 224: with this `unique` call here are we allowing users to mistakenly specify things twice and have it still work?

utils.groovy
```diff
@@ -199,4 +199,28 @@ def get_push_trigger() {
     ]
 }
 
+def change_metal_artifacts_list_order(artifacts) {
+    if (artifacts.contains('live')) {
```
dustymabe, line 203: should we check inside here to make sure both `metal` and `metal4k` exist?

utils.groovy
```diff
@@ -199,4 +199,28 @@ def get_push_trigger() {
...
+        artifacts.remove("live")
+
+        artifacts.add(0, "metal")
+        artifacts.add(1, "metal4k")
+        artifacts.add("live")
+    }
+    return artifacts
+}
+
+def get_artifacts_to_build(pipecfg, stream, basearch) {
```
dustymabe, line 215: add comment above this function? ``// gets desired artifacts to build from pipeline config`.

dustymabe (commented):
Mostly LGTM - Asked a few questions/comments..

utils.groovy
ravanelli: Um, not sure. I guess if we have `live` we must have these other two? Or is it not always the case?

dustymabe: IIUC if you have `live` you need the others. @jlebon would know if `metal4k` is optional, but at least `metal` I know is required.

utils.groovy
jlebon: Yeah exactly. I can imagine this happening when we promote an artifact to the default set.

jobs/build-arch.Jenkinsfile
```diff
@@ -434,20 +434,9 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}") {
...
-            }
-            // For all architectures we need to build the metal/metal4k artifacts and the Live ISO. Since the
-            // ISO depends on the Metal/Metal4k images we'll make sure to put the Metal* ones in the first run
-            // and the Live ISO in the second run.
-            artifacts.add(0, "Metal")
-            artifacts.add(1, "Metal4k")
-            artifacts.add("Live")
+            def artifacts = pipeutils.get_artifacts(pipecfg, params.STREAM, basearch)
+            // Sort the artifacts for two parallel runs
+            artifacts = pipeutils.change_metal_artifacts_list_order(artifacts)
```
jlebon, line 439: Minor: this hunk should be part of the same commit that modifies `build.Jenkinsfile` since we're always trying to keep them in sync.

config.yaml
```diff
@@ -27,6 +27,41 @@ source_config:
 # remove to disable S3 uploads
 s3_bucket: fcos-builds
 
+default_artifacts:
```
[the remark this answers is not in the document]
ravanelli: And where it should live? I will do it as a followup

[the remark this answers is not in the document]
dustymabe: Maybe in the toplevel `README.md` or we could create a new `CONFIG.md` and link to it from `README.md`.

utils.groovy
dustymabe: We can leave it like it is in the code right now and do any fixups later.

dustymabe (approved):
LGTM - do we need to test this somewhere or have you already done that?

ravanelli:
> LGTM - do we need to test this somewhere or have you already done that?

Already tested it in the internal rhcos namespace.

jlebon (approved):
LGTM! Will let you click the green button. :)

config.yaml
```diff
@@ -27,6 +27,41 @@ source_config:
...
+  aarch_64:
```
jlebon, line 31: Shouldn't this be `aarch64`?

dustymabe: yep. I'll fix it in my open PR.

dustymabe: added a commit to https://github.com/coreos/fedora-coreos-pipeline/pull/645

PR #643, opened 2022-09-28
tree: stop cloning releng-automation repo, use cosa

jlebon:
We've moved the scripts we need from that repo to cosa, so stop cloning the repo and use the ones from the cosa directory instead.

Requires: https://github.com/coreos/coreos-assembler/pull/3113

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -594,7 +594,6 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}") {
             }
             parallelruns['OSTree Import: Compose Repo'] = {
                 shwrap("""
-                cosa shell -- \
                 /usr/lib/coreos-assembler/fedmsg-send-ostree-import-request \
                     --build=${newBuildID} --arch=${basearch} \
                     --s3=${s3_stream_dir} --repo=compose \
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -171,13 +171,6 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}") {
 
         try { timeout(time: 240, unit: 'MINUTES') {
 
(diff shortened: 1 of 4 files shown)


jobs/build-arch.Jenkinsfile
```diff
@@ -234,8 +227,6 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}") {
                     /run/kubernetes/secrets/fedora-messaging-coreos-key
                 cosa remote-session sync {,:}/run/kubernetes/secrets/fedora-messaging-coreos-key/
             fi
-            # sync over the send-ostree-import-request.py from the automation repo
-            cosa remote-session sync {,:}/var/tmp/fcos-releng/coreos-ostree-importer/send-ostree-import-request.py
```
jlebon, line 238: @dustymabe Is there a reason we were running the script from the remote instead of directly on the controller? I couldn't think of one so simplified this, but I can undo it if required.

dustymabe: From memory (may lack accuracy): I think it's because the script does some verification at the end that required some of the local files. I know that's true for the signing requests we send.. Maybe not for the OSTree import request.

jobs/build-arch.Jenkinsfile
```diff
@@ -603,8 +594,7 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}") {
             }
             parallelruns['OSTree Import: Compose Repo'] = {
                 shwrap("""
-                cosa shell -- \
```
dustymabe, line 606: can we put the removal of `cosa shell --` here and the `sync` of the `send-ostree-import-request.py` in a separate initial commit here. I think the subtlety there is worth it's own descriptive commit message.

dustymabe:
Can you remove the `cosa remote-session sync` in the first commit too?

dustymabe (approved):
LGTM

PR #644, opened 2022-09-29
migrate credentials to kubernetes credentials plugin

dustymabe:
This patch set converts all credentials to be auto-created directly from kuberetes secrets. This means we don't have to mount the secrets into the jenkins or cosa pods and can bind them when needed at runtime using `withCredentials()`.


dustymabe:
Probably best to review each commit individually. I can break some of this out into other PRs if desired as well.

I'll attempt to do some testing on this now in staging.

dustymabe:
I've tested basic concepts here in staging:

- builds work
- archiving to s3 works
- multi-arch building works
- running an AWS kola test works

I'm ready to take this to production and flesh out any remaining issues or I'm open to input on how best to proceed.

jlebon:
Awesome work!

A lot of non-trivial changes here. To make review easier, can we split it into a couple of smaller PRs? Even just in two would help. Otherwise, I'm OK keeping it all together but it'll take longer.

dustymabe:
broke out the first 4 commits into https://github.com/coreos/fedora-coreos-pipeline/pull/648

dustymabe:
another commit broken into https://github.com/coreos/fedora-coreos-pipeline/pull/650

dustymabe:
and three more commits into https://github.com/coreos/fedora-coreos-pipeline/pull/651

dustymabe:
one more commit broken out into https://github.com/coreos/fedora-coreos-pipeline/pull/653

dustymabe:
two more commits broken out into https://github.com/coreos/fedora-coreos-pipeline/pull/656

dustymabe:
All of this is now merged in other PRs.

PR #645, opened 2022-09-29
jobs/build-arch: update hack for aarch64 AWS serial console

dustymabe:
Now that platforms.yaml is starting to roll out we need to stop applying the hack in those cases. Once it is everywhere we can drop the hack completely.


jlebon (approved):
IIUC, it looks like the patch itself already contains a conditional on `platforms.yaml` existing so this isn't strictly necessary but it's clearly more efficient to skip it earlier on.

jlebon (approved):

PR #647, opened 2022-09-30
utils: fix null entry in list of artifacts

dustymabe:
If default_artifacts or additional_atifacts are empty then we'll get `null` in the list of artifacts. Let's fix that.


HuijingHei (approved):
LGTM

PR #649, opened 2022-09-30
updates for aws secrets/config

dustymabe:
```
commit d1c58483349224ec27b5446f4464b4fa807e77ab
Date:   Tue Sep 27 16:17:11 2022 -0400

    treewide: migrate aws credentials/secrets to kube creds plugin
    
    This is a large commit that does a few things:
    
    - Simplifies the aws-build-upload-config secret by removing the
      accessKey and secretKey components of that secret
    - Converts the aws secrets into ones integrated by the kubernetes
      credentials plugin and deletes the aws jenkins credential
        - Also converts the sync-stream-metadata job to use a file based
          credential along with the AWS_CONFIG_FILE env variable.
        - Also removes the no longer used aws credential plugin
    - Removes mounts for AWS secrets from the pod and jenkins manifests

commit 2f65d4f6208b6d4c898f16d9930901be4895bba5
Date:   Tue Sep 27 16:24:04 2022 -0400

    utils: add tryWithOrWithoutCredentials()
    
    This allows for a piece of code to be run regardless of if the
    credential exists or not. Should allow for the credential to be
    dynamically loaded and then conditional logic to be utilized based
    on if the credential was loaded successfully.

```


jobs/build.Jenkinsfile
```diff
@@ -634,65 +640,68 @@ lock(resource: "build-${params.STREAM}") {
...
-          //    // We consider the Kubernetes kola tests to be a followup job, so we use `wait: false` here.
-          //    build job: 'kola-kubernetes', wait: false, parameters: [
-          //        string(name: 'STREAM', value: params.STREAM),
-          //        string(name: 'VERSION', value: newBuildID),
-          //        string(name: 'S3_STREAM_DIR', value: s3_stream_dir),
-          //        string(name: 'SRC_CONFIG_COMMIT', value: src_config_commit)
-          //    ]
-          //}
-        }
-        if (!params.MINIMAL && uploading &&
```
jlebon, line 661: Don't we still need those conditionals in this and following if-statements?

jobs/build.Jenkinsfile
```diff
@@ -145,6 +145,12 @@ lock(resource: "build-${params.STREAM}") {
 
         try { timeout(time: 240, unit: 'MINUTES') {
 
+        // Add in AWS Build Upload credentials here if they exist. In
+        // the future we might choose to be more granular about when
+        // we load this.
```
jlebon, line 150: That's unfortunate. The semantics of `tryWithOrWithoutCredentails` feel funny, but I see how you got here. We fetch things from AWS in a couple of places.

I think eventually we should clean up the `s3_stream_dir` thing though so that it doesn't carry so much weight. We now have a proper config file, so we can make the presence/absence of `s3_bucket` canonical (if it's defined, assume we're pushing; but neuter it if we're in the official Jenkins as a convenience to devs).

HACKING.md
```diff
@@ -80,41 +80,45 @@ create a credentials config as a secret within OpenShift.
...
+    jenkins.io/credentials-type=secretFile
+oc annotate secret/aws-build-upload-config \
+    jenkins.io/credentials-description="AWS build upload credentials config"
 ```
 
 We also have a second AWS config that can be used for running kola
 tests. If you have a single account that has enough permissions for
 both then you can use the same account for both uploading builds and
-running kola tests (i.e. re-use `upload-secret` from above. If not then
+running kola tests (i.e. re-use secret from above. If not then
```
jlebon, line 105: Not new, but might as well fix the missing closing parenthesis here.

jobs/build.Jenkinsfile
dustymabe: > I think eventually we should clean up the `s3_stream_dir` thing though so that it doesn't carry so much weight.

agree. I think there are improvements to be made here. It's hard while I'm in the code not to try to do everything, but that would be boiling the ocean.

Still lots of work to do!

jobs/build.Jenkinsfile
dustymabe: I converted it. Now all of these cloud test kickoff stanzas are nested inside a toplevel `if (!params.MINIMAL && uploading) {`.

dustymabe: If you ignore whitespace changes in your diff it might be harder to see this.

jlebon (approved):
LGTM!

PR #650, opened 2022-09-30
treewide: remove hardcoded git repo paths

dustymabe:
This attempts to remove instances where the fedora-coreos-config was hardcoded rather than pulling from the config.yaml. This should make it easier to support other src/config repos in the future.


jobs/kola-aws.Jenkinsfile
```diff
@@ -1,9 +1,10 @@
-def pipeutils, pipecfg, s3_bucket, official
+def pipeutils, pipecfg, s3_bucket, official, src_config_url
 node {
     checkout scm
     pipeutils = load("utils.groovy")
     pipecfg = pipeutils.load_pipecfg()
     s3_bucket = pipecfg.s3_bucket
+    src_config_url = pipecfg.source_config.url
```
jlebon, line 7: Cool with this, but I'd like to go the other direction actually where we directly reference e.g. `pipecfg.source_config.url` where needed instead of adding more and more aliases here. That was partly why I really wanted a short `pipecfg` name for the config.

jlebon (approved):

jobs/kola-aws.Jenkinsfile
dustymabe: I'm cool with that too. It just takes this a step farther IIUC. Let's do that in a followup.

PR #652, opened 2022-09-30
WIP: RHCOS pipeline changes

ravanelli:
This PR contains a rebase with main and `rhcos` branch

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -546,6 +546,14 @@ value from the `manifests/pipeline.yaml` OpenShift template. This is
 currently as designed (see
 [#65](https://github.com/coreos/fedora-coreos-pipeline/issues/65)).
 
+If you update the Jenkins config in `jenkins/config`, then you need to update
+it in the clsuter too:
+
+```
+oc delete configmap/jenkins-casc-cfg
+oc create configmap jenkins-casc-cfg --from-file=jenkins/config
+```
+
 ### Nuking everything
 
 One can leverage Kubernetes labels to delete all objects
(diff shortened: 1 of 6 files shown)


jlebon:
With the first commit in https://github.com/coreos/fedora-coreos-pipeline/pull/666, you should be able to just add `yumrepos` to the config YAML and drop [b0e4350](https://github.com/coreos/fedora-coreos-pipeline/pull/652/commits/b0e43508644f4e0540afbb00def17b68ad063a39).

jlebon:
When rebasing onto https://github.com/coreos/fedora-coreos-pipeline/pull/671, you'll want to rerun `jenkins.yaml` using something like this:

```
oc process --filename=manifests/jenkins.yaml --param=NAMESPACE=... ... | oc replace -f -
```

And then rerun `./deploy` as well and start another build of `jenkins-agent`.

cgwalters:
For testing purposes we can make `quay.io/rhel-coreos-devel` or whatever right?  I imagine we could also ask DPTP to run a quay instance in the CI cluster...

jlebon:
> For testing purposes we can make `quay.io/rhel-coreos-devel` or whatever right? I imagine we could also ask DPTP to run a quay instance in the CI cluster...

Right, we could do that. I'd prefer pushing to a namespace that already exists though so we don't have to manage yet another account somewhere and track its credentials, etc... If that doesn't exist, creating one sounds fine.

cgwalters:
Related to this though, I have some more work in the background for https://github.com/coreos/coreos-assembler/issues/2685 which as you know turns this all on its head - the *canonical source of truth* is a container image, so in this model a registry is required, s3 is optional.

PR #653, opened 2022-09-30
treewide: convert fedora messaging credentials to support k8s creds plugin

dustymabe:
This commit does several things:

- Converts the fedora-messaging-cfg secret (now fedora-messaging-config) into a secret rather than a configmap so it can be dynamically loaded with withCredentials and not required to be hard mounted into the pods.
- Sets up a new fedora-messaging-coreos-x509-cert secret (was fedora-messaging-coreos-key) in the x509ClientCert type for the kube credentials plugin.
- Removes the old configmap/secret mounts from the COSA pod.yaml
- Switches configs/fedmsg.toml to have a templated string `FEDORA_MESSAGING_X509_CERT_PATH` that will be substituted out at runtime with the actual path of the files.
- Adds tryWithMessagingCredentials() to aid in setup/use of the config and secrets for fedora messaging.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -341,17 +341,28 @@ oc annotate secret/cosa-push-registry-secret \
 
 ### [PROD] Create fedora-messaging configuration
 
-First create the configmap:
+First create the Fedora Messaging configuration secret:
 
 ```
-oc create configmap fedora-messaging-cfg --from-file=configs/fedmsg.toml
+oc create secret generic fedora-messaging-config \
+    --from-literal=filename=fedmsg.toml \
+    --from-file=data=configs/fedmsg.toml
+oc label secret/fedora-messaging-config \
+    jenkins.io/credentials-type=secretFile
+oc annotate secret/fedora-messaging-config \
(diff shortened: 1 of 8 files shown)


dustymabe:
rebased on top of latest `main`.

HACKING.md
```diff
@@ -341,17 +341,28 @@ oc annotate secret/cosa-push-registry-secret \
 
 ### [PROD] Create fedora-messaging configuration
 
-First create the configmap:
+First create the Fedora Messaging configuration secret:
 
 ```
-oc create configmap fedora-messaging-cfg --from-file=configs/fedmsg.toml
+oc create secret generic fedora-messaging-config \
```
jlebon, line 347: Storing the config as a secret is awkward, but I don't have a better suggestion. Would be cool if the file supported inlining the cert (it doesn't), then it could just be one secret.

jlebon (approved):
LGTM!

HACKING.md
dustymabe: yep :(

jlebon (approved):

PR #655, opened 2022-09-30
plugins: update and add comments to plugins.txt

dustymabe:


dustymabe:
I was just experimenting with this in the background. We can merge it after the releases next week.

jenkins/controller/plugins.txt
```diff
@@ -17,11 +17,15 @@
...
+github-oauth:0.39
+kubernetes-credentials-provider:1.199.v4a_1d1f5d074f
+pipeline-github:2.8-138.d766e30bb08b
+pipeline-utility-steps:2.13.0
+slack:625.va_eeb_b_168ffb_0
+timestamper:1.20
+
+# The below list are plugins that are also in base-plugins.txt but for
+# some reason or other we need to temporarily freeze or fast-track.
+token-macro:293.v283932a_0a_b_49
```
jlebon, line 31: Can we add a comment for this with some of what's in the commit message?

jlebon (approved):
Looks really nice to me! Would be cool to have a `bump-plugins` job that opened a PR with bumped versions every e.g. month. Testing is a pain, but we could script it.

jenkins/controller/plugins.txt
dustymabe: updated

jlebon (approved):

jlebon (approved):

PR #656, opened 2022-09-30
treewide: migrate slack-api-token for use with k8s creds plugin

dustymabe:
~~Also add `trySlackSend()` function, use it across the tree~~

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -238,8 +238,12 @@ If you want to be able to have build status messages appear in Slack,
 create a `slack-api-token` secret:
 
 ```
-$ echo -n "$TOKEN" > slack-token
-$ oc create secret generic slack-api-token --from-file=token=slack-token
+TOKEN=<token>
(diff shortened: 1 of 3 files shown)


utils.groovy
```diff
@@ -194,6 +194,18 @@ def tryWithMessagingCredentials(Closure body) {
     }
 }
 
+// Send slack message if the credential exists. Do nothing if not.
+def trySlackSend(color, message) {
```
jlebon, line 198: What does this get us? All the Slack sends are gated on `official` already. Keying off of the credential existing makes sense, but as a complexity trade I'd remove the `official` gates in that case.

jlebon: Maybe simpler we skip that commit for now and circle back to it?

dustymabe: yeah. The desire is to remove `official` in the future. I can drop that commit.

dustymabe: removed that 2nd commit

jlebon (approved):

PR #657, opened 2022-09-30
utils: add trySlackSend() function, use it across the tree

dustymabe:
This will make sending a slack message conditional on if the credential exists.


dustymabe:
broken out from https://github.com/coreos/fedora-coreos-pipeline/pull/656 for us to consider later

jlebon:
If the goal is to make it easier to use Slack from non-official pipelines, I'd say we should make the room a config knob instead, which we can do by removing `room` from `slack.yaml`, add it to `config.yaml`, and pass it when calling `slackSend`.

And possibly refuse to post to Slack if the official channel name is used in non-official.

PR #660, opened 2022-09-30
utils: fix undefined `COREOS_ASSEMBLER_REMOTE_SESSION` handling

jlebon:
We run with `set -u` so we need to provide a default value if `COREOS_ASSEMBLER_REMOTE_SESSION` is not defined.

Fixes ebdec22 ("treewide: convert fedora messaging credentials to support k8s creds plugin").


dustymabe (approved):
LGTM

PR #663, opened 2022-10-03
HACKING.md: ensure webhook secret has no newline

jlebon:
In GitHub, we'll insert the secret without a newline, so we need to make sure that we don't have a newline here either.


dustymabe (approved):
LGTM

PR #664, opened 2022-10-03
config.yaml: drop `type: fcos` knob

jlebon:
Let's not have code that explicitly switches on FCOS vs RHCOS in the pipeline. Instead, let's try to take each instance on a case-by-case basis and formalize the distinction they're trying to express at a higher level.

In the case of the aarch64 serial console issue, we'll get that via the recent `platforms.yaml` work. For now, to not regress let's introduce a temporary knob to activate the AWS aarch64 serial console hack.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -1,6 +1,3 @@
-# this is an FCOS pipeline
-type: fcos
-
 streams:
     stable:
       type: production
@@ -69,6 +66,9 @@ default_artifacts:
     - virtualbox
     - vmware
     - vultr
+
+# Temporary hack for AWS aarch64; see comment in `build-arch.Jenkinsfile`.
+aws_aarch64_serial_console_hack: true
 
 clouds:
   aws:
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -2,12 +2,10 @@ import org.yaml.snakeyaml.Yaml;
 
 def pipeutils, pipecfg, official, uploading, jenkins_agent_image_tag
 def src_config_url, s3_bucket, aws_test_accounts
-def pipetype
 node {
     checkout scm
     pipeutils = load("utils.groovy")
     pipecfg = pipeutils.load_pipecfg()
-    pipetype = pipecfg.type
     pod = readFile(file: "manifests/pod.yaml")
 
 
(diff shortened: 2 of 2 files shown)


config.yaml
```diff
@@ -75,6 +72,8 @@ clouds:
     # Accounts to share newly created AMIs with
     test_accounts:
       - "013116697141"
+    # Temporary hack for AWS aarch64; see comment in `build-arch.Jenkinsfile`.
+    aarch64_serial_console_hack: true
```
dustymabe, line 76: I *think* I'd like to put this somewhere different. I know this is temporary so it doens't really matter and I probably shouldn't argue this point, but...

This section to me is more about the cloud upload (i.e. not affecting the generated artifact (file), but rather the image creation in the cloud).

I'd be fine with a toplevel `aws_aarch64_serial_console_hack: true`, especially since it's temporary. This code/patch is completely a hack right now and hopefully we don't have to do stuff like this in the future.

jlebon: I'm fine with having it be at the top level.

dustymabe: I think this still needs to be moved in the config to the top level.

jobs/build-arch.Jenkinsfile
```diff
@@ -463,18 +461,18 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}") {
...
             // Right now we only patch if platforms.yaml hasn't made it to this stream yet.
             // Fold this back into the above parallel runs (i.e. add to config.yaml
-            // artifacts list for aarch64 and delete below code) once platforms.yaml
+            // artifacts list for aarch64 and delete below code and knob) once platforms.yaml
             // exists everywhere. https://github.com/coreos/fedora-coreos-config/pull/1181
             if (basearch == "aarch64") {
                 stage('AWS') {
-                    shwrap("""
-                    if [ "${pipetype}" == "fcos" ]; then
+                    if (pipecfg.aarch64_serial_console_hack) {
```
dustymabe, line 468: once you move the vaule in the config I think it will be:

```suggestion
                    if (pipecfg.aws_aarch64_serial_console_hack) {
```

config.yaml
jlebon: Ahh missed the `git add`.

dustymabe (approved):
LGTM. Feel free to merge once builds for this week's prod streams are done.

PR #665, opened 2022-10-03
jobs: drop unnecessary pipecfg aliases

jlebon:
Stop fishing values out of `pipecfg` into separate variable and get in
the habit of referencing `pipecfg` directly instead.

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -1,20 +1,13 @@
 import org.yaml.snakeyaml.Yaml;
 
 def pipeutils, pipecfg, official, uploading
-def src_config_url, s3_bucket, aws_test_accounts
 node {
     checkout scm
     pipeutils = load("utils.groovy")
     pipecfg = pipeutils.load_pipecfg()
     pod = readFile(file: "manifests/pod.yaml")
 
-
     def jenkinscfg = pipeutils.load_jenkins_config()
-    src_config_url = pipecfg.source_config.url
-    s3_bucket = pipecfg.s3_bucket
-
-    // Extra AWS testing accounts to share images with
-    aws_test_accounts = pipecfg.clouds?.aws?.test_accounts
 
     official = pipeutils.isOfficial()
     if (official) {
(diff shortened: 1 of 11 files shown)


jlebon:
Requires: https://github.com/coreos/fedora-coreos-pipeline/pull/664

dustymabe:
converting to draft so that we don't merge before this week's triple release

jlebon:
Rebased!

This was mostly a mechanical change but deserves a careful review.

jobs/build-arch.Jenkinsfile
```diff
@@ -191,6 +184,15 @@ lock(resource: "build-${params.STREAM}-${params.ARCH}") {
...
 
+        def local_builddir = "/srv/devel/streams/${params.STREAM}"
+        def ref = params.STREAM
+        def src_config_commit
+        if (params.SRC_CONFIG_COMMIT) {
+            src_config_commit = params.SRC_CONFIG_COMMIT
+        } else {
+            src_config_commit = shwrapCapture("git ls-remote ${pipecfg.source_config.url} ${ref} | cut -d \$'\t' -f 1")
+        }
+
```
dustymabe, line 195: I think this is an artifact of a bad rebase

jobs/kola-azure.Jenkinsfile
```diff
@@ -101,6 +93,8 @@ lock(resource: "kola-azure-${params.ARCH}") {
                              file(variable: 'AZURE_KOLA_TESTS_CONFIG_AUTH',
                                   credentialsId: 'azure-kola-tests-config-auth')]) {
 
+                // to make things less wordy
+                def azure_cfg = pipecfg.clouds.azure
```
dustymabe, line 96: For this file I'm a little less convinced of the value of doing this. We re-use the variables often below. Maybe we just take the variables that were defined at the top and define them here instead and leave the below variables as they were? So block right here would become:

```
def azure_testing_resource_group = pipecfg.clouds.azure.test_resource_group
def azure_testing_storage_account = pipecfg.clouds.azure.test_storage_account
def azure_testing_storage_container = pipecfg.clouds.azure.test_storage_container
```

jlebon: No strong opinion, but I do like the concept of aliasing a whole group and referencing that instead of multiple individual variables. It makes it clear they're related and referencing them is more or less the same number of characters.

If you'd prefer I switch it back, I'm fine with that.

jobs/build-arch.Jenkinsfile
jlebon: Fixed!

jobs/kola-azure.Jenkinsfile
dustymabe: Yeah I think I'd prefer to switch back. This file is small and now we'll be declaring the variables closer to where we use them, which will help cut down on confusion.

jlebon:
> still one instance of `src_config_url` in `build-arch.Jenkinsfile`.

Ahh yup, this snuck in during the rebase.

Updated!

dustymabe (approved):
LGTM

ravanelli (approved):

PR #668, opened 2022-10-04
move fedora messaging cred back to x509ClientCert type

dustymabe:


dustymabe:
marked as draft so I can make sure to update the fedmsg.toml at the same time it merges.

jlebon (approved):

PR #669, opened 2022-10-04
jobs/kola-aws: parallelize tests; switch xen to i3.large

dustymabe:
```
commit fc19699a056b4f195972413abd9868b8b7ca96c2
Date:   Tue Oct 4 11:27:37 2022 -0400

    jobs/kola-aws: switch Xen test to i3.large to verify NVME
    
    Since we had a user report of broken NVME on i3.large let's switch
    our Xen test to that instance type and verify the NVME device works
    there. Also add in a test to make 100% sure we chose a Xen instance.

commit 45fe3510a5846f3388a40cc559558c9dd1c55fd7
Date:   Tue Oct 4 10:11:34 2022 -0400

    jobs/kola-aws: parallelize tests
    
    We don't typically have problems with quota on AWS so let's run all
    these in parallel. It has a few benefits:
    
    1. Faster
    2. If a test in Kola:Full fails because of a flake we still get the
       other tests (Xen, Ice Lake, Graviton3) to run. So we can safely
       ignore the flake in the future, but also get test results for
       those other tests.

```

Requires https://github.com/coreos/fedora-coreos-config/pull/2005

### DIFF
--- a/jobs/kola-aws.Jenkinsfile
+++ b/jobs/kola-aws.Jenkinsfile
@@ -92,13 +92,22 @@ try { timeout(time: 90, unit: 'MINUTES') {
                 }
                 parallelruns['Kola:Xen'] = {
                     // https://github.com/coreos/fedora-coreos-tracker/issues/997
+                    // Run this test on i3.large so we can also run ext.config.platforms.aws.nvme
+                    // to verify access to instance storage nvme disks works
+                    // https://github.com/coreos/fedora-coreos-tracker/issues/1306
+                    // Also add in the ext.config.platforms.aws.assert-xen test just
+                    // to sanity check we are on a Xen instance.
+                    def xen_tests = tests
+                    if (xen_tests == "basic") {
+                        xen_tests = "basic ext.config.platforms.aws.nvme ext.config.platforms.aws.assert-xen"
+                    }
                     fcosKola(cosaDir: env.WORKSPACE,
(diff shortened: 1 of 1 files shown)


jlebon (commented):
I *think* the problem with this is that all `fcosKola` instances will want to write to the same output dir. We need to add a knob there first that we can drive from here to avoid that.

jobs/kola-aws.Jenkinsfile
```diff
@@ -74,27 +74,42 @@ try { timeout(time: 90, unit: 'MINUTES') {
...
                     fcosKola(cosaDir: env.WORKSPACE,
                              build: params.VERSION, arch: params.ARCH,
-                             extraArgs: tests,
+                             extraArgs: xen_tests,
                              skipUpgrade: true,
                              skipBasicScenarios: true,
-                             marker: "kola-m4",
-                             platformArgs: '-p=aws --aws-region=us-east-1 --aws-type=m4.large')
-                }, m6i: {
+                             marker: "kola-i3",
```
jlebon, line 109: Can we have these makers match the stage name? That way it's easier for someone less familiar with AWS to download the right artifact after seeing the failing stage from the pipeline artifact.

E.g. `kola-xen`, `kola-intel-ice-lake`, and `kola-graviton`?

jlebon (commented):
Ahh right, I forgot about the marker stuff we added.

jlebon (approved):

PR #671, opened 2022-10-04
Rework jenkins-agent handling

jlebon:
Because of the additional root CA, we need the ability to optionally set
up a BuildConfig that will layer the root CA onto the base jenkins-agent
image.

However, the way we did this was exposing this choice to the Jenkins
jobs where they had to know which tag to use. This becomes cumbersome if
we also want to make use of coreos-ci-lib helpers.

Rather than trying to plumb that all the way through, we change tactics:
1. set a system property telling the Kubernetes plugin to default to our
   Jenkins agent imagestream tagged `:latest`
2. default `:latest` to be the base OpenShift Jenkins agent image
3. if an additional root CA is provided, make `:latest` be the output of
   the `jenkins-agent` BuildConfig

That way, jobs no longer have to set any explicit Jenkins agent image,
and the Kubernetes plugin will automatically choose the correct one.

This actually also fixes an issue with CoreOS CI where it's using the
upstream Jenkins agent image instead of the default image from the
OpenShift release payload because there, we haven't been explicitly
overriding the image for a long time (see
https://github.com/coreos/coreos-ci-lib/commit/f1eca31ed7d7).

This is also prep for dropping `pod.yaml` and using `cosaPod` instead.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -441,9 +441,9 @@ a secret. This assumes `ca.crt` is a file in the working directory:
 ```
 oc create secret generic additional-root-ca-cert \
     --from-literal=filename=ca.crt --from-file=data=ca.crt
-oc label secret/oscontainer-push-registry-secret \
+oc label secret/additional-root-ca-cert \
     jenkins.io/credentials-type=secretFile
-oc annotate secret/oscontainer-push-registry-secret \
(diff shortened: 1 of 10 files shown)


dustymabe (approved):
LGTM

PR #672, opened 2022-10-05
Add  upload_to_clouds function with multiple cloud configurations

ravanelli:


utils.groovy
```diff
@@ -282,4 +282,110 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
...
+                def family =  (pipecfg.clouds."${cloud}".family ? '--family ' + pipecfg.clouds."${cloud}".family: "")
+                def acl = (pipecfg.clouds."${cloud}".public ? "--public" : "")
+                def gcp_licenses = pipecfg.clouds."${cloud}".licenses
+                def licenses = gcp_licenses.collect{"--license " + it.replace('STREAM', stream)}.join(" ")
+                def today = shwrapCapture("date +%Y-%m-%d")
+                family = family.replace("STREAM", stream)
+                description = description.replace("BUILDID", buildID)
+                description = description.replace("STREAM", stream)
+                description = description.replace("BASEARCH", basearch)
+                description = description.replace("TODAY", today)
```
cgwalters, line 349: I'm fine with this as is.

But I'd still say that this kind of nontrivial logic ultimately should live in coreos-assembler and not the pipeline.

The GCP licenses in particular to me seems like a strong match for https://github.com/coreos/coreos-assembler/pull/3094

utils.groovy
```diff
@@ -282,4 +282,110 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
...
+                    --upload \
+                    --arch=${basearch} \
+                    --build=${buildID} \
+                    --region=${primary_region} ${grant_user_args} \
+                    --bucket ${bucket} \
+                    --credentials-file=\${${var_name}_BUILD_UPLOAD_CONFIG}
+                """)
+           }
+           if (cloud == 'azure') {
+               def container = pipecfg.clouds."${cloud}".container
```
dustymabe, line 322: in cases where we are inside an `if (cloud == 'value')` already I think we should remove `"${cloud}"`:

```suggestion
               def container = pipecfg.clouds.azure.container
```

config.yaml
```diff
@@ -72,12 +72,25 @@ aws_aarch64_serial_console_hack: true
...
   gcp:
     bucket: fedora-coreos-cloud-image-uploads
+    path: image-import
+    create_image: true
+    deprecated: true
+    description: "Fedora, Fedora CoreOS STREAM, BUILDID, BASEARCH published on TODAY"
+       family: "fedora-coreos-STREAM"
+       licenses:
+          - "fedora-coreos-STREAM"
+          - "https://compute.googleapis.com/compute/v1/projects/vm-options/global/licenses/enable-vmx
```
dustymabe, line 96: spacing here looks off? I would also maybe make `family:` `image_family:`

utils.groovy
```diff
@@ -282,4 +282,110 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
...
+                   --container ${container} \
+                   --profile \${AZURE_IMAGE_UPLOAD_CONFIG} \
+                   --resource-group ${resource_group} \
+                   --storage-account ${storage_account} \
+                   --force
+                """)
+           }
+           if (cloud == 'gcp') {
+                def create_image =  (pipecfg.clouds."${cloud}".create_image ? "--create-image=true": "")
+                def deprecated =  (pipecfg.clouds."${cloud}".deprecated ? "--deprecated": "")
```
dustymabe, line 338: Maybe here we can just assume `--deprecated` if the user provided an image family since that's the context where it makes sense.

utils.groovy
```diff
@@ -282,4 +282,110 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
...
+                def description =  (pipecfg.clouds."${cloud}".description ? ('--description=' + pipecfg.clouds."${cloud}".description): "")
+                def family =  (pipecfg.clouds."${cloud}".family ? '--family ' + pipecfg.clouds."${cloud}".family: "")
+                def acl = (pipecfg.clouds."${cloud}".public ? "--public" : "")
```
dustymabe, line 341: I don't think we pass this for FCOS but I think we don't need to because the project is configured to be public. Does RHCOS pass `--public` when uploading to GCP?

utils.groovy
```diff
@@ -282,4 +282,110 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
...
+                def gcp_licenses = pipecfg.clouds."${cloud}".licenses
+                def licenses = gcp_licenses.collect{"--license " + it.replace('STREAM', stream)}.join(" ")
+                def today = shwrapCapture("date +%Y-%m-%d")
+                family = family.replace("STREAM", stream)
+                description = description.replace("BUILDID", buildID)
```
dustymabe, line 346: I think @jlebon is going to create some templating function we can use here.

utils.groovy
```diff
@@ -282,4 +282,110 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
...
+                description = description.replace("STREAM", stream)
+                description = description.replace("BASEARCH", basearch)
+                description = description.replace("TODAY", today)
+                shwrap("""
+                # pick up the project to use from the config
+                # gcp_project=\$(jq -r .project_id \${GCP_IMAGE_UPLOAD_CONFIG})
+                gcp_project=aa
```
dustymabe, line 353:
```suggestion
                gcp_project=\$(jq -r .project_id \${GCP_IMAGE_UPLOAD_CONFIG})
```

Go back to what it was?

utils.groovy
```diff
@@ -282,4 +282,110 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
...
+    def meta = readJSON file: meta_json
+    def artifacts = meta['images'].keySet() as List
+    def clouds = pipecfg.clouds.keySet() as List
+    // Check if we have cloud config to upload these artifacts
+    def uploads = artifacts.intersect(clouds)
+    parallel uploads.inject([:]) { branch, cloud -> branch["Upload ${cloud}"] = {
+
+        def var_name = cloud.toUpperCase()
+        tryWithCredentials([file(variable: "${var_name}_BUILD_UPLOAD_CONFIG",
+                                 credentialsId: "${cloud}-build-upload-config")]) {
```
dustymabe, line 298: unfortunately this isn't 100% effective.

For AWS we use `aws-build-upload-config` but for all other clouds we've been using `"${cloud}-image-upload-config"`.

For AWS we're uploading the build artifacts (hence `aws-build-upload-config` name) and also an AMI. We just happened to re-use the secret for the AMI creation. For others we're only creating the images, so they were named appropriately.

dustymabe (commented):
some initial review.

I think this is a large enough change where we'll need to dig through it a little more and run some tests. Maybe next week.

utils.groovy
```diff
@@ -282,4 +282,110 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
...
+// Upload artifacts to clouds
+def upload_to_clouds(pipecfg, basearch, buildID, stream) {
+    // Get artifacts to upload
+    def meta_json = "builds/${buildID}/${basearch}/meta.json"
+    def meta = readJSON file: meta_json
+    def artifacts = meta['images'].keySet() as List
+    def clouds = pipecfg.clouds.keySet() as List
+    // Check if we have cloud config to upload these artifacts
+    def uploads = artifacts.intersect(clouds)
+    parallel uploads.inject([:]) { branch, cloud -> branch["Upload ${cloud}"] = {
```
jlebon, line 294: Just to note here some of the OOB discussions we had about this. I think rather than trying to make a single closure be generic, you can have multiple separate closures for each cloud. The idea I had was doing something like this:

```groovy
def uploaders = [:]
uploaders["aws"] = {
  ...
}
uploaders["gcp"] = {
  ...
}
...
// use actual artifacts list to filter out uploaders we don't need
// then:
parallel uploaders
```

Also, if possible I think it'd be good to move this to a dedicated module, e.g. `uploads.groovy`?

In the future, I think it'd be pretty cool to be able to just do `cosa buildextend-${cloud} --config config.yaml`. Then we could do something similar to what you originally had in mind here.

config.yaml
```diff
@@ -72,12 +72,25 @@ aws_aarch64_serial_console_hack: true
 
 clouds:
   aws:
+    upload_accounts:
+      - "013116697141"
```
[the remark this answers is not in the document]
ravanelli: For `rhcos` we have 2 accounts, one specifically for China

utils.groovy
ravanelli: Yes, it is  from `rhcos`

utils.groovy
ravanelli: Does it make sense to have 2 secrets to `aws` so? So we can match the names, or even change it to match the others?

utils.groovy
ravanelli: +1 I think the idea is to moved it as a followup in the future ;)

config.yaml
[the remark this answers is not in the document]
dustymabe: I think we'd need to represent that a different way here. i.e. we'd need to specifically name it here so we'd know which account was which, right? i.e. you'd need to specify a different region or endpoint or something for the AWS china or govcloud partition?

Even with that I don't think we need an entry here. The information can be implied from the credential I think.

utils.groovy
dustymabe: Yes, It could make sense to have a separate secret here, but I kind of think we should only do that if we plan to create AMIs in a different account that where we are doing the build upload. I guess maybe that's the case for China or GovCloud?

utils.groovy
ravanelli: >> I think rather than trying to make a single closure be generic, you can have multiple separate closures for each cloud
In this case we would lose the `tryWithCredentials` generic utilization and would need to call it for all + the generic utilization for `aws` and `azure` that are almost identical.  Is that what we want?

uploads.groovy
```diff
@@ -0,0 +1,144 @@
...
+    def artifacts = meta['images'].keySet() as List
+    def uploaders = [:]
+
+    artifacts.each {
+
+        def bucket = pipecfg.clouds."${it}".bucket
+        def path = pipecfg.clouds."${it}".path
+        def primary_region = pipecfg.clouds."${it}".primary_region
+
+        if (it == 'aws') {
```
jlebon, line 18: I think we could make this more elegant by instead doing something like:

```groovy
def uploaders = [:]
uploaders["aws"] = {
    ...
}
uploaders["gcp"] = {
    ...
}
...
uploaders.keySet().retainAll(artifacts)
parallel uploaders
```

dustymabe:
Since Renata will be AFK for some time I'll take over this PR and try to test and push updates to get it over the finish line.

uploads.groovy
```diff
@@ -0,0 +1,144 @@
...
+                    def family =  (pipecfg.clouds.gcp.family.image_family ? '--family ' + pipecfg.clouds.gcp.family.image_family + '--deprecated': "")
+                    def acl = (pipecfg.clouds.gcp.public ? "--public" : "")
+                    def gcp_licenses = pipecfg.clouds.gcp.licenses
+                    def licenses = gcp_licenses.collect{"--license " + it.replace('STREAM', stream)}.join(" ")
+                    def today = shwrapCapture("date +%Y-%m-%d")
+                    family = family.replace("STREAM", stream)
+                    description = description.replace("BUILDID", buildID)
+                    description = description.replace("STREAM", stream)
+                    description = description.replace("BASEARCH", basearch)
+                    description = description.replace("TODAY", today)
```
jlebon, line 92: Now with https://github.com/coreos/coreos-ci-lib/pull/119, this would be something like:

```suggestion
                    description = utils.substituteStr(description, [
                        BUILDID: buildID,
                        STREAM: stream,
                        BASEARCH: basearch,
                        TODAY: shwrapCapture("date +%Y-%m-%d")
                    ])
```

config.yaml
```diff
@@ -72,12 +72,26 @@ aws_aarch64_serial_console_hack: true
...
   gcp:
     bucket: fedora-coreos-cloud-image-uploads
+    path: image-import
+    create_image: true
+    deprecated: true
+    description: "Fedora, Fedora CoreOS STREAM, BUILDID, BASEARCH published on TODAY"
+    family:
+      image_family: "fedora-coreos-STREAM"
+    licenses:
+      - "fedora-coreos-STREAM"
```
jlebon, line 96: And the templated bits here need to use e.g. `${STREAM}`, `${BUILDID}`, etc...

dustymabe:
Closing in favor of #699.

PR #673, opened 2022-10-05
jobs/build: templatize source config ref to use

jlebon:
RHCOS uses a different naming scheme from FCOS. Support a new `ref` member under `source_config` and also per-stream overrides (for the `master` branch).

```yaml
source_config:
  url: https://github.com/openshift/os
  ref: release-${STREAM}

streams:
  4.12:
    source_config_ref: master
```


dustymabe (approved):
LGTM

PR #676, opened 2022-10-06
manifests/jenkins.yaml: use imagestream name for jenkins-agent

jlebon:
The imagestream already has the local lookupPolicy enabled, so this already works. That way we avoid hardcoding the registry name.


dustymabe (approved):
LGTM

PR #679, opened 2022-10-07
s390x: Add builder config for RHCOS

jschintag:
Add the builder config for the s390x RHCOS builder. Add ignition files for both so that they can be fetched during installation.

Signed-off-by: Jan Schintag <jan.schintag@de.ibm.com>


jschintag:
I am unsure wether or not we want the ignition files in here. They are needed for the LPAR installation, since i need to fetch them over http(s) and this is the easiest way for me. However i guess technically they count as artifacts and not source code.

dustymabe:
Thanks for working on this @jschintag - I'll look at it on Monday when I'm back.

jschintag:
No problem @dustymabe, i generally don't expect reviews on Fridays :)

jschintag:
Since i noticed this one was still open, i added the changes to initialize the secure execution volume during installation here.

dustymabe:
@jschintag I think for now let's leave the generated `*.ign` files out of the git repo and just update the README with further instructions. 

I understand why it's useful to have them there but I think it causes more problems than it solves. If we did end up wanting to store them in git I think we should probably front that with a github action or some CI to validate that the `*ign` files are consistent with the butane sources for each PR.

jschintag:
@dustymabe Yes, i wasn't sure about the ignition files either. I removed them.

>@jschintag I think for now let's leave the generated *.ign files out of the git repo and just update the README with further instructions.

Since we already have an internal how-to on how to deploy the RHCOS Builder, with instructions for the specific hardware, i think it does not make sense do add that bit here.

multi-arch-builders/coreos-s390x-rhcos-builder.bu
```diff
@@ -0,0 +1,124 @@
...
+      group:
+        name: root
+      contents:
+        inline: |
+            [Unit]
+            Description=Create secex-data volume
+            [Service]
+            Type=oneshot
+            ExecStart=/home/core/create-secex-data.sh
+    - path: /home/builder/.config/systemd/user/secex-data-keepalive.service
```
dustymabe, line 99: my understanding is that we can drop this service when https://github.com/containers/podman/issues/17051 is fixed and we update our filter for podman prune?

multi-arch-builders/coreos-s390x-rhcos-builder.bu
```diff
@@ -0,0 +1,124 @@
...
+            chown builder:builder "/var/home/builder/${TARBALL}"
+
+            if ! $(sudo -u builder -H /bin/bash -c "cd /var/home/builder; podman volume list" | grep -q secex-data); then
+                echo "Creating volume secex-data"
+                sudo -u builder -H /bin/bash -c "cd /var/home/builder; podman volume create --label=persistent secex-data"
+            fi
+
+            echo "Importing tarball into volume"
+            sudo -u builder -H /bin/bash -c "cd /var/home/builder; podman volume import secex-data /var/home/builder/${TARBALL}"
+    - path: /etc/systemd/secex-data-volume.service
```
dustymabe, line 86: since this is a system systemd unit (not a user unit) let's just use a [systemd directive](https://coreos.github.io/butane/examples/#systemd-units) rather than create the file manually. It has the benefit of being able to just set `enabled: true` without having to manually create the symlink below.

multi-arch-builders/coreos-s390x-rhcos-builder.bu
```diff
@@ -0,0 +1,124 @@
...
+# - Set a hostname
+# - Create and initialize the secex-data volume
+#
+variant: fcos
+version: 1.4.0
+ignition:
+  config:
+    merge:
+      - source: http://foo/bar/builder-common.ign
+passwd:
```
dustymabe, line 14: this passwd section is mostly duplicated with what's in `builder-common.bu` Maybe let's just add your SSH key to `builder-common.bu` and remove this?

multi-arch-builders/coreos-s390x-rhcos-builder.bu
jschintag: Yes, as soon as we have a podman with the fix, we can drop this and instead use something like `--filter label!=persistent`

multi-arch-builders/coreos-s390x-rhcos-builder.bu
```diff
@@ -0,0 +1,124 @@
...
+# - Allow the builder user to log in with the associated ssh key
+# - Set a hostname
+# - Create and initialize the secex-data volume
+#
+variant: fcos
+version: 1.4.0
+ignition:
+  config:
+    merge:
+      - source: http://foo/bar/builder-common.ign
```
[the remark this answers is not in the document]
jschintag: No, since in order to install it on LPAR, we need to fetch the files over http(s). That is why i have the placeholder here.

multi-arch-builders/coreos-s390x-rhcos-builder.bu
jschintag: Yes, i think this makes it easier to maintain and also makes `coreos-s390x-rhcos-builder.bu` simply a variant of `coreos-s390x-builder.bu`.
I updated the PR accordingly.

multi-arch-builders/coreos-s390x-rhcos-builder.bu
jschintag: I updated the PR accordingly

dustymabe:
I think in the latest update we lost setting the hostname to something different. Not sure if that was intended or not.

jschintag:
> I think in the latest update we lost setting the hostname to something different. Not sure if that was intended or not.

Yes, i lost it intentionally since it is not really needed to keep the hostname of the machine special.

multi-arch-builders/coreos-s390x-rhcos-builder.bu
```diff
@@ -0,0 +1,103 @@
+# This butane config will do the following:
+#
+# - Merge in the builder-common.ign Ignition file
+# - Allow the builder user to log in with the associated ssh key
+# - Set a hostname
+# - Create and initialize the secex-data volume
```
dustymabe, line 6: This needs updating:

```suggestion
# - Merge in the coreos-s390x-builder.ign Ignition file
# - Create and initialize the secex-data volume
```

Might have to be split across the two commits you have in this PR.

jschintag: I updated the description accordingly and split it between the 2 commits

dustymabe (approved):
LGTM

PR #680, opened 2022-10-07
jobs: get source config ref using new helper

jlebon:
We need the ref override logic everywhere that wants to clone the repo.


dustymabe (approved):
LGTM

PR #682, opened 2022-10-07
docs: document config.yaml schema

jlebon:
We're adding a lot of keys to `config.yaml` right now. We need to make sure we do a proper job of documenting them so that it's easier for others to make use of the file.

Let's document the keys we currently support and from now on require any new keys to be documented there.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -1,0 +1,80 @@
+# REQUIRED: knobs related to the source config
+source_config:
+  # REQUIRED: source config repo URL to build
+  url: https://github.com/jlebon/my-coreos
+  # OPTIONAL: source config ref to build
+  # Supports '${STREAM}' variable.
+  ref: release-${STREAM}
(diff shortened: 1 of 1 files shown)


docs/config.yaml
```diff
@@ -0,0 +1,80 @@
...
+      # This will be the default stream for jobs that take a 'STREAMS' parameter
+      default: true
+      # OPTIONAL: override source config ref to use for this stream
+      source_config_ref: main
+    stable:
+      type: production
+      # OPTIONAL: override cosa image to use for this stream
+      cosa_image: "quay.io/jlebon/coreos-assembler:stable"
+      # OPTIONAL: additional artifacts to build for this stream
+      additinal_artifacts:
```
dustymabe, line 42:
```suggestion
      additional_artifacts:
```

dustymabe: I pushed up a fix for this.

dustymabe (approved):
LGTM

PR #683, opened 2022-10-07
utils: add minimal string templating helper

jlebon:
We have a few fields in `config.yaml` which will need to support
templating. Add a helper to make this easier and use it in the one field
that currently needs it.

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -30,12 +30,23 @@ def load_pipecfg() {
     return readYaml(file: "pipecfg/config.yaml")
 }
 
+def substitute_str(str, stream, buildid = "") {
+    def date = shwrapCapture("date +%Y-%m-%d")
+    def basearch = shwrapCapture("cosa basearch")
+    def bindings = ["STREAM": stream, "ARCH": arch, "DATE": date]
+    if (buildid) {
+        bindings['BUILDID'] = buildid
+    }
+    def engine = new groovy.text.SimpleTemplateEngine()
+    def tmpl = engine.createTemplate(str).make(binding)
+    return tmpl.toString()
+}
+
 def get_source_config_ref_for_stream(pipecfg, stream) {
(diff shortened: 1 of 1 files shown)


jlebon:
Requires: https://github.com/coreos/fedora-coreos-pipeline/pull/680

Still need to test this.

utils.groovy
```diff
@@ -30,12 +30,23 @@ def load_pipecfg() {
     return readYaml(file: "pipecfg/config.yaml")
 }
 
+def substitute_str(str, stream, buildid = "") {
+    def date = shwrapCapture("date +%Y-%m-%d")
+    def basearch = shwrapCapture("cosa basearch")
```
dustymabe, line 35: This probably needs to be a parameter too because i.e. we'd want to be able to operate against `aarch64` even though we're on `x86_64` jenkins.

utils.groovy
```diff
@@ -30,12 +30,23 @@ def load_pipecfg() {
     return readYaml(file: "pipecfg/config.yaml")
 }
 
+def substitute_str(str, stream, buildid = "") {
```
dustymabe, line 33: is groovy like python where you can provider arguments by keyword? so if we had something like:

```
def substitute_str(str, stream = "", buildid = "", basearch = "", foo = "", bar = "", baz = "")
```

Then we could just call:

```
substitute_str("foo=${FOO}", foo="myfoo")
```

and it work without having to pass every possible template token?

jlebon:
Hmm, unfortunately this doesn't work because `SimpleTemplateEngine` has to be allowed in the list of Groovy scripts. We haven't had to declaratively do that via JCasC so far, and I'd rather not start having to track these exceptions.

Might just hack this up so we manually `replace` instead. :cry:

jlebon:
Moved this to https://github.com/coreos/coreos-ci-lib/pull/119.

PR #684, opened 2022-10-09
utils: fix rsyncing X.509 fedmsg cert directory

jlebon:
We need the trailing slash here so that rsync copies the contents of the directory rather than the directory itself into the new directory.


dustymabe (approved):
LGTM

PR #685, opened 2022-10-09
utils.groovy: Change `additional_artifacts` to `skip_artifacts`

ravanelli:
- Instead of using additional artifacts use skip artifacts, in this way the default artifact list will use the artifact list from the latest version and not the old.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -38,12 +38,12 @@ streams:
       type: production
       # OPTIONAL: override cosa image to use for this stream
       cosa_image: "quay.io/jlebon/coreos-assembler:stable"
-      # OPTIONAL: additional artifacts to build for this stream
-      additional_artifacts:
+      # OPTIONAL: skip some of the default artifacts for this stream
+      skip_artifacts:
         aarch64:
-          - azure
-        ppc64le:
-          - powervs
+          - openstack
+        x86_64:
+          - vmware
     rawhide:
       type: mechanical
 
--- a/utils.groovy
+++ b/utils.groovy
@@ -282,14 +282,14 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
     def artifacts = []
     if  (pipecfg.streams[stream].artifacts?."${basearch}" != null) {
         artifacts = pipecfg.streams[stream]['artifacts'][basearch]
-    } else { // Merge default with additional artifacts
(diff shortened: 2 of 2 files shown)


ravanelli:
> an open question here: do we think `additional_artifacts` is useful? should we support both `additional_artifacts` and `skip_artifacts`?

We can still support it, but we also have `artifacts` as a way to replace everything.

dustymabe:
This LGTM - will hold off on merge until this week's `next` release is done building.

dustymabe:
Can you update the docs now that we're changing it up:

https://github.com/coreos/fedora-coreos-pipeline/blob/da6e33edd58907b5fa7f40369e813403717220cc/docs/config.yaml#L42

docs/config.yaml
```diff
@@ -38,8 +38,8 @@ streams:
       type: production
       # OPTIONAL: override cosa image to use for this stream
       cosa_image: "quay.io/jlebon/coreos-assembler:stable"
-      # OPTIONAL: additional artifacts to build for this stream
-      additional_artifacts:
+      # OPTIONAL: skip artifacts for this stream
+      skip_artifacts:
         aarch64:
           - azure
         ppc64le:
```
dustymabe, line 45: Let's update the example to make more sense based on the rest of the file.

```suggestion
      # OPTIONAL: skip artifacts for this stream
      skip_artifacts:
        aarch64:
          - openstack
        x86_64:
          - vmware
```

dustymabe (approved):
LGTM

utils.groovy
```diff
@@ -282,14 +282,14 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
...
-        def additional_atifacts = pipecfg.streams[stream].additional_artifacts?."${basearch}"
+        def skip_artifacts = pipecfg.streams[stream].skip_artifacts?."${basearch}"
         if (default_artifacts != null) {
-            artifacts += default_artifacts
+            artifacts = default_artifacts
         }
-        if (additional_atifacts != null) {
-            artifacts += additional_atifacts
+        if (skip_artifacts != null) {
+            artifacts -= skip_artifacts
```
jlebon, line 292: Cool that this simply works!

jlebon (approved):
LGTM!

PR #686, opened 2022-10-10

dustymabe:
I think we can drop this now that we have `next-devel` on F37. We'll re-enable it if we find a good reason to.


ravanelli (approved):
LGTM

PR #687, opened 2022-10-11
jobs/{build,kola*}: bump memory requests

dustymabe:
Now that we set the memory limit as well we find out we're using more than we are requesting. Let's request a more appropriate amount of memory for our jobs.

For the build job we set it to 8.5 Gi because we run two testiso runs
at the same time and those can take up to 4Gi each.

For the kola-* jobs the request is updated to 1Gi if the job is going
to download and decompress the disk image artifact and 512Mi otherwise.

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -75,7 +75,7 @@ def strict_build_param = stream_info.type == "mechanical" ? "" : "--strict"
 // here; we can look into making them configurable through the template if
 // developers really need to tweak them (note that in the default minimal devel
 // workflow, only the qemu image is built).
-def cosa_memory_request_mb = 6.5 * 1024 as Integer
+def cosa_memory_request_mb = 8.5 * 1024 as Integer
 
 // Now that we've established the memory constraint based on xz above, derive
 // kola parallelism from that. We leave 512M for overhead and VMs are 1G each
--- a/jobs/kola-aws.Jenkinsfile
+++ b/jobs/kola-aws.Jenkinsfile
@@ -52,7 +52,7 @@ if (s3_stream_dir == "") {
 }
 
 try { timeout(time: 90, unit: 'MINUTES') {
-    cosaPod(memory: "256Mi", kvm: false,
+    cosaPod(memory: "512Mi", kvm: false,
             image: params.COREOS_ASSEMBLER_IMAGE) {
 
         stage('Fetch Metadata') {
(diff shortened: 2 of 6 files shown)


jobs/build.Jenkinsfile
```diff
@@ -75,7 +75,7 @@ def strict_build_param = stream_info.type == "mechanical" ? "" : "--strict"
 // here; we can look into making them configurable through the template if
 // developers really need to tweak them (note that in the default minimal devel
 // workflow, only the qemu image is built).
-def cosa_memory_request_mb = 6.5 * 1024 as Integer
+def cosa_memory_request_mb = 10 * 1024 as Integer
```
jlebon, line 78: Hmm, but this will increase the number of parallel kola tests. I think we want instead to flip it around so that we derive the memory request from `ncpus`? E.g.

```groovy
def ncpus = 6
def cosa_memory_request_mb = ((ncpus * 1024) + X) as Integer
```

And then solve for X since it sounds like 512M is actually not enough. :)

jlebon:
High-level, I think having more accurate memory requests is good and makes us better k8s citizens. An improvement in the future would be to split things up so that we only request peak memory for the bits that actually need it.

jobs/build.Jenkinsfile
dustymabe: the kola tests aren't actually the problem. it's the `testiso` tests which don't don't have a direct relationship with this memory request IIUC.

jlebon: https://github.com/coreos/fedora-coreos-pipeline/pull/688

dustymabe:
ok updated with context from https://github.com/coreos/fedora-coreos-pipeline/pull/688 incorporated.

jlebon (approved):

jobs/build.Jenkinsfile
```diff
@@ -75,7 +75,7 @@ def strict_build_param = stream_info.type == "mechanical" ? "" : "--strict"
 // here; we can look into making them configurable through the template if
 // developers really need to tweak them (note that in the default minimal devel
 // workflow, only the qemu image is built).
-def cosa_memory_request_mb = 6.5 * 1024 as Integer
+def cosa_memory_request_mb = 8.5 * 1024 as Integer
```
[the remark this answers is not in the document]
dustymabe: it may even be desirable :)

jlebon (approved):

PR #688, opened 2022-10-11
jobs/build: merge `metal` and `metal4k` testiso runs

jlebon:
The `uefi` branch runs very fast since we're just doing two lightweight scenarios. Merge it with the `metal4k` branch which currently is faster than the `metal` branch since the former runs less scenarios. This should reduce our peak memory requirements by 4G without affecting build times much.

Related: https://github.com/coreos/fedora-coreos-pipeline/pull/687


jlebon:
We should probably also look at setting more reasonable memory requirements in `testiso` based on what's actually being tested. `iso-live-login` for example should be fine with 1G of RAM.

dustymabe:
I'm not the biggest fan of this approach, but I think there are many improvements we need to make in this space and it's not worth nit picking on this right now. Will merge.

dustymabe (approved):
LGTM

jlebon:
Don't disagree. :)

Personally, one thing I'd like we investigate is getting rid of `kola testiso` and doing a similar move to external tests like we did for `kola run`. E.g. you could imagine something like:

```bash
#!/bin/bash
# kola: { "platforms": "qemu-iso", "minMemory": 4096, "additionalDisks": ["10G"] }
set -xeuo pipefail

. $KOLA_EXT_DATA/commonlib.sh

case "${AUTOPKGTEST_REBOOT_MARK:-}" in
  "")
      # run coreos-installer

      /tmp/autopkgtest-reboot installed
      ;;

  installed)
      # run checks

      ;;
  *) fatal "unexpected mark: ${AUTOPKGTEST_REBOOT_MARK}";;
esac
```

PR #689, opened 2022-10-11
tree: use `tryWithCredentials` from coreos-ci-lib

jlebon:
It was upstreamed in https://github.com/coreos/coreos-ci-lib/pull/117.

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -155,17 +155,6 @@ def triggered_by_push() {
     return (currentBuild.getBuildCauses('com.cloudbees.jenkins.GitHubPushCause').size() > 0)
 }
 
-// Runs closure if credentials exist, otherwise gracefully return.
-def tryWithCredentials(creds, Closure body) {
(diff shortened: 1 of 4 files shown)


utils.groovy
```diff
@@ -155,17 +155,6 @@ def triggered_by_push() {
     return (currentBuild.getBuildCauses('com.cloudbees.jenkins.GitHubPushCause').size() > 0)
 }
 
-// Runs closure if credentials exist, otherwise gracefully return.
-def tryWithCredentials(creds, Closure body) {
```
dustymabe, line 159: are the uses of `tryWithCredentials` in this file safe after this removal?

dustymabe (approved):
LGTM - one question

utils.groovy
jlebon: Yup, it's similar to our use of `shwrap` in the file.

PR #690, opened 2022-10-12
jobs/build: bump per-VM memory requirement to 1.5G

jlebon:
Some of the tests require 1.5G of RAM now due to
https://bugzilla.redhat.com/show_bug.cgi?id=1907030. We need to bump our calculation of parallelism accordingly in the absence of https://github.com/coreos/coreos-assembler/issues/3118. Otherwise depending on how kola schedules tests, we might go over the limit.


dustymabe (approved):
LGTM

PR #692, opened 2022-10-13
utils: use new substituteStr function for source_config

jlebon:
This was added to coreos-ci-lib in
https://github.com/coreos/coreos-ci-lib/pull/119.


dustymabe (approved):
LGTM

PR #695, opened 2022-10-14
jobs/{build,build-arch}: fix off by one issue with artifact builds

dustymabe:
We've structured our artifact builds into two runs and we've also gone to some effort to put the metal and metal4k builds into the first run and the live ISO build into the second run (since the ISO depends on the metal/metal4k). The way we had it before if you trimmed down your list such that you were only building metal/metal4k/live then you'd end up with metal in the first run and metal4k/live in the second run, which fails for obvious reasons. Let's flip that so that metal and metal4k will be in the first run.


dustymabe:
added comment!

jlebon (approved):
LGTM!

PR #698, opened 2022-10-20
Add support for pushing extensions container and legacy oscontainer

jlebon:
Those are two containers we need to push in RHCOS that's not in FCOS.
Expand the list of supported keys in `registry_repos` for those and
accordingly rework the logic in the release job to be more functional.

Make it automatically skip containers that weren't actually built (e.g.
`extensions` are not built on FCOS and RHCOS <= 4.11).

Support stream-specific additional `registry_repos` so that RHCOS <=
4.11 can specify the `registry.ci.openshift.org` location, which is no
longer needed in 4.12.

Add a new `add_build_tag` which is required to ensure that the pushed
images are not garbage-collected.

Use the new `cosa copy-container` which knows to "unmanifest list" when
copying from Quay.io to `registry.ci.openshift.org`.

### DIFF
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -151,31 +151,74 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
             }
         }
 
-        stage("Push OSContainer Manifest") {
-            // Ship a manifest list containing all requested architectures.
-            def oscontainer_registry_repo = pipecfg.registry_repos?.oscontainer
-            if (oscontainer_registry_repo) {
-                withCredentials([file(variable: 'REGISTRY_SECRET',
-                                      credentialsId: 'oscontainer-push-registry-secret')]) {
-                    def arch_args = basearches.collect{"--arch ${it}"}.join(" ")
-                    shwrap("""
-                    cosa push-container-manifest --auth=\${REGISTRY_SECRET} \
-                        --repo=${oscontainer_registry_repo} --tag=${params.STREAM} \
-                        --artifact=ostree --metajsonname=base-oscontainer \
-                        --build=${params.VERSION} ${arch_args}
-                    """)
-                }
-            }
-            def oscontainer_old_registry_repo = pipecfg.registry_repos?.oscontainer_old
-            if (oscontainer_old_registry_repo) {
-                withCredentials([file(credentialsId: 'oscontainer-secret', variable: 'REGISTRY_SECRET')]) {
-                    shwrap("""
-                    skopeo copy --all --authfile \$REGISTRY_SECRET \
-                        docker://${oscontainer_registry_repo}:${params.STREAM} \
-                        docker://${oscontainer_old_registry_repo}:${params.STREAM}
-                    """)
-                }
-            }
+        def registry_repos = pipeutils.get_registry_repos(pipecfg, params.STREAM)
+
+        // [config.yaml name -> [meta.json artifact name, meta.json toplevel name, tag suffix]]
+        // The config.yaml name is the name used in the `registry_repos` object. The
+        // meta.json artifact name is the "cosa name" for the artifact (in the `images`
+        // object). The top-level name is the key inserted with the pushed info. The tag
+        // suffix is needed to avoid the containers clashing in the OCP ART monorepo. It
+        // could be made configurable in the future. For now since FCOS doesn't need it and
+        // OCP ART doesn't actually care what the tag name is (it's just to stop GC), we
+        // hardcode it.
+        def push_containers = ['oscontainer': ['ostree', 'base-oscontainer', ''],
+                               'extensions': ['extensions-container', 'extensions-container', '-extensions'],
+                               'legacy_oscontainer': ['legacy-oscontainer', 'oscontainer', '-legacy']]
+
+        // filter out those not defined in the config
+        push_containers.keySet().retainAll(registry_repos.keySet())
+
+        // filter out those not built. this step makes the assumption that if an
+        // image isn't built for x86_64, then we don't upload it at all
+        def meta_x86_64 = readJSON file: "builds/${params.VERSION}/x86_64/meta.json"
+        def artifacts = meta_x86_64.images.keySet()
+
+        // in newer Groovy, retainAll can take a closure, which would be nicer here
+        for (key in (push_containers.keySet() as List)) {
+          if (!(push_containers[key][0] in artifacts)) {
+            push_containers.remove(key)
+          }
+        }
+
+        if (push_containers) {
+          stage("Push Containers") {
+            parallel push_containers.collectEntries{configname, val -> [configname, {
+              withCredentials([file(variable: 'REGISTRY_SECRET',
+                                    credentialsId: 'oscontainer-push-registry-secret')]) {
+                  def repo = registry_repos[configname]
+                  def (artifact, metajsonname, tag_suffix) = val
+                  def arch_args = basearches.collect{"--arch ${it}"}.join(" ")
+                  def tag_args = " --tag=${params.STREAM}${tag_suffix}"
+                  if (registry_repos.add_build_tag) {
+                      tag_args += " --tag ${params.VERSION}${tag_suffix}"
+                  }
+                  shwrap("""
+                  cosa push-container-manifest --auth=\${REGISTRY_SECRET} \
+                      --repo=${repo} ${tag_args} \
+                      --artifact=${artifact} --metajsonname=${metajsonname} \
+                      --build=${params.VERSION} ${arch_args}
+                  """)
+
+                  def old_repo = registry_repos["${configname}_old"]
+                  if (old_repo) {
+                    // a separate credential for the old location is optional; we support it
+                    // being merged as part of oscontainer-push-registry-secret
+                    pipeutils.tryWithOrWithoutCredentials([file(variable: 'OLD_REGISTRY_SECRET',
+                                                                credentialsId: 'oscontainer-push-old-registry-secret')]) {
+                      def authArg = "--authfile=\${REGISTRY_SECRET}"
+                      if (env.OLD_REGISTRY_SECRET) {
+                        authArg += " --dest-authfile=\${OLD_REGISTRY_SECRET}"
+                      }
+                      shwrap("""
+                      cosa copy-container ${authArg} ${tag_args} \
+                          --manifest-list-to-arch-tag=auto \
+                          ${repo} ${old_repo}
+                      """)
+                    }
+                  }
+              }
+            }]}
+          }
         }
 
         stage('Publish') {
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -168,9 +168,6 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
             }
             def oscontainer_old_registry_repo = pipecfg.registry_repos?.oscontainer_old
             if (oscontainer_old_registry_repo) {
-                // For a period of time let's also mirror into the old location too
-                // Drop this after October 2022. See
-                // https://discussion.fedoraproject.org/t/updated-registry-location-for-fedora-coreos-ostree-native-container/42740
                 withCredentials([file(credentialsId: 'oscontainer-secret', variable: 'REGISTRY_SECRET')]) {
                     shwrap("""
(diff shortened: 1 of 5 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/3130
Also, I've tested this in parts but not as a whole.

utils.groovy
```diff
@@ -282,4 +282,12 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
     }
     return artifacts.unique()
 }
+
+def get_registry_repos(pipecfg, stream) {
+    def registry_repos = pipecfg.registry_repos ?: [:]
```
dustymabe, line 287: what does `?:` do?

jobs/release.Jenkinsfile
```diff
@@ -152,34 +152,64 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
+        // in newer Groovy, retainAll can take a closure, which would be nicer here
+        for (key in (push_containers.keySet() as List)) {
+          if (!(push_containers[key][0] in artifacts)) {
+            push_containers.remove(key)
+          }
+        }
+
+        if (push_containers) {
+          stage("Push Containers") {
+            parallel container_types.collectEntries{configname, val -> [key, {
```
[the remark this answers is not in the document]
dustymabe: also I feel like this syntax is starting to get a little advanced for the "easy read".

jobs/release.Jenkinsfile
```diff
@@ -152,34 +152,64 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
+              withCredentials([file(variable: 'REGISTRY_SECRET',
+                                    credentialsId: 'oscontainer-push-registry-secret')]) {
```
dustymabe, line 179: assuming here that we'll be using the same secret for all probably won't hold forever. I guess if it is true now then let's go with it.

It also is a little weird to call it the `oscontainer-push-registry-secret` when we have extensions in there too.

jobs/release.Jenkinsfile
```diff
@@ -152,34 +152,64 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
+                  if (old_repo) {
+                    tryWithOrWithoutCredentials([file(variable: 'OLD_REGISTRY_SECRET',
+                                                      credentialsId: 'oscontainer-push-old-registry-secret')]) {
+                      def authArg = "--authfile=\${REGISTRY_SECRET}"
+                      if (env.OLD_REGISTRY_SECRET) {
+                        authArg += " --dest-authfile=\${OLD_REGISTRY_SECRET}"
+                      }
+                      shwrap("""
+                      cosa copy-container ${authArg} ${tag_args} \
+                          --arch-tag-manifest-fallback \
```
dustymabe, line 205:
```suggestion
                          --arch-tag-manifest-fallback \
                          --manifest-list-to-arch-tag=auto \
```

jobs/release.Jenkinsfile
```diff
@@ -152,34 +152,64 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
+                               'extensions': ['extensions-container', 'extensions-container', 'extensions'],
+                               'legacy_oscontainer': ['ostree', 'oscontainer', 'legacy']]
+
+        // filter out those not defined in the config
+        push_containers.keySet().retainAll(registry_repos.keySet())
+
+        // filter out those not built. this step makes the assumption that if an
+        // image isn't built for x86_64, then we don't upload it at all
+        def artifacts = readJSON(text: shwrapCapture("cosa meta --dump")).images.keySet()
+        // in newer Groovy, retainAll can take a closure, which would be nicer here
```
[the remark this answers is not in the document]
jlebon: Looks like it works for collections but not maps. And yup, compare the "Since" in https://docs.groovy-lang.org/latest/html/groovy-jdk/java/util/Collection.html#retainAll(groovy.lang.Closure) vs https://docs.groovy-lang.org/latest/html/groovy-jdk/java/util/Map.html#retainAll(groovy.lang.Closure).

jobs/release.Jenkinsfile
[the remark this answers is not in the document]
jlebon: Hmm, the form is similar to what we use in other places, like `build-cosa` and `bump-lockfile`.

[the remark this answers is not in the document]
dustymabe:
```
collectEntries{configname, val -> [key,
```

where is `key` defined?

jobs/release.Jenkinsfile
jlebon: Yup, agreed. I'd rather not add support for it until we actually need it.

Re. naming, let's keep that for a follow-up?

utils.groovy
jlebon: It's like the GCC extension ternary operator (though pretty sure most mainstream C compilers support it too by now): https://gcc.gnu.org/onlinedocs/gcc/Conditionals.html#Conditionals

jlebon:
Still need to test this.

utils.groovy
dustymabe: oh wow.. I've only ever used the `x ? x : y` form. I didn't know `x ? : y` existed. How did I not know this?

For future me could we just the longer form?

jobs/release.Jenkinsfile
dustymabe: > Re. naming, let's keep that for a follow-up?

That's fair, but note we will have to updated the existing `oscontainer-secret` in our pipeline if this merges as is.

jobs/release.Jenkinsfile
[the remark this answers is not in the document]
dustymabe: :) - I just assumed there was some syntax I didn't understand.

jobs/release.Jenkinsfile
```diff
@@ -152,34 +152,64 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
+
+        if (push_containers) {
+          stage("Push Containers") {
+            parallel container_types.collectEntries{configname, val -> [key, {
+              withCredentials([file(variable: 'REGISTRY_SECRET',
+                                    credentialsId: 'oscontainer-push-registry-secret')]) {
+                  def repo = registry_repos[configname]
+                  def (artifact, metajsonname, tag_suffix) = val
+                  def arch_args = basearches.collect{"--arch ${it}"}.join(" ")
+                  tag_suffix = suffix ? "-${suffix}" : "";
```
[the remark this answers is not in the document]
dustymabe: Not sure this question was answered. I see you deleted a line of code but the variable is still used and the data map still has an entry for a suffix.

jobs/release.Jenkinsfile
```diff
@@ -152,34 +152,64 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
-                    shwrap("""
-                    skopeo copy --all --authfile \$REGISTRY_SECRET \
-                        docker://${oscontainer_registry_repo}:${params.STREAM} \
-                        docker://${oscontainer_old_registry_repo}:${params.STREAM}
-                    """)
-                }
-            }
+        def registry_repos = pipeutils.get_registry_repos(pipecfg, params.STREAM)
+
+        // [config.yaml name -> [meta.json 'images' name, meta.json toplevel name, tag suffix]]
```
dustymabe, line 157: This comment above could probably be expanded. Why is it needed? What is all this info, etc.. What you have is a brief version, but doesn't fill in all the clues I don't think.

utils.groovy
dustymabe: if not could we at least separate the `?:` into `? :` ?

jobs/release.Jenkinsfile
jlebon: Yes indeed. Planning to coordinate that with the merge.

jobs/release.Jenkinsfile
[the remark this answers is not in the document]
jlebon: I added details for this in the comment. Does that help?

utils.groovy
jlebon: Trust me, once you use `?:`, you'll want to keep using it. :) It's a really nice QOL code improvement. The canonical way to type it [in most languages](https://en.wikipedia.org/wiki/Elvis_operator) is `?:` and that's how Groovy [documents it](http://groovy-lang.org/operators.html#_elvis_operator).

jobs/release.Jenkinsfile
```diff
@@ -152,34 +152,70 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
-            }
+        def registry_repos = pipeutils.get_registry_repos(pipecfg, params.STREAM)
+
+        // [config.yaml name -> [meta.json artifact name, meta.json toplevel name, tag suffix]]
+        // The config.yaml name is the name used in the `registry_repos` object. The
+        // meta.json artifact name is the "cosa name" for the artifact (in the `images`
+        // object). The top-level name is the key inserted with the pushed info. The tag
+        // suffix is needed to avoid the containers clashing in the ART monorepo. It could
+        // be made configurable in the future. For now since FCOS doesn't need it and ART
+        // doesn't actually care what the tag name is (it's just stop GC), we hardcode it.
```
dustymabe, line 163:
```suggestion
        // doesn't actually care what the tag name is (it's just to stop GC), we hardcode it.
```

jobs/release.Jenkinsfile
```diff
@@ -152,34 +152,70 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
-                }
-            }
+        def registry_repos = pipeutils.get_registry_repos(pipecfg, params.STREAM)
+
+        // [config.yaml name -> [meta.json artifact name, meta.json toplevel name, tag suffix]]
+        // The config.yaml name is the name used in the `registry_repos` object. The
+        // meta.json artifact name is the "cosa name" for the artifact (in the `images`
+        // object). The top-level name is the key inserted with the pushed info. The tag
+        // suffix is needed to avoid the containers clashing in the ART monorepo. It could
+        // be made configurable in the future. For now since FCOS doesn't need it and ART
```
dustymabe, line 162:
```suggestion
        // be made configurable in the future. For now since FCOS doesn't need it and = RHCOS ART
```

jobs/release.Jenkinsfile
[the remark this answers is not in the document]
dustymabe: Yes it does. Thanks!

jobs/release.Jenkinsfile
dustymabe: I think you did this.

dustymabe:
Looking good! - Let me know when you want another pass after testing.

jlebon:
OK, finally tested this successfully in stg!

jobs/release.Jenkinsfile
```diff
@@ -151,34 +151,71 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
+                  shwrap("""
+                  cosa push-container-manifest --auth=\${REGISTRY_SECRET} \
+                      --repo=${repo} ${tag_args} \
+                      --artifact=${artifact} --metajsonname=${metajsonname} \
+                      --build=${params.VERSION} ${arch_args}
+                  """)
+
+                  def old_repo = registry_repos["${configname}_old"]
+                  if (old_repo) {
+                    pipeutils.tryWithOrWithoutCredentials([file(variable: 'OLD_REGISTRY_SECRET',
```
dustymabe, line 204: This `tryWithOrWithoutCredentials` could probably use a comment about the circumstances.

jobs/release.Jenkinsfile
```diff
@@ -151,34 +151,71 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
+                                                                credentialsId: 'oscontainer-push-old-registry-secret')]) {
+                      def authArg = "--authfile=\${REGISTRY_SECRET}"
+                      if (env.OLD_REGISTRY_SECRET) {
+                        authArg += " --dest-authfile=\${OLD_REGISTRY_SECRET}"
+                      }
+                      shwrap("""
+                      cosa copy-container ${authArg} ${tag_args} \
+                          ${repo} ${old_repo}
```
dustymabe, line 212: Purely optional: Maybe we should still specify `--manifest-list-to-arch-tag=auto` here even though it's the default just to give a clue to the user something special is going on?

dustymabe (approved):
LGTM

PR #701, opened 2022-10-21
utils: add build_artifacts() function; use it

dustymabe:
Hoist the code used to build the secondary artifacts into our
pipeline utils.groovy file so we can share that code between the
two main jobs. Also rework it a bit to use a different maxRuns
heuristic to achieve the same goal it had before.


dustymabe:
the second commit here is something we talked about this morning @jlebon. I can squash it into the first commit if you think that would be best.

jlebon (approved):

PR #703, opened 2022-10-21
rename to libupload; add emojis to stage names

dustymabe:
```
commit 20c5e875fc8f2f0bf8fccbb9746a7b1eae1b4da0
Date:   Fri Oct 21 16:46:38 2022 -0400

    rename uploads.groovy to libupload.groovy
    
    This is mostly just a preference. `libupload.upload_to_clouds()` reads
    a little better than `uploads.upload_to_clouds()`.

commit a7c1fd256076a5e447e36063b861d098180be19e
Date:   Fri Oct 21 10:10:12 2022 -0400

    add emojis to cloud uploads and build artifacts parallel stage names
    
    Since there are multiple runs sometimes the individual parallel stages
    get separated from the stage title. Add an extra indicator to let us
    know which parallel stages are grouped together.

```


utils.groovy
```diff
@@ -301,7 +301,7 @@ def build_artifacts(pipecfg, stream, basearch) {
 
     // Define the parallel jobs in a map
     parallelruns = artifacts.collectEntries {
-        [it, { shwrap("cosa buildextend-${it}") }]
+        ["💽:${it}", { shwrap("cosa buildextend-${it}") }]
```
jlebon, line 304: Minor: not sure if it's the Firefox rendering, but here there's no space before the colon while in the cloud upload there is.

jlebon (approved):

utils.groovy
dustymabe: yeah I think it's the choice of emoji. The intention is for the emojis to not have space between them. Let me do a fixup for the other ones in libupload.groovy.

dustymabe: well then it makes it look weird in VIM. I'm just going to leave it like it is. Let me know how it looks on the other end if we want to change it.

PR #705, opened 2022-10-22
jobs/{build,build-arch}: fix AMI detection logic for cloud test

dustymabe:
The key in the meta.json is `amis`, not `aws`.

Fixes 0779a76.

Also add emoji to build-arch job aws build stage.

PR #707, opened 2022-10-22
factor out cloud test kickoff into run_cloud_tests() function

dustymabe:
```
commit 197e3cd3f5f7595e1974e7607c5f91dd8c91e28e
Date:   Fri Oct 21 23:12:32 2022 -0400

    config: add cloud.test_architectures key
    
    This key is used to limit which architecures we attempt to run cloud
    tests for on a given cloud/platform. For example, we create OpenStack
    artifacts for every architecture, but we don't have an OpenStack that
    we can test in that has all of those architectures so we need to limit
    the architectures that those tests are attempted on.

commit 8648df677b414e181e0997d2b2f1e4900b17f4d8
Date:   Fri Oct 21 22:35:34 2022 -0400

    factor out cloud test kickoff into run_cloud_tests() function
    
    This code is now pretty much duplicated. Let's factor it out and
    simplify it.

```

Requires #705

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -315,6 +315,20 @@ def get_registry_repos(pipecfg, stream) {
     return registry_repos
 }
 
+// Determine if the config.yaml has a test_architectures entry for
+// this cloud which is intended to limit the architectures that
+// are tested for the given cloud.
+def cloud_testing_enabled_for_arch(pipecfg, cloud, basearch) {
+    if (pipecfg.clouds."${cloud}"?.test_architectures) {
+        // The list exists. Return true/false if the arch is in the list.
+        return basearch in pipecfg.clouds."${cloud}".test_architectures
+    } else {
+        // If no test_architectures defined for cloud then default
+        // to running on all architectures (no limits).
+        return true
+    }
+}
+
 
 // Runs followup cloud tests based on conditions
 def run_cloud_tests(pipecfg, stream, version, s3_stream_dir, basearch, commit) {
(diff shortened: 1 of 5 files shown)


dustymabe:
This should prevent the issue where we have cloud tests running on architectures that aren't possible.

dustymabe:
I still need to test this.

dustymabe:
OK This has been tested now.

utils.groovy
```diff
@@ -315,4 +315,54 @@ def get_registry_repos(pipecfg, stream) {
     return registry_repos
 }
 
+
+// Runs followup could tests based on conditions
```
jlebon, line 337: cloud

utils.groovy
```diff
@@ -315,9 +315,27 @@ def get_registry_repos(pipecfg, stream) {
...
+        // If no test_architectures defined for cloud then default
+        // to running on all architectures (no limits).
+        return true
+    }
+}
+
 
 // Runs followup could tests based on conditions
-def run_cloud_tests(stream, version, s3_stream_dir, basearch, commit) {
+def run_cloud_tests(pipecfg, stream, version, s3_stream_dir, basearch, commit) {
```
jlebon, line 338: This should be in the first commit.

utils.groovy
```diff
@@ -315,9 +315,27 @@ def get_registry_repos(pipecfg, stream) {
...
+        if (basearch in pipecfg.clouds."${cloud}".test_architectures) {
+            return true
+        } else {
+            return false
+        }
+    } else {
+        // If no test_architectures defined for cloud then default
+        // to running on all architectures (no limits).
+        return true
+    }
```
jlebon, line 333: This can be:

```groovy
return basearch in pipecfg.clouds."${cloud}"?.test_architectures ?: []
```

dustymabe: I'm not quite sure it's equivalent. I think this would return false if `test_architectures` wasn't defined for a cloud, which isn't what we want. I did simplify this a little in the next upload, though.

jlebon (approved):
Nice!

PR #708, opened 2022-10-22
utils: set maxRuns to 10 when not building `live` artifacts

dustymabe:
In the case where we're not building live artifacts we were passing in a maxRuns value of `null` to the utils.ParallelRuns() function. Let's add a condition such that we'll set it to 10 if maxRuns is still uninitialized (`null`).

Fixes 8af6653


bgilbert (approved):
LGTM

PR #710, opened 2022-10-24
config.yaml: support artifacts in common for all arches

jlebon:
Every arch supports metal, metal4k, live, and openstack artifacts.
Instead of repeating it four times, add support for a generic "all" key
which is merged on all arches.

Prep for adding more artifacts in that key.

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -248,6 +248,7 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
     if (artifacts == null) {
         // Get the list difference from default with skip artifacts
         artifacts = pipecfg.default_artifacts."${basearch}" ?: []
+        artifacts += pipecfg.default_artifacts.all ?: []
         artifacts -= pipecfg.streams[stream].skip_artifacts?."${basearch}" ?: []
     }
     return artifacts.unique()
--- a/utils.groovy
+++ b/utils.groovy
@@ -243,18 +243,12 @@ def get_push_trigger() {
 
 // Gets desired artifacts to build from pipeline config
 def get_artifacts_to_build(pipecfg, stream, basearch) {
-    def artifacts = []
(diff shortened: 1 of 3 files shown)


dustymabe:
I'm not 100% sure we should take it to the next level here. Is it so bad to have them replicated?

I think it's easier when hacking to just consider and trim down the list for the architecture I care about limiting. This makes it a little more complicated to do that.

Other than dedupe is there other benefits to doing this way?

utils.groovy
```diff
@@ -243,18 +243,12 @@ def get_push_trigger() {
...
-        if (default_artifacts != null) {
-            artifacts = default_artifacts
-        }
-        if (skip_artifacts != null) {
-            artifacts -= skip_artifacts
-        }
+    // Explicit stream-level override takes precedence
+    def artifacts = pipecfg.streams[stream].artifacts?."${basearch}"
+    if (artifacts == null) {
+      // Get the list difference from default with skip artifacts
```
dustymabe, line 249: spacing of this comment is off

dustymabe (approved):
One open question and one suggested change.

jlebon:
Yeah I can see both ways. What prompted me to do this was that https://github.com/coreos/fedora-coreos-pipeline/pull/711 adds yet more artifacts to that list. You're right that it makes limiting a specific artifact harder, but I think it's going to be very rare the case where we need to limit e.g. the metal and live images for a specific architecture (on that note, I can see not putting `openstack` in that list). I think also when it does happen, it would be a stream-specific issue and not necessarily global, so then a better tool would be `skip_artifacts`. So this is optimizing for the common case to make the config more legible which has value too.

jlebon:
Now tested!

dustymabe (approved):
LGTM

PR #711, opened 2022-10-24
[rhcos] config.yaml: build legacy-oscontainer and extensions-container

jlebon:
These are additional containers we need to ship in RHCOS. Add them.

Requires: https://github.com/coreos/coreos-assembler/pull/3133
Requires: https://github.com/coreos/fedora-coreos-pipeline/pull/710


dustymabe (approved):
LGTM

PR #713, opened 2022-10-25
jobs: set xz memory limit when (de)compressing

jlebon:
We hit against memory limits in the `kola-openstack` job where we
decompress the image. We need the same memory limiting we do in the main
`build` job.


jlebon:
Though we should also investigate whether memory limits can be queried via cgroupsfs. Then we could lower this into cosa directly like we did with CPU count.

jobs/build-arch.Jenkinsfile
```diff
@@ -374,9 +374,10 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
         }
 
         stage('Archive') {
-            shwrap("""
-            cosa compress --compressor xz
-            """)
+            // lower to make sure we don't go over and account for overhead
+            pipeutils.withXzMemLimit(cosa_memory_request_mb - 256) {
+                shwrap("cosa compress --compressor xz")
+            }
```
dustymabe, line 380: This is technically running on the remote. Should we still apply the limit? If we do still apply the limit should we call out the elephant in the room (`"Yes, this is running on the remote, but we limit anyway because ..."`).

utils.groovy
```diff
@@ -378,4 +378,10 @@ def run_cloud_tests(pipecfg, stream, version, s3_stream_dir, basearch, commit) {
     parallel testruns
 }
 
+def withXzMemLimit(limitMi, Closure body) {
```
dustymabe, line 381: Maybe we should add a comment to this function about why it's needed so often.

dustymabe (approved):
LGTM - Added some suggestions for comments.

jobs/build-arch.Jenkinsfile
jlebon: Oh yes, good catch. I dropped that hunk. I think it could make sense to apply it to the remote to be considerate to other workloads, but would require doing something like `cosa shell -- env XZ_DEFAULTS=... cosa compress`? And the limit itself doesn't have to match what we use for the pod request of course.

dustymabe:
reading the man page on this I see:

```
If  the  specified  memory  usage limit is exceeded when decompressing,
xz will display an error and decompressing the file will fail. If the
limit is exceeded when compressing, xz will try to scale the settings
down so that the limit is no longer exceeded.
```

Does that mean that setting the memlimit for decompression doesn't
really do anything other than have XZ OOM itself?

bgilbert:
Yes, as I understand it xz RAM requirements during decompression are solely a property of how the file was encoded.

jlebon:
> Does that mean that setting the memlimit for decompression doesn't really do anything other than have XZ OOM itself?

Thanks, that's an important detail. Follow-up in https://github.com/coreos/fedora-coreos-pipeline/pull/717.

PR #714, opened 2022-10-25
utils: lower artifact building parallelism to 8

jlebon:
We're hitting up against memory limits when building the legacy oscontainer in the RHCOS pipeline. Before bumping our request higher, I'd like to try first reducing slightly our parallelism. (Worth noting that pipeline currently builds 14 artifacts, so this will still keep us within 2 parallel runs, but better distributed.)

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -265,7 +265,8 @@ def build_artifacts(pipecfg, stream, basearch) {
     //
     // 1. The `live` build depends on `metal` and `metal4k` artifacts to exist.
     // 2. We can't start too many parallel processes because we bump up into
-    //    PID limits in the pipeline. See 24c5265.
+    //    PID limits in the pipeline. See 24c5265. It also makes it easier to
+    //    bump up against memory limits.
 
     // For 1. we'll sort the artifact list such that `metal` and `metal4k`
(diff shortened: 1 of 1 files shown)


dustymabe:
> (Worth noting that pipeline currently builds 14 artifacts, so this will still keep us within 2 parallel runs, but better distributed.)

hmm. I count 16, but that doesn't change the accuracy or your statement.

jlebon:
> > (Worth noting that pipeline currently builds 14 artifacts, so this will still keep us within 2 parallel runs, but better distributed.)
> 
> hmm. I count 16, but that doesn't change the accuracy or your statement.

Yeah, I was strictly talking about the artifacts build by that stage, so not counting `ostree` and `qemu`.

dustymabe:
Looking at the [most recent build](https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/blue/organizations/jenkins/build/detail/build/442/pipeline) I still see `16` unless I can't count any longer:

Screenshot 2022-10-25 at 11-09-21 jenkins _ build _ #442

dustymabe (approved):
LGTM

PR #715, opened 2022-10-25
libupload: a few fixes for corner cases

dustymabe:
```
commit 6413b25f8ae3bb8fd995d3913e8a65b96dfb4ff5
Date:   Tue Oct 25 14:30:11 2022 -0400

    libupload: don't buildextend with --force for azure
    
    If we make it to the cloud uploads in the pipeline then we've
    reserved a build ID and any new runs will reserved a different
    build ID. We shouldn't need to pass `--force` here.
    
    In it's current form we're building the azure image in the
    "Build Artifacts" stage and again in the "Cloud Upload" stage
    befause `--force` tells `buildextend-azure` to ignore the
    existing built artifact.

commit bbf300c3b011f584b9eaa9b5502ae086533dc805
Date:   Tue Oct 25 14:28:02 2022 -0400

    libcloud: avoid null pointer exception
    
    If there are no AWS test accounts specified then the `.join()` will
    end up getting called on a null object. Let's check first if there
    is an entry for test accounts before we try to collect/join.

```

### DIFF
--- a/libupload.groovy
+++ b/libupload.groovy
@@ -41,7 +41,10 @@ def upload_to_clouds(pipecfg, basearch, buildID, stream) {
             withCredentials(creds) {
                 utils.syncCredentialsIfInRemoteSession(["AWS_BUILD_UPLOAD_CONFIG"])
                 def c = pipecfg.clouds.aws
-                def grant_user_args = c.test_accounts?.collect{"--grant-user ${it}"}.join(" ")
+                def grant_user_args = ""
+                if (c.test_accounts) {
+                    grant_user_args = c.test_accounts.collect{"--grant-user ${it}"}.join(" ")
+                }
                 shwrap("""
                 cosa buildextend-aws \
                     --upload \
--- a/libupload.groovy
+++ b/libupload.groovy
@@ -75,8 +75,7 @@ def upload_to_clouds(pipecfg, basearch, buildID, stream) {
                     --build=${buildID} \
                     --resource-group ${c.resource_group} \
                     --storage-account ${c.storage_account} \
-                    --container=${c.storage_container} \
-                    --force
+                    --container=${c.storage_container}
                  """)
             }
         }


libupload.groovy
```diff
@@ -41,7 +41,10 @@ def upload_to_clouds(pipecfg, basearch, buildID, stream) {
             withCredentials(creds) {
                 utils.syncCredentialsIfInRemoteSession(["AWS_BUILD_UPLOAD_CONFIG"])
                 def c = pipecfg.clouds.aws
-                def grant_user_args = c.test_accounts?.collect{"--grant-user ${it}"}.join(" ")
+                def grant_user_args = ""
+                if (c.test_accounts) {
+                    grant_user_args = c.test_accounts.collect{"--grant-user ${it}"}.join(" ")
```
gursewak1997, line 46: Not super familiar with Groovy but shouldn't this be handled by null safe operator (?.) that we already used.
Or maybe adding an additional null safe operator 
`def grant_user_args = c.test_accounts?.collect{"--grant-user ${it}"}?.join(" ")`

jlebon (approved):

libupload.groovy
jlebon: Good point. Thanks for raising this. The Groovy documentation is [not very thorough](http://groovy-lang.org/operators.html#_safe_navigation_operator) around this operator, but I think the reason why the previous code doesn't work is because `.` and `?.` have equal precedence: http://groovy-lang.org/operators.html#_operator_precedence. So it went left to right, converting `c.test_accounts?.collect{}` to `null`.

I think your suggestion would work, but the problem is that we don't want to render a literal `"null"` in the `cosa buildextend-aws` below. Interestingly, `null.collect{}` actually works fine (returns an empty list), so we don't even need the first `?`. A slight tweak on your suggestion then is:

```groovy
def grant_user_args = c.test_accounts.collect{"--grant-user ${it}"}?.join(" ")
```

I.e. relying on `null.collect{}` returning an empty list. But this is so subtle that it's probably worth the more verbose version.

PR #717, opened 2022-10-26
jobs/kola-{azure,openstack}: bump memory request

jlebon:
As brought up in #713, setting a memory limit for xz decompression doesn't cause it to adapt the limit, but rather to automatically abort if the limit is reached.

Keep limiting it because the error mode is clearer to debug than otherwise, but also increase the memory request since we know 1Gi is currently not enough.


dustymabe (approved):
LGTM

PR #719, opened 2022-10-26
Replace coreos-assembler ImageStream by new knobs in the pipecfg

jlebon:
See individual commits.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -21,6 +21,10 @@ default_artifacts:
     - aws
     - vmware
 
+# OPTIONAL: coreos-assembler image to build with. supports ${STREAM} variable.
+# If unset, defaults to `quay.io/coreos-assembler/coreos-assembler:main`.
+cosa_img: quay.io/jlebon/coreos-assembler:pr3139-${STREAM}
+
 # REQUIRED: streams to build
 streams:
     testing-devel:
@@ -49,6 +53,8 @@ streams:
       additional_registry_repos:
         oscontainer_old: registry.ci.openshift.org/rhcos-devel/machine-os-oci-content
         legacy_oscontainer_old: registry.ci.openshift.org/rhcos-devel/machine-os-content
+      # OPTIONAL: override default coreos-assembler image to build with
+      cosa_img: quay.io/coreos-assembler/coreos-assembler:stable
     rawhide:
       type: mechanical
 
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -55,6 +55,11 @@ streams:
         legacy_oscontainer_old: registry.ci.openshift.org/rhcos-devel/machine-os-content
       # OPTIONAL: override default coreos-assembler image to build with
       cosa_img: quay.io/coreos-assembler/coreos-assembler:stable
+      # OPTIONAL/TEMPORARY: Override default coreos-assembler image to use for
+      # the controller pod in the build-arch job. This is only useful when
+      # building with a cosa too old to support `cosa remote-session`, which
+      # should only be the case for RHCOS 4.11 and older.
+      cosa_controller_img_hack: quay.io/coreos-assembler/coreos-assembler:latest
     rawhide:
       type: mechanical
 
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -46,7 +46,7 @@ properties([
                    description: 'Force AWS AMI replication for non-production'),
       string(name: 'COREOS_ASSEMBLER_IMAGE',
              description: 'Override coreos-assembler image to use',
-             defaultValue: "coreos-assembler:main",
+             defaultValue: "quay.io/coreos-assembler/coreos-assembler:main",
              trim: true),
       booleanParam(name: 'KOLA_RUN_SLEEP',
                    defaultValue: false,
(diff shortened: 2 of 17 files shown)


jlebon:
Still need to test this in stage.

docs/config.yaml
```diff
@@ -55,6 +55,9 @@ streams:
         legacy_oscontainer_old: registry.ci.openshift.org/rhcos-devel/machine-os-content
       # OPTIONAL: override default coreos-assembler image to build with
       cosa_img: quay.io/coreos-assembler/coreos-assembler:stable
+      # OPTIONAL/TEMPORARY: override default coreos-assembler image to use for
+      # the controller pod in a multi-arch build
+      cosa_controller_img: quay.io/coreos-assembler/coreos-assembler:latest
```
dustymabe, line 60: suggestion: `cosa_build_arch_job_controller_img_hack`

This maybe should have a more complete comment with more context too.

jobs/build-arch.Jenkinsfile
```diff
@@ -70,6 +70,8 @@ def cosa_img = params.COREOS_ASSEMBLER_IMAGE ?: pipeutils.get_cosa_img(pipecfg,
 
 def stream_info = pipecfg.streams[params.STREAM]
 
+def cosa_controller_img = stream_info.cosa_controller_img ?: cosa_img
```
dustymabe, line 73: this probably needs an appropriate comment about the context like the comment from the code block we're removing.

dustymabe (commented):
Thanks for working on this!

For all of the `bump-lockfile` and `kola-*` jobs we've never run them in a context other than latest cosa. Not sure if it makes sense to templatize them on stream or not like the build jobs.

docs/config.yaml
```diff
@@ -21,6 +21,10 @@ default_artifacts:
     - aws
     - vmware
 
+# OPTIONAL: coreos-assembler image to build with. supports ${STREAM} variable.
+# If unset, defaults to `coreos-assembler:main` (ImageStream).
```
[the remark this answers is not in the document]
jlebon: Ahh yup, that's leftover from when I initially wanted to do these changes in a different order (adding a knob before nuking the ImageStream).

jobs/build-arch.Jenkinsfile
jlebon: I was thinking the comment block deleted here would be moved (and reworded) to RHCOS' `config.yaml` above `cosa_build_arch_job_controller_img_hack` since that's the only place we'd actually configure it. WDYT?

jlebon:
> For all of the `bump-lockfile` and `kola-*` jobs we've never run them in a context other than latest cosa. Not sure if it makes sense to templatize them on stream or not like the build jobs. If we do decided it matters then we should make the automated spawning of the `kola-*` jobs from the `build*` jobs specify the COSA image as a parameter too.

I think for `bump-lockfile` it makes sense to respect the stream-level cosa image. Once we start having lockfiles in RHCOS, we'll want `bump-lockfile` to build with the same cosa it would've in e.g. `build`.

For the `kola-*` jobs it's less clear. It makes sense too for correctness, but with almost all the tests being external tests now, it's not as crucial to match. I'm also not sure if the cloud provisioning code in older cosa branches would need updating. So maybe we can leave those off for now and revisit if we actually hit something.

Will update.

jobs/build-arch.Jenkinsfile
dustymabe: I guess that's sufficient. In that case let's add a comment in the docs/config.yaml about RHCOS being the only consumer of this right now and we'll be able to remove it in the future.

jlebon:
Updated for comments! I've restored the `kola-*` jobs for now and split out the `bump-lockfile` change in a separate commit since it's technically a semantic change.

docs/config.yaml
```diff
@@ -49,6 +53,13 @@ streams:
       additional_registry_repos:
         oscontainer_old: registry.ci.openshift.org/rhcos-devel/machine-os-oci-content
         legacy_oscontainer_old: registry.ci.openshift.org/rhcos-devel/machine-os-content
+      # OPTIONAL: override default coreos-assembler image to build with
+      cosa_img: quay.io/coreos-assembler/coreos-assembler:stable
+      # OPTIONAL/TEMPORARY: Override default coreos-assembler image to use for
+      # the controller pod in the build-arch job. This is only useful when
+      # building with a cosa too old to support `cosa remote`, which should only
```
dustymabe, line 60: I think it's `remote-session`:

```suggestion
      # building with a cosa too old to support `cosa remote-session`, which should only
```

jobs/build-arch.Jenkinsfile
```diff
@@ -66,8 +66,12 @@ properties([
     durabilityHint('PERFORMANCE_OPTIMIZED')
 ])
 
+def cosa_img = params.COREOS_ASSEMBLER_IMAGE ?: pipeutils.get_cosa_img(pipecfg, params.STREAM)
```
dustymabe, line 69: would this work? 

```suggestion
def cosa_img = params.COREOS_ASSEMBLER_IMAGE ?: \
                   pipeutils.get_cosa_img(pipecfg, params.STREAM)
```

These lines are getting a little long

dustymabe (approved):
LGTM

jlebon:
Now tested that the various knobs in `build` and `build-arch` in stg correctly get wired!

jobs/build-arch.Jenkinsfile
jlebon: I'm not sure, but I reworked it some other other into two lines and added a clarifying comment.

jobs/build-arch.Jenkinsfile
```diff
@@ -46,7 +46,7 @@ properties([
                    description: 'Force AWS AMI replication for non-production'),
       string(name: 'COREOS_ASSEMBLER_IMAGE',
              description: 'Override coreos-assembler image to use',
-             defaultValue: "coreos-assembler:main",
+             defaultValue: "",
```
jlebon, line 49: One weird thing I noticed while testing in stg is that for some reason changing the default value back to the empty string didn't take hold. Even after running the job multiple times, it would still be `coreos-assembler:main`. I ended up having to rerun the `seed` job with `FORCE=1`.

dustymabe (approved):
LGTM

PR #722, opened 2022-10-28
jobs/release: add Fetch Artifacts stage

dustymabe:
For the Push Container stage we need to have all of those artifacts
actually fethed to the local pod. Because of the way buildfetch works
(stomping on meta.json) we need to run the fetch of the artifacts before
any stages that modify/update the meta.json file. Let's add a new stage
that will handle this for us before any of the later stages that modify
meta.json.


jlebon (approved):
Typo in second commit: fethed -> fetched
Otherwise LGTM!

PR #724, opened 2022-10-28
jobs/release: put new oscontainer under `oscontainer` key

jlebon:
It turns out that we won't need to support building *both* the legacy oscontainer and the new one for 4.12. In the new transition plan, the new format will directly take the place of the old one as `oscontainer` in `meta.json` and `machine-os-content` in the release payload. This minimizes the impact on OpenShift tooling.

Update the release job to set the `oscontainer` key for the new container and verify that we're not pushing both the legacy and new container since they would conflict.

This change almost matches the latest behaviour of the old RHCOS pipeline. One difference is that we're no longer setting the key `base-oscontainer`. I think we should delete it from the schema eventually.

See also: https://issues.redhat.com/browse/MCO-392

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -61,10 +61,12 @@ s3_bucket: fcos-builds
 # OPTIONAL: container registry-related keys
 registry_repos:
   # OPTIONAL: repo to which to push oscontainer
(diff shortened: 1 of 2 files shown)


dustymabe (approved):
LGTM

PR #726, opened 2022-10-28
utils: add AWSBuildUpload credential helper functions

dustymabe:
This adds the following functions:

- withAWSBuildUploadCredentials(Closure body)
- shwrapWithAWSBuildUploadCredentials(cmds)
- AWSBuildUploadCredentialExists()

and changes code throughout the repo to take advantage of them.

This has the side effect of dropping the use of tryWithOrWithoutCredentials at the top of the build and build-arch jobs.

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -103,16 +103,11 @@ lock(resource: "build-${params.STREAM}") {
         currentBuild.description = "[${params.STREAM}][${basearch}] Running"
 
 
-        // Add in AWS Build Upload credentials here if they exist. In
-        // the future we might choose to be more granular about when
-        // we load this.
-        pipeutils.tryWithOrWithoutCredentials([file(variable: 'AWS_BUILD_UPLOAD_CONFIG',
-                                                    credentialsId: 'aws-build-upload-config')]) {
 
         // this is defined IFF we *should* and we *can* upload to S3
         def s3_stream_dir
 
-        if (pipecfg.s3_bucket && utils.pathExists("\${AWS_BUILD_UPLOAD_CONFIG}")) {
+        if (pipecfg.s3_bucket && pipeutils.AWSBuildUploadCredentialExists()) {
             // see bucket layout in https://github.com/coreos/fedora-coreos-tracker/issues/189
             s3_stream_dir = "${pipecfg.s3_bucket}/prod/streams/${params.STREAM}"
         }
@@ -174,14 +169,14 @@ lock(resource: "build-${params.STREAM}") {
         // buildfetch previous build info
         stage('BuildFetch') {
             if (s3_stream_dir) {
-                shwrap("""
+                pipeutils.shwrapWithAWSBuildUploadCredentials("""
                 cosa buildfetch --arch=${basearch} \
                     --url s3://${s3_stream_dir}/builds \
                     --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG}
                 """)
                 if (parent_version != "") {
                     // also fetch the parent version; this is used by cosa to do the diff
-                    shwrap("""
+                    pipeutils.shwrapWithAWSBuildUploadCredentials("""
                     cosa buildfetch --arch=${basearch} \
                         --build ${parent_version} \
                         --url s3://${s3_stream_dir}/builds \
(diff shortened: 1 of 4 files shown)


utils.groovy
```diff
@@ -396,4 +396,29 @@ def get_cosa_img(pipecfg, stream) {
...
+    withAWSBuildUploadCredentials() {
+        shwrap(cmds)
+    }
+}
+
+// Return true or false based on if the AWS Build Upload credential exists
+def AWSBuildUploadCredentialExists() {
+    creds = [file(variable: 'AWS_BUILD_UPLOAD_CONFIG',
+                credentialsId: 'aws-build-upload-config')]
+    return utils.credentialsExist(creds)
```
jlebon, line 421: Minor: feels like we can inline this.

jobs/build.Jenkinsfile
```diff
@@ -319,18 +314,18 @@ lock(resource: "build-${params.STREAM}") {
...
-              pipeutils.bump_builds_json(
-                  params.STREAM,
-                  newBuildID,
-                  basearch,
-                  s3_stream_dir)
+                // run with --force here in case the previous run of the
+                // pipeline died in between buildupload and bump_builds_json()
+                pipeutils.shwrapWithAWSBuildUploadCredentials("""
+                cosa buildupload --force --skip-builds-json --artifact=ostree \
+                    s3 --aws-config-file=\${AWS_BUILD_UPLOAD_CONFIG} \
```
jlebon, line 321: Do we actually need `--aws-config-file` since `withAWSBuildUploadCredentials` sets ` AWS_CONFIG_FILE` as we did directly before here?

Similarly in some other hunks.

jlebon (commented):
Some comments, but big :+1: to the idea!

jobs/build.Jenkinsfile
dustymabe: We do if we want it to work in a remote session.

utils.groovy
dustymabe: you mean instead of using a separate `creds` variable. I had it like that before but didn't like it so I switched it to this.

utils.groovy
```diff
@@ -396,4 +396,29 @@ def get_cosa_img(pipecfg, stream) {
     return utils.substituteStr(cosa_img, [STREAM: stream])
 }
 
+// Run a closure inside a context that has access to the AWS Build
+// Upload credentials.
+def withAWSBuildUploadCredentials(Closure body) {
+    withCredentials([file(variable: 'AWS_CONFIG_FILE',
```
jlebon, line 402: Shouldn't this be `AWS_BUILD_UPLOAD_CONFIG`?

dustymabe: yep!

dustymabe: fixed!

jobs/release.Jenkinsfile
```diff
@@ -71,13 +71,13 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
 
         // Fetch metadata files for the build we are interested in
         stage('Fetch Metadata') {
-            withCredentials([file(variable: 'AWS_CONFIG_FILE',
```
jlebon, line 74: Hmm right, I forgot `AWS_CONFIG_FILE` is a well-defined env var that some of the stack understands. Should we have `withAWSBuildUploadCredentials` define that instead and update the bits which currently use AWS creds to drop the `--aws-config-file`-style arg when we can?

dustymabe: I'd prefer to default to passing the file via an argument (rather than via an env var) everywhere since that's the lowest common denominator that works locally or in a remote session.

utils.groovy
```diff
@@ -396,4 +396,29 @@ def get_cosa_img(pipecfg, stream) {
...
+// Upload credentials.
+def shwrapWithAWSBuildUploadCredentials(cmds) {
+    withAWSBuildUploadCredentials() {
+        shwrap(cmds)
+    }
+}
+
+// Return true or false based on if the AWS Build Upload credential exists
+def AWSBuildUploadCredentialExists() {
+    creds = [file(variable: 'AWS_BUILD_UPLOAD_CONFIG',
```
jlebon, line 419: Noticed while working on top of this:

```suggestion
    def creds = [file(variable: 'AWS_BUILD_UPLOAD_CONFIG',
```

dustymabe: fixed!

jlebon (commented):
I think we're missing an update to `aws_s3_cp_allow_noent` and `bump_builds_json` too, no?

dustymabe:
> I think we're missing an update to `aws_s3_cp_allow_noent` and `bump_builds_json` too, no?

fixed in latest upload.

dustymabe:

dustymabe:
> Actually, see [#729 (comment)](https://github.com/coreos/fedora-coreos-pipeline/pull/729#discussion_r1010757232).

fixed

PR #727, opened 2022-10-28
Add hack knob to skip SecureBoot tests

jlebon:
For the ART hotfix process, we sometimes need to be able to build with an unsigned kernel. This means that UEFI SecureBoot tests will fail. We should support skipping UEFI tests from the src config repo via e.g. `kola-denylist.yaml`. But to do this properly, we need to first lower `cosa kola --basic-qemu-scenarios` into kola proper, and also ideally merge `kola testiso` into `kola run` (maybe as external tests?) or introduce a new `kola-testiso-denylist.yaml` file (or change the schema of the existing `kola-denylist.yaml`).

Short-term, let's just hack around this by adding a `skip_secureboot_tests_hack` knob in the pipecfg which automatically tells `kola()` and `kolaTestIso()` to skip over SecureBoot tests.

Requires: https://github.com/coreos/coreos-ci-lib/pull/133

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -305,7 +305,8 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
         stage("Kola") {
             def n = 4 // VMs are 2G each and arch builders have approx 32G
             kola(cosaDir: env.WORKSPACE, parallel: n, arch: basearch,
-                 allowUpgradeFail: params.ALLOW_KOLA_UPGRADE_FAILURE)
+                 allowUpgradeFail: params.ALLOW_KOLA_UPGRADE_FAILURE,
+                 skipSecureBoot: pipecfg.hotfix?.skip_secureboot_tests_hack)
         }
 
         // Build the remaining artifacts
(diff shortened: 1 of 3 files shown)


jobs/build-arch.Jenkinsfile
```diff
@@ -315,7 +315,8 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
         stage("Kola") {
             def n = 4 // VMs are 2G each and arch builders have approx 32G
             kola(cosaDir: env.WORKSPACE, parallel: n, arch: basearch,
-                 allowUpgradeFail: params.ALLOW_KOLA_UPGRADE_FAILURE)
+                 allowUpgradeFail: params.ALLOW_KOLA_UPGRADE_FAILURE,
+                 skipSecureBoot: pipecfg.hotfix?.skip_secureboot_tests_hack)
```
dustymabe, line 319: what does `pipecfg.hotfix?.skip_secureboot_tests_hack` evaluate to if that isn't in the config? `null`?

dustymabe (approved):
LGTM

dustymabe:

PR #728, opened 2022-10-31
Support pushing containers in v2s2 format

jlebon:
A bunch of tooling around OpenShift don't support the OCI format yet, so
for the RHCOS images we need to push in v2s2 format.

For FCOS, we want to keep pushing in OCI format. Add a new config.yaml
knob to override this for RHCOS.

Related: https://github.com/coreos/coreos-assembler/pull/2726

### DIFF
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -223,7 +223,10 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
                                           credentialsId: 'oscontainer-push-registry-secret')]) {
                         def repo = registry_repos[configname]
                         def (artifact, metajsonname, tag_suffix) = val
-                        def arch_args = basearches.collect{"--arch ${it}"}.join(" ")
+                        def extra_args = basearches.collect{"--arch ${it}"}.join(" ")
+                        if (registry_repos.v2s2) {
+                            extra_args += " --v2s2"
+                        }
                         def tag_args = " --tag=${params.STREAM}${tag_suffix}"
                         if (registry_repos.add_build_tag) {
                             tag_args += " --tag ${params.VERSION}${tag_suffix}"
@@ -232,7 +235,7 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
                         cosa push-container-manifest --auth=\${REGISTRY_SECRET} \
                             --repo=${repo} ${tag_args} \
                             --artifact=${artifact} --metajsonname=${metajsonname} \
-                            --build=${params.VERSION} ${arch_args}
+                            --build=${params.VERSION} ${extra_args}
                         """)
 
                         def old_repo = registry_repos["${configname}_old"]
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -223,19 +223,19 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
                                           credentialsId: 'oscontainer-push-registry-secret')]) {
                         def repo = registry_repos[configname]
                         def (artifact, metajsonname, tag_suffix) = val
-                        def extra_args = basearches.collect{"--arch ${it}"}.join(" ")
+                        def extra_args = basearches.collect{"--arch ${it}"}
                         if (registry_repos.v2s2) {
-                            extra_args += " --v2s2"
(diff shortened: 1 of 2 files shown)


jlebon:
Let me verify locally this actually fixes the issue given that some of the other containers did push successfully.

config.yaml
```diff
@@ -29,10 +29,6 @@ s3_bucket: fcos-builds
 
 registry_repos:
   oscontainer: quay.io/fedora/fedora-coreos
-  # For a period of time let's also mirror into the old location too
-  # Drop this after October 2022. See
-  # https://discussion.fedoraproject.org/t/updated-registry-location-for-fedora-coreos-ostree-native-container/42740
-  oscontainer_old: quay.io/coreos-assembler/fcos
```
dustymabe, line 35: 👍

dustymabe (approved):

cgwalters:
Hmm...but the two images will have different digests now, which I think is going to cause problems.  We could push to quay.io in *both* formats (with different tags), but it's probably simplest to add a global pipeline flag to push in v2s2 format for all containers we build by default, and use it for now with RHCOS.

config.yaml
jlebon: Split this out into https://github.com/coreos/fedora-coreos-pipeline/pull/730.

jlebon:
> Hmm...but the two images will have different digests now, which I think is going to cause problems. We could push to quay.io in _both_ formats (with different tags), but it's probably simplest to add a global pipeline flag to push in v2s2 format for all containers we build by default, and use it for now with RHCOS.

Indeed, overriding the format at `skopeo copy` time is fishy.

Anyway if I understand right, not all OCP tooling support the OCI format yet (as per https://github.com/coreos/fedora-coreos-pipeline/pull/730). So I think the correct fix for this is to (1) push the Quay.io container in v2s2 for RHCOS, and (2) support copying v2s2 manifest lists from Quay.io to registry.ci. `skopeo copy` by default will retain the same format.

This PR will enable support for (1). (2) is fixed by https://github.com/coreos/fedora-coreos-pipeline/pull/730.

jobs/release.Jenkinsfile
```diff
@@ -223,7 +223,10 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
                                           credentialsId: 'oscontainer-push-registry-secret')]) {
                         def repo = registry_repos[configname]
                         def (artifact, metajsonname, tag_suffix) = val
-                        def arch_args = basearches.collect{"--arch ${it}"}.join(" ")
+                        def extra_args = basearches.collect{"--arch ${it}"}.join(" ")
+                        if (registry_repos.v2s2) {
+                            extra_args += " --v2s2"
+                        }
```
dustymabe, line 229: I wonder if it would be worth changing this code to use a extra_args list and then do the `.join(' ')` in the shwrap call below. like we do with `fetch_args` in the `Fetch Artifacts` stage above. Prevents us having to handle whitespace when we define the args.

jlebon: Added a commit for that!

jobs/release.Jenkinsfile
```diff
@@ -223,16 +223,19 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
-                        def tag_args = " --tag=${params.STREAM}${tag_suffix}"
+                        def extra_args = basearches.collect{"--arch ${it}"}
+                        if (registry_repos.v2s2) {
+                            extra_args += "--v2s2"
+                        }
+                        def tag_args = ["--tag=${params.STREAM}${tag_suffix}"]
                         if (registry_repos.add_build_tag) {
-                            tag_args += " --tag ${params.VERSION}${tag_suffix}"
+                            tag_args += "--tag=${params.VERSION}${tag_suffix}"
                         }
```
dustymabe, line 233: merely preference here: I would just add these into `extra_args`. I don't think the order of the arguments(options) matters does it?

dustymabe: oh I see - we re-use the tag_args below. NVM

dustymabe (approved):
LGTM

dustymabe:
This PR is approved and will be merged after we unfreeze this repo (after this week's FCOS releases).

Issue #723, opened 2022-10-28
Cannot push images in OpenShift 4.11


jlebon:
In a 4.11 cluster, the release job fails with:

```
+ cosa push-container-manifest --auth=**** --repo=quay.io/openshift-release-dev/rhel-coreos-dev --tag=4.12-extensions --tag 412.86.202210272216-0-extensions --artifact=extensions-container --metajsonname=extensions-container --build=412.86.202210272216-0 --arch x86_64
Targeting build: 412.86.202210272216-0
2022-10-28 03:40:14,963 INFO - Running command: ['podman', 'manifest', 'create', 'quay.io/openshift-release-dev/rhel-coreos-dev:4.12-extensions']
...
time="2022-10-28T03:40:15Z" level=warning msg="\"/\" is not a shared mount, this could cause issues or missing mounts with rootless containers"
time="2022-10-28T03:40:15Z" level=error msg="cannot find UID/GID for user cosa: no subuid ranges found for user \"cosa\" in /etc/subuid - check rootless mode in man pages."
time="2022-10-28T03:40:15Z" level=warning msg="Using rootless single mapping into the namespace. This might break some images. Check /etc/subuid and /etc/subgid for adding sub*ids if not using a network user"
Error: failed to mount overlay for metacopy check with "" options: permission denied
```

I think this is fallout from [the enablement of the Pod Security Admission in OCP 4.11](https://docs.openshift.com/container-platform/4.11/release_notes/ocp-4-11-release-notes.html#ocp-4-11-auth-pod-security-admission). Part of the change includes defaulting to a tighter `restricted-v2` SCC which by default drops all caps and sets the seccomp profile to `runtime/default` where before the default was `unconfined`. This prevents podman from working. [There is an RFE](https://issues.redhat.com/browse/RFE-2868) for shipping a `nested` SCC for this use case.

Interestingly, the change was ratcheted in such that upgrading clusters still had access to the old `restricted` SCC. But specifically for seccomp, AIUI a pod that doesn't specify a profile will still match `restricted-v2` and not `restricted`. And explicitly setting a seccomp profile is not permitted by the SCC.

jlebon:
Some possible workarounds:
1. Request access to more powerful SCCs. This increases our constraints on the cluster setup. Currently we can work fine with the default restricted SCC and only KVM.
2. Do pushing from supermin. We do this for other privileged operations already. Adds complexity to the cosa code.
[...]

PR #729, opened 2022-10-31
Support running the release job on a multi-arch builder

dustymabe:
We thought we were going to need this as a workaround for https://github.com/coreos/fedora-coreos-pipeline/issues/723. It turns out we found another path, but I think it's worth us merging this code anyway so we can easily switch to using one of our builders if we need to in the future.

### DIFF
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -49,12 +49,14 @@ if (params.VERSION == "") {
 def cosa_img = params.COREOS_ASSEMBLER_IMAGE
 cosa_img = cosa_img ?: pipeutils.get_cosa_img(pipecfg, params.STREAM)
 
+def stream_info = pipecfg.streams[params.STREAM]
+
+def cosa_controller_img = stream_info.cosa_controller_img_hack ?: cosa_img
+
 currentBuild.description = "[${params.STREAM}][${params.ARCHES}] - ${params.VERSION}"
 
 // Get the list of requested architectures to release
 def basearches = params.ARCHES.split() as Set
-
-def stream_info = pipecfg.streams[params.STREAM]
 
 // We just lock here out of an abundance of caution in case somehow two release
 // jobs run for the same stream, but that really shouldn't happen. Anyway, if it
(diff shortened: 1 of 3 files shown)


jobs/release.Jenkinsfile
```diff
@@ -49,26 +49,38 @@ if (params.VERSION == "") {
...
-def stream_info = pipecfg.streams[params.STREAM]
-
 // We just lock here out of an abundance of caution in case somehow two release
 // jobs run for the same stream, but that really shouldn't happen. Anyway, if it
 // *does*, this makes sure they're run serially.
 // Also lock version-arch-specific locks to make sure these builds are finished.
 def locks = basearches.collect{[resource: "release-${params.VERSION}-${it}"]}
 lock(resource: "release-${params.STREAM}", extra: locks) {
-    cosaPod(cpu: "1", memory: "512Mi", image: cosa_img) {
+    cosaPod(cpu: "1", memory: "512Mi", image: cosa_controller_img) {
```
jlebon, line 67: We need to undo this too.

utils.groovy
```diff
@@ -97,8 +97,7 @@ def bump_builds_json(stream, buildid, arch, s3_stream_dir) {
     lock(resource: "bump-builds-json-${stream}") {
         def remotejson = "s3://${s3_stream_dir}/builds/builds.json"
         aws_s3_cp_allow_noent(remotejson, './remote-builds.json')
-        shwrap("""
-        export AWS_CONFIG_FILE=\${AWS_BUILD_UPLOAD_CONFIG}
+        shwrapWithAWSBuildUploadCredentials("""
```
jlebon, line 100: Ahh, there it is. :) This needs to be in the other PR that adds that function instead.

utils.groovy
```diff
@@ -109,7 +108,9 @@ def bump_builds_json(stream, buildid, arch, s3_stream_dir) {
             mkdir -p builds
             cp \$TMPD/builds/builds.json builds/builds.json
         fi
-        aws s3 cp --cache-control=max-age=300 --acl=public-read builds/builds.json s3://${s3_stream_dir}/builds/builds.json
+        export AWS_CONFIG_FILE=\${AWS_BUILD_UPLOAD_CONFIG}
+        aws s3 cp --cache-control=max-age=300    \
+            --acl=public-read builds/builds.json s3://${s3_stream_dir}/builds/builds.json
```
jlebon, line 113: I'd prefer we split this into its own `shwrapWithAWSBuildUploadCredentials` instead and leave the main on as `shwrap`.

jobs/release.Jenkinsfile
dustymabe: good catch!

utils.groovy
dustymabe: ok fixed.

utils.groovy
dustymabe: will move to the other PR.

jobs/release.Jenkinsfile
```diff
@@ -78,10 +90,12 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
                 --arch=all --url=s3://${s3_stream_dir}/builds \
                 --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG}
             """)
-            }
```
jlebon, line 81: This needs to be in the other AWS creds PR.

dustymabe: moved!

jlebon (approved):

dustymabe:

PR #731, opened 2022-10-31
jobs/release: fix legacy-oscontainer handling

jlebon:
On RHCOS 4.11 and older, we will have the legacy oscontainer and the OSTree archive but we only want to push the former. Hack around this by detecting whether the legacy-oscontainer was built and skipping the `ostree` artifact fetch and push.

Fixes fa256df ("jobs/release: put new oscontainer under `oscontainer` key").

Co-authored-by: Dusty Mabe <dusty@dustymabe.com>


dustymabe (approved):
LGTM

PR #732, opened 2022-11-01
add in workarounds for new restricted permissions in OCP 4.11

dustymabe:
```
commit 02ba886a0337c9e7805533fd49b41e47d20c02ab
Date:   Mon Oct 31 20:30:07 2022 -0400

    tree-wide: use VFS storage driver for container pushes
    
    This is a less hacky workaround for
    https://github.com/coreos/fedora-coreos-pipeline/issues/723
    
    Since we're just pushing containers the performance hit won't be
    much.

commit f3a0dc7b175455998f9ef2c23b810deba3a02682
Date:   Mon Oct 31 16:50:04 2022 -0400

    tree-wide: use different home dir to workaround podman mount issue
    
    This is to address https://github.com/coreos/fedora-coreos-pipeline/issues/723
    
    One workaround I've found is to change my `$HOME` to a different
    filesystem (`/var/tmp`, which is overlayfs) and it seems to workaround
    the offending check [1].
    
    This shouldn't work because there is another check [2] that should fail
    if we're on an overlay filesystem, but it doesn't in this case. I suspect
    we're hitting some kind of corner case.
    
    [1] https://github.com/containers/storage/blob/232bf398bd68595d2a6e53ec07d8b3c4424be3c0/drivers/overlay/check.go#L158-L164
    [2] https://github.com/containers/storage/blob/232bf398bd68595d2a6e53ec07d8b3c4424be3c0/drivers/overlay/overlay.go#L314-L318
```

There's a lot going on in the commit messages.. TL;DR use the `vfs`
storage driver to workaround https://github.com/coreos/fedora-coreos-pipeline/issues/723

### DIFF
--- a/jobs/build-cosa.Jenkinsfile
+++ b/jobs/build-cosa.Jenkinsfile
@@ -136,6 +136,8 @@ try {
                     images += " --image=docker://${params.CONTAINER_REGISTRY_STAGING_REPO}:${arch}-${shortcommit}"
                 }
                 shwrap("""
+                mkdir /var/tmp/builder
+                export HOME=/var/tmp/builder # https://github.com/coreos/fedora-coreos-pipeline/issues/723#issuecomment-1297668147
                 cosa push-container-manifest --v2s2 \
                     --auth=\$REGISTRY_SECRET --tag ${gitref} \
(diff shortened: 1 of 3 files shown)


jlebon (approved):

dustymabe:

PR #733, opened 2022-11-01
multi-arch-builders: README updates for AWS instance bringup

dustymabe:
- use a more locked down security group
- use a subnet from us-east-1a
    - had trouble getting an instance in the other one
- Don't specify an AMI
    - forces the person to look up the latest one

### DIFF
--- a/multi-arch-builders/README.md
+++ b/multi-arch-builders/README.md
@@ -25,11 +25,11 @@ Bring the instance up with appropriate details:
 
 ```bash
 NAME='coreos-aarch64-builder'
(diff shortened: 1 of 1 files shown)


dustymabe:
> Minor/optional: commit title/message should mention aarch64 somewhere

but this applies to `aarch64` and `x86_64`, right?

jlebon:
> > Minor/optional: commit title/message should mention aarch64 somewhere
> 
> but this applies to `aarch64` and `x86_64`, right?

Ahh gotcha, missed that.

PR #735, opened 2022-11-03
jobs/build-cosa: add in hack for build COSA against dev 4.X branches

dustymabe:
For a period of time we're going to be hacking on the RHCOS pipeline older streams and we've created some branches in the COSA git repo that we'd like to build and push only to the staging quay repo.

### DIFF
--- a/jobs/build-cosa.Jenkinsfile
+++ b/jobs/build-cosa.Jenkinsfile
@@ -87,6 +87,14 @@ node {
     shortcommit = commit.substring(0,7)
 }
 
+// HACK: for a period of time we're going to be hacking on the RHCOS
+// pipeline older streams and we've created some branches in the COSA
+// git repo that we'd like to build and push only to the staging quay repo.
+def container_registry_repo = params.CONTAINER_REGISTRY_REPO
+if ((gitref.startsWith("rhcos-") && gitref.endsWith("-new")) {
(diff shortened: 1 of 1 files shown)


jobs/build-cosa.Jenkinsfile
```diff
@@ -87,6 +87,13 @@ node {
     shortcommit = commit.substring(0,7)
 }
 
+// HACK: for a period of time we're going to be hacking on the RHCOS
+// pipeline older streams and we've created some branches in the COSA
+// git repo that we'd like to build and push only to the staging quay repo.
+if ((gitref.startsWith("rhcos-") && gitref.endsWith("-new")) {
+    params.CONTAINER_REGISTRY_REPO = params.CONTAINER_REGISTRY_STAGING_REPO
```
jlebon, line 94: Sadly (but likely for the best), Jenkins will not let you change the value of a parameter.

dustymabe: hmm. could have sworn I'd done that before.. oh well - new upload incoming.

jlebon (approved):

jobs/build-cosa.Jenkinsfile
jlebon: Me too actually, but I actually tested it this morning for different reasons and hit that limitation.

PR #737, opened 2022-11-03
Lower FCOS S3 bucket prefix into config.yaml

jlebon:
The `prod/streams/${params.STREAM}` is an FCOS thing. Let's have the `s3_bucket` key in the `config.yaml` also incorporate the subpath.

This will allow the RHCOS config to specify its own subpath and is prep for hotfix builds specifying a different subpath.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -24,8 +24,10 @@ additional_arches: [aarch64, ppc64le, s390x]
 source_config:
   url: https://github.com/coreos/fedora-coreos-config
 
-# remove to disable S3 uploads
-s3_bucket: fcos-builds
+s3:
+  bucket: fcos-builds
+  # see bucket layout in https://github.com/coreos/fedora-coreos-tracker/issues/189
+  builds_key: "prod/streams/${STREAM}"
 
 registry_repos:
   oscontainer: quay.io/fedora/fedora-coreos
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -106,7 +106,7 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
             // We need to fetch a few artifacts if they were built. This assumes if
             // it was built for one platform it was built for all.
             def fetch_artifacts = ['ostree', 'extensions-container', 'legacy-oscontainer']
-            def meta = readJSON(text: shwrapCapture("cosa meta --arch=x86_64 --dump"))
+            def meta = readJSON(text: shwrapCapture("cosa meta --build=${params.VERSION} --arch=x86_64 --dump"))
             fetch_artifacts.retainAll(meta.images.keySet())
 
             def fetch_args = basearches.collect{"--arch=${it}"}
@@ -121,7 +121,7 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
         }
 
         for (basearch in basearches) {
-            def meta = readJSON(text: shwrapCapture("cosa meta --arch=${basearch} --dump"))
+            def meta = readJSON(text: shwrapCapture("cosa meta --build=${params.VERSION} --arch=${basearch} --dump"))
 
             // For production streams, import the OSTree into the prod
             // OSTree repo.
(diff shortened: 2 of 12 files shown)


config.yaml
```diff
@@ -24,8 +24,7 @@ additional_arches: [aarch64, ppc64le, s390x]
 source_config:
   url: https://github.com/coreos/fedora-coreos-config
```
dustymabe, line 26: should we add `see bucket layout in https://github.com/coreos/fedora-coreos-tracker/issues/189` in a comment here?

jobs/release.Jenkinsfile
```diff
@@ -285,11 +285,12 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
                 // Run plume to publish official builds; This will handle modifying
                 // object ACLs, modifying AMI image attributes,
                 // and creating/modifying the releases.json metadata index
+                def (bucket, subpath) = pipeutils.splitBucketKey(s3_stream_dir)
                 shwrap("""
                 cosa shell -- plume release --distro fcos \
                     --version ${params.VERSION} \
                     --stream ${params.STREAM} \
-                    --bucket ${pipecfg.s3_bucket} \
+                    --bucket ${bucket} \
```
dustymabe, line 293: underneath the covers what is plume doing with this bucket?

It feels a bit like we're reverse engineering the change we're making here and it feels a bit like the relationship here isn't clean.

jlebon:
> I wonder if there is a better way to achieve the goal here. To me if we change the top level `s3_bucket` to have a bucket+path then I'd expect the pipeline to not touch anything outside of that path in the bucket, which clearly isn't true (see `sync-stream-metadata` and also maybe the `plume` call in `release`).
> 
> Maybe what we should do here instead is separate things out a bit more:
> 
> ```
> s3:
>   bucket: fcos-builds
>   builds_path: prod/streams/${STREAM}
>   release_metadata_path: releases
> ```
>
> Obviously the stuff that happens in `sync-stream-metadata` is FCOS only which is why we're ok breaking the cleanliness there, but it might be nice to have the relationship be clearer.

Yeah, I pretty much optimized it for the jobs which are in common between FCOS and RHCOS. It didn't seem valuable to try to extend the config schema right now for prod FCOS-specific things.

Re. `builds_path`, I debated indeed whether to have it split from the bucket or unified. I ended up with the latter because that's how most jobs consume it anyway, but I think splitting now with an eye towards expanding in the future makes sense. Will tweak.

jobs/release.Jenkinsfile
jlebon: It ends up doing only things actually under the `prod/streams/${STREAM}` key. It could make sense to have `--bucket` take the key too, but I didn't want to tackle that right now. I might revisit this when looking at the whole `plume` call soon since that's another delta we need to parameterize.

dustymabe: Maybe call out the elephant in the room with a comment then?

utils.groovy
```diff
@@ -57,6 +57,10 @@ boolean checkKolaSuccess(file) {
     return true
 }
 
+def splitBucketKey(bucket_path) {
```
dustymabe, line 60: bikeshed: `splitS3BucketKey()`

Not required.

jobs/release.Jenkinsfile
```diff
@@ -273,11 +273,14 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
                 // Run plume to publish official builds; This will handle modifying
                 // object ACLs, modifying AMI image attributes,
                 // and creating/modifying the releases.json metadata index
+                // Note here we pass the bucket instead of `s3_stream_dir` but
+                // plume only actually interacts with objects under the
+                // `s3_stream_dir` key.
```
dustymabe, line 278:
```suggestion
                // Note here we pass the bucket instead of `s3_stream_dir` but
                // plume is hard coded to only interact with objects under the
                // `s3_stream_dir` key (i.e. it doesn't consider `s3.builds_key`).
                // This needs to be revisited in a followup.
```

dustymabe (approved):
LGTM - one requested comment update

dustymabe (approved):
LGTM

PR #738, opened 2022-11-03
Drop `official` conditional on `slackSend` and make Slack room more configurable

jlebon:

Requires: https://github.com/coreos/fedora-coreos-pipeline/pull/737

### DIFF
--- a/jenkins/config/slack.yaml
+++ b/jenkins/config/slack.yaml
@@ -2,4 +2,3 @@ unclassified:
   slackNotifier:
     teamDomain: coreos
     tokenCredentialId: slack-api-token
-    room: "#jenkins-fedora-coreos"
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -453,9 +453,7 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
     }
 
     echo message
-    if (official) {
-        slackSend(color: color, message: message)
-    }
+    pipeutils.trySlackSend(color: color, message: message)
     if (official) {
         pipeutils.tryWithMessagingCredentials() {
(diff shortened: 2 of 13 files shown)


HACKING.md
```diff
@@ -268,9 +268,14 @@ oc create secret generic slack-api-token --from-literal=text="${TOKEN}"
 oc label secret/slack-api-token \
     jenkins.io/credentials-type=secretText
 oc annotate secret/slack-api-token \
-    jenkins.io/credentials-description="Slack API token"
+    jenkins.io/credentials-description="Slack API token" \
+    jenkins.io/default-channel="#fedora-coreos-jenkins"
```
dustymabe, line 272: don't know if you were trying to be accurate or just giving an example here but the actual channel we use is `#jenkins-fedora-coreos`

dustymabe: actually it might be better her if we just make it generic and drop the paragraph below (the actual values will be in bitwarden anyway).

```
TOKEN=<token>
CHANNEL=<defaultChannel>
...
```

dustymabe (commented):
looking good - a few comments

HACKING.md
jlebon: Made into a variable. I'll update the BitWarden secrets once this merges.

jlebon:
Cool, CI caught the first genuine syntax error: https://jenkins-coreos-ci.apps.ocp.fedoraproject.org/blue/organizations/jenkins/fedora-coreos-pipeline/detail/PR-738/8/pipeline! :tada:

jlebon:
Tested and ready for review! As per the last commit, we actually don't need to specify the channel at all, which simplifies things.

dustymabe (approved):
LGTM

PR #739, opened 2022-11-03
Drop `official` concept, replace with hardcoded list of well-known protected Jenkins instances

jlebon:

Requires: https://github.com/coreos/fedora-coreos-pipeline/pull/738

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -32,6 +32,8 @@ s3:
 registry_repos:
   oscontainer: quay.io/fedora/fedora-coreos
 
+versionary_hack: true
+
 default_artifacts:
   all:
     - metal
--- a/utils.groovy
+++ b/utils.groovy
@@ -15,19 +15,21 @@ def load_pipecfg() {
     def url = jenkinscfg['pipecfg-url']
     def ref = jenkinscfg['pipecfg-ref']
 
+    def pipecfg
     if (url == 'in-tree') {
-        return readYaml(file: "config.yaml")
+        pipecfg = readYaml(file: "config.yaml")
+    } else {
+        // this uses the `checkout` workflow step instead of just manually cloning so
+        // that changes show up in the Jenkins UI
+        checkout([
+            $class: 'GitSCM',
+            branches: [[name: "origin/${ref}"]],
+            extensions: [[$class: 'RelativeTargetDirectory', relativeTargetDir: 'pipecfg']],
+            userRemoteConfigs: [[url: url]]
+        ])
+        pipecfg = readYaml(file: "pipecfg/config.yaml")
     }
-
-    // this uses the `checkout` workflow step instead of just manually cloning so
-    // that changes show up in the Jenkins UI
-    checkout([
-        $class: 'GitSCM',
-        branches: [[name: "origin/${ref}"]],
-        extensions: [[$class: 'RelativeTargetDirectory', relativeTargetDir: 'pipecfg']],
-        userRemoteConfigs: [[url: url]]
-    ])
-    return readYaml(file: "pipecfg/config.yaml")
+    return pipecfg
 }
 
 def get_source_config_ref_for_stream(pipecfg, stream) {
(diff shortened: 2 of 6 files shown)


utils.groovy
```diff
@@ -431,9 +431,19 @@ def AWSBuildUploadCredentialExists() {
...
 
+def getSlackDefaultChannel() {
+    return shwrapCapture("""
+        oc get secret -n ${env.PROJECT_NAME} -o json slack-api-token | \
+          jq -r '.metadata.annotations["jenkins.io/default-channel"]'
+    """)
+}
+
 // Emits Slack message if set up, otherwise does nothing.
 def trySlackSend(params) {
```
cgwalters, line 477: Personally, I think the Slack integration is a good example of how Jenkins pipelines can grow in an ad-hoc/unbounded way.  IMO, a better model is for things like chat notifications to *poll* (with opt-in event triggers) to send a message as part of a separate process (container).  This is how e.g. the Prow job alerts work as far as I know.

utils.groovy
```diff
@@ -29,9 +39,36 @@ def load_pipecfg() {
...
+        if (pipecfg.s3_bucket && bucket_prefix != pipecfg.s3_bucket) {
+            error("S3 bucket on ${env.JENKINS_URL} must be ${bucket_prefix} (found ${pipecfg.s3_bucket})")
+        }
+    } else if (AWSBuildUploadCredentialExists()) {
+        // We're *not* running on a protected Jenkins instance and we have AWS
+        // creds available. Check that we're not trying to push to a protected
+        // S3 bucket.
+        def protected_buckets = PROTECTED_JENKINSES.collectEntries{jenkins_url, v ->
+            def (bucket_prefix, allow_hotfixes) = v
+            [splitBucketKey(bucket_prefix)[0], jenkins_url]
```
dustymabe, line 63: What is this code doing? I don't quite understand it.

jlebon: I added a comment. Does that help?

config.yaml
```diff
@@ -29,6 +29,8 @@ s3_bucket: "fcos-builds/prod/streams/${STREAM}"
 registry_repos:
   oscontainer: quay.io/fedora/fedora-coreos
 
+versionary_hack: true
```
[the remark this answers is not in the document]
jlebon: I'd like to aim for something cleaner, so I'd prefer we mark it as temporary for now.

jlebon:
This is tested now!

utils.groovy
dustymabe: Yeah I mean I get what you are doing. I just don't really understand how the code achieves the goal. The `[bucket, jenkins_url]` on a line by itself confuses me.

I'm not too worried about it, though.

utils.groovy
```diff
@@ -15,19 +25,51 @@ def load_pipecfg() {
...
+
+        // Transform PROTECTED_JENKINSES to a map of {bucket -> jenkins URL}
+        // for easier lookup and reporting
+        def protected_buckets = PROTECTED_JENKINSES.collectEntries{jenkins_url, v ->
+            def (bucket, key, allow_hotfixes) = v
+            [bucket, jenkins_url]
+        }
+        if (pipecfg.s3.bucket in protected_buckets) {
+            error("S3 bucket ${pipecfg.s3.bucket} can only be used on ${protected_buckets[pipecfg.s3.bucket]}")
+        }
```
dustymabe, line 71: So here we are saying that no one can use the same bucket as a not-hotfix build on these jenkins instances. I imagine they might argue that they want to use the same bucket but a different path within the bucket, which we don't allow for here.

jlebon: Almost. This is in the non-protected Jenkins instance branch. We're checking that such Jenkins instances don't try to upload artifacts to the canonical S3 buckets.

The hotfix work in the next PR does support writing to the same bucket but different path for hotfix builds. The expectation is that they'll be using the same Jenkins instance for regular and hotfix builds.

dustymabe: oh OK. +1

dustymabe (approved):

PR #742, opened 2022-11-04
ci: add basic syntax checking of Groovy files

jlebon:
I realized recently that we could use `load()` in CI as a way to syntax-check our Groovy files. Add a CI check that uses this.

This works for regular Jenkinsfiles too (since it's still just the same custom Groovy interpreter). But `load()` will try to execute them, which is not what we want. Hack around this by adding a `return this` at the top. The parser will still actually scan the full file.

This only catches basic syntax errors like unbalanced parentheses or brackets, and not semantic errors like typoed function names or variables. We should be able to add a more complete CI run here in the future that e.g. could execute a part of the pipeline, but it requires some thought.

### DIFF
--- a/.cci.jenkinsfile
+++ b/.cci.jenkinsfile
@@ -3,8 +3,32 @@
 node {
     checkout scm
 
-    stage("Test") {
-        // for now, just sanity check all the manifests
+    stage("Check Groovy") {
+        // hacky way to convert Jenkinsfiles into a "library" without executing it
+        shwrap("""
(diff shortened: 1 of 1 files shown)


jlebon:
This seemed to work well in testing, but there may be gotchas I didn't think of wrt the Jenkinsfile hack that yields false positives. We can always go back and limit it to just the `*.groovy` files if that happens.

dustymabe (approved):
NOICE!!!

PR #744, opened 2022-11-04
ci: tweak syntax error and path printing

jlebon:
Don't reprint the error since `load` already prints it.

Print the path being tested before `load` otherwise there's a weird rendering issue where the error message is associated with the previous file tested.


jlebon:
One can see what it looks like in https://github.com/coreos/fedora-coreos-pipeline/pull/743.

dustymabe (approved):
LGTM

PR #745, opened 2022-11-07
utils: modify bump_builds_json to work if non x86_64 first

dustymabe:
In the case that there is a one-off mArch build that runs first for a particular stream then we need to copy the builds.json back to the local node before we upload.


jlebon (approved):
LGTM, but I'm not confident in the rest of our pipeline's ability to actually handle this use case (i.e. I wouldn't be surprised if other things break).

PR #746, opened 2022-11-08
Add replication to clouds 

ravanelli:


libupload.groovy
```diff
@@ -1,4 +1,92 @@
...
+            withCredentials(creds) {
+                utils.syncCredentialsIfInRemoteSession(["POWERVS_IMAGE_UPLOAD_CONFIG"])
+                def c = pipecfg.clouds.powervs
+                // for powervs in RHCOS the images are uploaded to a bucket in each
+                // region that is uniquely named with the region as a suffix
+                // i.e. `rhcos-powervs-images-us-east`
+                def regions = []
+                if (c.regions) {
+                     regions = c.regions.join(" ")
+                }
```
dustymabe, line 75: I wonder if it's possible to autodetect here like we do for the other clouds?

dustymabe (commented):
A few initial comments throughout the code.

dustymabe:
As part of this PR we probably need to do two other things:

- replace the aws replication code in `release.Jenkinsfile` with a call to the new replicate function.
- make the parameter to the different jobs for `AWS_REPLICATION` be something like `CLOUD_REPLICATION` now.

libcloud.groovy
```diff
@@ -1,4 +1,92 @@
...
+                def regions = []
+                if (c.regions) {
+                     regions = c.regions.join(" ")
+                }
+                shwrap("""
+                cosa powervs-replicate --cloud-object-storage ${c.cloud_object_storage_service_instance} \
+                    --bucket-prefix ${c.bucket} \
+                    --regions ${regions} \
+                    --credentials-file \${"POWER_IMAGE_UPLOAD_CONFIG"} \
+                    --build ${buildID}  \
```
dustymabe, line 81: pass in `--arch` here?

libupload.groovy
```diff
@@ -1,4 +1,92 @@
...
+    if (meta.aliyun) {
+        def creds = credentials
+        replicators["☁️ ⬆️ :aliyun"] = {
+            withCredentials(creds) {
+                utils.syncCredentialsIfInRemoteSession(["ALIYUN_IMAGE_UPLOAD_CONFIG"])
+                def c = pipecfg.clouds.aliyun
+                def extraArgs = []
+                if (c.regions) {
+                    extraArgs = "--regions" +  c.regions.join(" ")
+                }
```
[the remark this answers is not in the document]
ravanelli: Unless you add `regions` the default behavior is to try to all regions. The questions is, do we want to have the ability to replicate to a specific region/regions, or do we want to always replicate to all?

libupload.groovy
ravanelli: I'm still waiting for a test secret to see how things go. =/. I will try as son as I got it.

libupload.groovy
```diff
@@ -1,4 +1,92 @@
...
+                if (c.public) {
+                    extraArgs += "--public"
+                }
```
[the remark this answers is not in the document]
ravanelli: +1 . Yeah, at least we do have `--public` for RHCOS. Let me check it better and if that's the case I will open a PR to `upload_to_clouds()` and update the docs

libupload.groovy
```diff
@@ -1,4 +1,85 @@
...
+            def replicated = false
+            def ids = ['aws-build-upload-config', 'aws-govcloud-image-upload-config']
+            def c = pipecfg.clouds.aws
+            for (id in ids) {
+                tryWithCredentials([file(variable: 'AWS_CONFIG_FILE', credentialsId: id)]) {
+                    shwrap("""
+                    cosa aws-replicate \
+                        --build=${buildID} \
+                        --arch=${basearch} \
+                        --source-region ${c.primary_region} \
```
dustymabe, line 42: we can delete `--source-region` here. The code will autodetect it from the `meta.json` if not specified.

libupload.groovy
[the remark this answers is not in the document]
dustymabe: still need to add `public` in `docs/config.yaml`

also feel free to add another commit in this PR that adds the `--public` to the `upload_to_clouds()` function for `aliyun`.

libcloud.groovy
dustymabe: still need to do this?

libupload.groovy
dustymabe: Let's not wait on the credentials for this. Let's leave this code in for now but also document it in `docs/config.yaml`. We can post any followup fixes/improvements that we find in the coming days.

I'd like to get this PR merged sooner than later so we can start really testing it out (in the FCOS pipeline).

libupload.groovy
```diff
@@ -1,4 +1,84 @@
...
+                }
+                shwrap("""
+                cosa powervs-replicate \
+                    --cloud-object-storage ${c.cloud_object_storage_service_instance} \
+                    --build ${buildID} \
+                    --arch=${basearch} \
+                    --bucket-prefix ${c.bucket} \
+                    --regions ${regions} \
+                    --credentials-file \${"POWER_IMAGE_UPLOAD_CONFIG"} \
+                    --force
```
dustymabe, line 74: I think we should probably remove the `--force` here and in the `upload_to_clouds()` function below. In the new pipeline structure we just allocate a new build ID typically.

dustymabe (commented):
mostly LGTM - a few final comments and then we should be able to push this I think.

dustymabe (approved):
LGTM

PR #747, opened 2022-11-09
Support skipping or overriding additional arches per stream

jlebon:
Add support for `skip_additional_arches` and `additional_arches` keys in
the stream-level section. E.g. currently for FCOS, we're not shipping
ppc64le for the production streams.

This will also be useful for RHCOS in the future when e.g. keeping older
streams alive on only a subset of arches.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -70,8 +70,14 @@ streams:
       # building with a cosa too old to support `cosa remote-session`, which
       # should only be the case for RHCOS 4.11 and older.
       cosa_controller_img_hack: quay.io/coreos-assembler/coreos-assembler:latest
+      # OPTIONAL: override additional architectures to build for
+      additional_arches:
+        - s390x
     rawhide:
       type: mechanical
+      # OPTIONAL: additional architectures to skip building for
+      skip_additional_arches:
+        - s390x
 
 # REQUIRED: architectures to build for other than x86_64
 additional_arches: [aarch64, ppc64le, s390x]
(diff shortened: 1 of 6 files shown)


jlebon:
Requires: https://github.com/coreos/fedora-coreos-pipeline/pull/740

docs/config.yaml
```diff
@@ -70,8 +70,14 @@ streams:
       # building with a cosa too old to support `cosa remote-session`, which
       # should only be the case for RHCOS 4.11 and older.
       cosa_controller_img_hack: quay.io/coreos-assembler/coreos-assembler:latest
+      # OPTIONAL: override additional architectures to building for
```
dustymabe, line 73: to build for

docs/config.yaml
```diff
@@ -70,8 +70,14 @@ streams:
...
+      additional_arches:
+        - s390x
     rawhide:
       type: mechanical
+      # OPTIONAL: additional architectures to skip building for
+      skip_additional_arches:
+        - s390x
```
dustymabe, line 80: The use of the term `additional` in the toplevel key makes this key confusing. I would almost expect it to be `skip_arches` and not `skip_additional_arches`, but I understand tying it back to the name of the toplevel key.

I almost wonder if `secondary_arches` should be the toplevel key.

jlebon: Yeah, it's a bit awkward but I wanted to be consistent with the naming. It also conveys a bit more the fact that you can't skip x86_64, only arches that are valid in `additional_arches`.

Switching to `secondary` would indeed read a bit better I think. Let's do that in a follow-up?

dustymabe (approved):
LGTM

PR #748, opened 2022-11-09
jobs/release: don't push base-oscontainer on releases missing it

jlebon:
For RHCOS 4.11 and older, there is no native oscontainer to push so we don't want to even try. But unlike other cloud-y things, the artifact does exist (`ostree`), so we can't just rely on the fact that we're not building it.

Rather than adding more temporary knobs at the pipecfg, hack around this by checking if the cosa schema supports `base-oscontainer` and if not, don't try to push it.


jlebon:
(Untested, but I'm thinking we should just test it in the `rhcos` branch by rerunning one of the failed `release` jobs.)

dustymabe (approved):
Seems to be working in a `rhcos` pipeline test. LGTM

PR #750, opened 2022-11-11
jobs/{build,bump-lockfile}: bump up memory for testISO failure

dustymabe:
We're seeing an obscure `Got EOF from completion channel` error when running testISO. The running theory is that we're running out of memory and that seems to be accurate in a test I did earlier today. Let's bump for now and investigate when we have more time.

See https://github.com/coreos/fedora-coreos-tracker/issues/1339


marmijo (approved):
LGTM

PR #752, opened 2022-11-15
jobs/release: bump memory for release job

dustymabe:
We've seen the buildfetch getting killed and we think it's because we are spiking briefly above the 512Mi limit we had set for ourselves.


marmijo (approved):
LGTM

PR #753, opened 2022-11-15
add in hacks for rhcos 

dustymabe:
We want to drop these hacks in the near future, but for now we need to carry them so that we don't have to fork the pipeline.

```
commit 918b4d628c16baf7dfd941c65fd24f457f026b17
Date:   Mon Nov 14 16:54:03 2022 -0500

    add hacks.use_yumrepos_branch_workaround
    
    For the moment the git repo that we are pulling our yum repo files
    from has a branch per stream of RHCOS (i.e. 4.9, 4.10...). The plan
    is to get to a single branch that has all the yum repo files. Until
    that happens let's bake in a hack to support what currently exists.

commit 283d5d4a3cad3a75f85ba248cb1ecd0cba465035
Date:   Mon Nov 14 16:52:41 2022 -0500

    add hacks.skip_uefi_tests_on_older_rhcos
    
    On older RHCOS (i.e. 4.9) the UEFI+`iso-live-login` test isn't passing
    right now and the `iso-as-disk` test doesn't exist at all. Let's skip
    these for now. This will last until we either investigate and fix the
    failures or 4.9 no longer exists.

commit 4c08ef2ad220ba333f596a99a9ec965cd0e68901
Date:   Mon Nov 14 16:51:35 2022 -0500

    add hacks.skip_plume_release_task
    
    In the RHCOS pipeline for now we aren't running this `plume release`
    call. Let's add a hack to enable us to not run it for now while we
    further evaluate the need in the RHCOS pipeline.

commit 6f1f954f0931a7fccd5bb2f34156e70356e64825
Date:   Mon Nov 14 16:50:20 2022 -0500

    add hacks.skip_upgrade_tests
    
    We'll set this in the RHCOS pipeline until we figure out a more
    elegant way to get around the fact that we need to not run the upgrade
    tests in the RHCOS pipeline for now.

```

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -295,6 +295,7 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
         stage("Kola") {
             def n = 4 // VMs are 2G each and arch builders have approx 32G
             kola(cosaDir: env.WORKSPACE, parallel: n, arch: basearch,
+                 skipUpgrade: pipecfg.hacks?.skip_upgrade_tests,
                  allowUpgradeFail: params.ALLOW_KOLA_UPGRADE_FAILURE,
                  skipSecureBoot: pipecfg.hotfix?.skip_secureboot_tests_hack)
         }
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -155,11 +155,26 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
         }
 
         stage('Init') {
-            def yumrepos = pipecfg.source_config.yumrepos ? "--yumrepos ${pipecfg.source_config.yumrepos}" : ""
-
-            shwrap("""
-            cosa init --force --branch ${ref} --commit=${src_config_commit} ${yumrepos} ${pipecfg.source_config.url}
-            """)
+            if (pipecfg.hacks?.use_yumrepos_branch_workaround) {
+                // Right now the git repo that has our yum repo files isn't a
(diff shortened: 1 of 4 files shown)


jobs/build-arch.Jenkinsfile
```diff
@@ -155,11 +155,26 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
...
+                def yumrepos_ref = params.STREAM
+                if (yumrepos_ref == '4.13') {
+                    yumrepos_ref = 'master'
+                }
+                shwrap("""
+                cosa shell -- mkdir ./yumrepos
+                cosa shell -- git clone --depth=1 --branch=${yumrepos_ref} ${pipecfg.source_config.yumrepos} ./yumrepos
+                yumrepos=\$(cosa shell -- readlink -f ./yumrepos)
+                cosa init --force --branch ${ref} --commit=${src_config_commit} --yumrepos=\${yumrepos} ${pipecfg.source_config.url}
+                """)
```
jlebon, line 171: I think this would be cleaner as

```groovy
shwrap("""
cosa init --force --branch ${ref} --commit=${src_config_commit} ${pipecfg.source_config.url}
cosa shell -- git clone --depth=1 --branch=${yumrepos_ref} src/yumrepos
""")
```

I'm also more confident with `src/yumrepos` being a bona fide directory rather than a symlink (the latter was added mostly to support the developer case like we do `src/config`).

dustymabe: yeah. Since I've tested the current workflow AND we'll be removing this code soon when the yumrepos are all in a single branch I'll leave what I've got for now.

PR #756, opened 2022-11-16
jobs/build: add EARLY_ARCH_JOBS parameter

dustymabe:
This will give us control of when the multi-arch jobs are kicked off. In some cases we want them to be kicked off early (to save time). Unfortunately in that case if the x86_64 build later fails (which means the build ID can no longer be used) the work done by the multi-arch builds gets trashed as well.

Let's add a knob so we can kick them off early when desired, but allow for them to be kicked off later in situations where time isn't a large factor.

### DIFF
--- a/jobs/build-development.Jenkinsfile
+++ b/jobs/build-development.Jenkinsfile
@@ -23,7 +23,8 @@ node {
     stream = pipeutils.stream_from_branch(change.GIT_BRANCH)
     if (stream in development_streams) {
         build job: 'build', wait: false, parameters: [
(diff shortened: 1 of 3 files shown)


dustymabe:
Of course, we can bikeshed on the parameter name.

jlebon (approved):
We discussed this OOB. It's hard to evaluate the complexity increase here against the slight semantic change. I'm good with this but I think we should investigate replacing it with something that aborts the builds instead (similar to the [milestone](https://plugins.jenkins.io/pipeline-milestone-step/) plugin but that actually works across different jobs).

PR #757, opened 2022-11-16
jenkins/config: add empty `room` field

jlebon:
After 27693a6, the Slack plugin fails to send messages on new Jenkins instances with:

    java.lang.IllegalArgumentException: Project Channel or Slack User ID
    must be specified.

This seems to be due to some bug in the plugin where without a `room` setting, it doesn't know to rely on the default room associated with the token. This is filed upstream in:

https://github.com/jenkinsci/slack-plugin/issues/857

Short-term, let's just work around this by specifying an empty `room`.


dustymabe (approved):
LGTM

PR #758, opened 2022-11-16
add cloud_replicate.Jenkinsfile

dustymabe:


jobs/release.Jenkinsfile
```diff
@@ -117,6 +117,10 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
             }
         }
 
+        // Update description based on updated set of architectures
+        build_description += "[${params.STREAM}][${basearches.join(' ')}][${params.VERSION}]"
```
jlebon, line 121: Shouldn't this be `=`?

jobs/cloud_replicate.Jenkinsfile
```diff
@@ -0,0 +1,147 @@
...
+                echo "ERROR: Some requested architectures did not successfully build"
+                echo "ERROR: Detected built architectures: $builtarches"
+                echo "ERROR: Requested base architectures: $basearches"
+                currentBuild.result = 'FAILURE'
+                return
+            }
+        }
+
+        // Update description based on updated set of architectures
+        build_description += "[${params.STREAM}][${basearches.join(' ')}][${params.VERSION}]"
```
jlebon, line 112: Here too.

jobs/cloud_replicate.Jenkinsfile
```diff
@@ -0,0 +1,147 @@
...
+
+def stream_info = pipecfg.streams[params.STREAM]
+
+build_description += "[${basearches.join(' ')}][${params.VERSION}]"
+currentBuild.description = "${build_description} Waiting"
+
+// We just lock here out of an abundance of caution in case somehow
+// two jobs run for the same stream, but that really shouldn't happen.
+// Also lock version-arch-specific locks to make sure these builds are finished.
+lock(resource: "cloud-replicate-${params.STREAM}") {
```
jlebon, line 72: I don't think we need this lock.

jobs/cloud_replicate.Jenkinsfile
```diff
@@ -0,0 +1,147 @@
...
+                --arch=all --url=s3://${s3_stream_dir}/builds \
+                --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG}
+            """)
+        }
+
+        def builtarches = shwrapCapture("""
+                          cosa shell -- cat builds/builds.json | \
+                              jq -r '.builds | map(select(.id == \"${params.VERSION}\"))[].arches[]'
+                          """).split() as Set
+        assert builtarches.contains("x86_64"): "The x86_64 architecture was not in builtarches."
```
jlebon, line 97: Minor: feels like we could split this out into a shared function.

jobs/cloud_replicate.Jenkinsfile
```diff
@@ -0,0 +1,147 @@
...
+                def arch_args = basearches.collect{"--arch ${it}"}.join(" ")
+                shwrap("""
+                cosa generate-release-meta --build-id ${params.VERSION} --workdir .
+                cosa buildupload --build=${params.VERSION} --skip-builds-json \
+                    ${arch_args} s3 --aws-config-file=\${AWS_BUILD_UPLOAD_CONFIG} \
+                    --acl=public-read ${s3_stream_dir}/builds
+                """)
+
+            }
+
```
jlebon, line 133: Extraneous empty lines.

jobs/cloud_replicate.Jenkinsfile
```diff
@@ -0,0 +1,147 @@
...
+        currentBuild.result = 'SUCCESS'
+        currentBuild.description = "${build_description} ✓"
+
+// main try finishes here
+} catch (e) {
+    currentBuild.result = 'FAILURE'
+    throw e
+} finally {
+    if (currentBuild.result != 'SUCCESS') {
+        pipeutils.trySlackSend(color: 'danger', message: ":fcos: :bullettrain_front: :trashfire: cloud-replicate <${env.BUILD_URL}|#${env.BUILD_NUMBER}> [${params.STREAM}][${basearches.join(' ')}] (${params.VERSION})")
```
jlebon, line 145: Let's use a different emoji from the release job one? Maybe e.g. `:cloud:`?

jobs/release.Jenkinsfile
dustymabe: will fix

jobs/cloud_replicate.Jenkinsfile
dustymabe: will fix

jobs/cloud_replicate.Jenkinsfile
dustymabe: you're probably right, but we definitely wouldn't want two jobs with same stream/version to run at the same time (on accident), right?

jobs/cloud_replicate.Jenkinsfile
dustymabe: yeah maybe later.

jobs/cloud_replicate.Jenkinsfile
dustymabe: updated to match what jenkins prints for the stage name for the cloud replication.

jobs/cloud_replicate.Jenkinsfile
dustymabe: fixed

dustymabe:
Added a new commit that updates emoji use in for slack notifications.

jobs/cloud_replicate.Jenkinsfile
jlebon: Same version, no. Same stream... I guess it's uncommon but perfectly safe, so I wouldn't go out of my way to block it. E.g. maybe you've already started a replication job for a version but then another build comes through that supersedes it but the replicate job isn't done yet and it's too late to cancel it without leaking resources.

I guess the simplest patch is to do `s/STREAM/VERSION/`.

dustymabe: done

jlebon (approved):

PR #759, opened 2022-11-17
A few emoji updates for slack notifs

dustymabe:


utils.groovy
```diff
@@ -552,4 +552,10 @@ def get_supported_additional_arches() {
     return supported
 }
 
+// Get an emoji to use in slack notifications based on what version
+// of CoreOS we're operating on.
+def get_coreos_emoji(pipecfg) {
+    return pipecfg.streams.stable ? ':fcos:' : ':corehat:'
```
jlebon, line 558: That's a sneaky way of boiling it down to "FCOS or not". :)

I know this is just for a Slack emoji so it doesn't matter much, but I'm not a fan of the precedence it sets. I'd rather have it be configurable. One candidate would be in the `jenkins-config` configmap. So it'd be driven via `--deploy` and could default to `:coreos:`.

Also feels like we could fold this into `trySlackSend` to keep the callsites less verbose?

dustymabe: > That's a sneaky way of boiling it down to "FCOS or not". :)
> 
> I know this is just for a Slack emoji so it doesn't matter much, but I'm not a fan of the precedence it sets. I'd rather have it be configurable. One candidate would be in the `jenkins-config` configmap. So it'd be driven via `--deploy` and could default to `:coreos:`.

Not a big fan of controlling it via `--deploy` because someone has to remember to add the option and also remember what the magic string sequence is. How about something in the pipecfg?

> 
> Also feels like we could fold this into `trySlackSend` to keep the callsites less verbose?

hmm. The message is pretty customized for each use though. Do you have an example of what you are suggesting?

jlebon: > Not a big fan of controlling it via `--deploy` because someone has to remember to add the option and also remember what the magic string sequence is.

It could be codified in the `HACKING.md` steps, no?

> How about something in the pipecfg?

Yeah, I thought of the `pipecfg` too, but it feels a bit weird. It's not really something you'd change very often and feels more tied to the Jenkins installation. I guess it could make sense if we start supporting different emojis per-stream? Not sure if we'd ever really want that though.

Another approach would be an annotation on the Slack secret. I still have code laying around for that if you'd like. :) But OK with `pipecfg` if you'd prefer.

> hmm. The message is pretty customized for each use though. Do you have an example of what you are suggesting?

I mean removing the prefix from all the callsites and prefixing it with the configured emoji in `trySlackSend()`. That way, we know also that all messages are consistently prefixed.

dustymabe: ok I added a new commit that I think does what you were suggesting here.

jlebon (approved):

utils.groovy
```diff
@@ -535,6 +535,11 @@ def AWSBuildUploadCredentialExists() {
 // Emits Slack message if set up, otherwise does nothing.
 def trySlackSend(params) {
     if (utils.credentialsExist([string(credentialsId: 'slack-api-token', variable: 'UNUSED')])) {
+        // Prefix all slack messages with the emoji representing the platform
+        // either being the fcos pipeline or the rhcos pipeline. Default to
+        // just a generic CoreOS emoji if none given.
+        def emoji = pipecfg.misc?.slack_platform_emoji_string ?: ':coreos:'
```
jlebon, line 541: The `pipecfg` isn't available here. We could pass it in though it's less elegant and messes up the nice parameter passthrough we do in this function. That's where the `jenkins-config` ConfigMap would be nicer since it's akin to "global" state and we could just requery it cheaply by calling `load_jenkins_config()`.

Short-term I guess let's just pass it as a keyed param and fetch it here via `params['pipecfg']`?

dustymabe: ok pushed an update that will use an annotation on the secret instead

utils.groovy
```diff
@@ -532,9 +532,22 @@ def AWSBuildUploadCredentialExists() {
     return utils.credentialsExist(creds)
 }
 
+// Grabs the jenkins.io/emoji-prefix annotation from the slack-api-token
+def getSlackEmojiPrefix() {
+    return shwrapCapture("""
+        oc get secret -n ${env.PROJECT_NAME} -o json slack-api-token | \
+          jq -r '.metadata.annotations["jenkins.io/emoji-prefix"]'
```
jlebon, line 539: If we want to handle the missing case as the caller implies, one easy way to do it is to handle it in `jq` directly here. This should work:

```
jq -r '.metadata.annotations["jenkins.io/emoji-prefix"] // ":coreos:"'
```

jlebon: To clarify, the current approach to the fallback won't work. `jq` will return "null". So a more verbose approach that would keep the fallback logic in the caller would be to check for "null" in this function and return an empty string instead.

dustymabe: I think I see what you are saying. I'd prefer not to make the `jq` call any more complicated if we can.

Let's see if the latest upload addresses the concern.

jlebon (approved):

PR #767, opened 2022-11-21
config: drop support for `override_compression_format` hack

jlebon:
We don't need this anymore. On FCOS, the `xz` setting is part of `image.yaml` which `cosa compress` now knows to read from. On RHCOS, we rely on `cosa compress` defaulting to `gzip`. This works for both new and old RHCOS streams using a cosa without the `image.yaml` support.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -36,8 +36,6 @@ hacks:
   skip_plume_release_task: true
   # OPTIONAL: skip UEFI on older RHCOS
   skip_uefi_tests_on_older_rhcos: true
-  # OPTIONAL: use a different compression format when archiving
-  override_compression_format: gzip
 
 # OPTIONAL/TEMPORARY: whether to use `versionary` to derive version numbers
 versionary_hack: true
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -356,9 +356,7 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
         }
 
         stage('Archive') {
-            def format = pipecfg.hacks?.override_compression_format
-            format = format ?: 'xz' // Default to xz
-            shwrap("cosa compress --compressor ${format}")
+            shwrap("cosa compress")
 
             if (uploading) {
                 def acl = pipecfg.s3.acl ?: 'public-read'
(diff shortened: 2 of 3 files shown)


jlebon:
Requires https://github.com/coreos/fedora-coreos-config/pull/2097 landing in all FCOS streams.

jlebon:
Rebased! But we still need to hold this until https://github.com/coreos/fedora-coreos-config/pull/2097 reaches `stable`. Though we could also fast-track it.

dustymabe:
> Though we could also fast-track it.

:+1:

dustymabe:
> > Though we could also fast-track it.
> 
> +1

https://github.com/coreos/fedora-coreos-config/pull/2112

jlebon:
And for stable: https://github.com/coreos/fedora-coreos-config/pull/2113

dustymabe:
> And for stable: [coreos/fedora-coreos-config#2113](https://github.com/coreos/fedora-coreos-config/pull/2113)

`:)` I opened my PR for `testing` since the promotion will just overwrite whatever is in `stable`. I realize that was lazy of me because if we do an `ad-hoc` release we'd need to also cherry-pick this, but I was willing to live with that.

I assume you were wanting to account for that chase when you opened https://github.com/coreos/fedora-coreos-config/pull/2113 ?

jlebon:
> > And for stable: [coreos/fedora-coreos-config#2113](https://github.com/coreos/fedora-coreos-config/pull/2113)
> 
> `:)` I opened my PR for `testing` since the promotion will just overwrite whatever is in `stable`. I realize that was lazy of me because if we do an `ad-hoc` release we'd need to also cherry-pick this, but I was willing to live with that.
> 
> I assume you were wanting to account for that case when you opened [coreos/fedora-coreos-config#2113](https://github.com/coreos/fedora-coreos-config/pull/2113) ?

I didn't think of it so explicitly, but yes. Just having it everywhere we care about meant it was definitely safe to get this in. I assumed that's where you were going too but wanted the `testing` one to merge first. So I thought I just beat you to it. :)

Anyway, CI on https://github.com/coreos/fedora-coreos-config/pull/2113 is green now, so we might as well get it in? That way, if indeed there's an ad-hoc release we don't even have to think about this.

dustymabe:
> Anyway, CI on [coreos/fedora-coreos-config#2113](https://github.com/coreos/fedora-coreos-config/pull/2113) is green now, so we might as well get it in?

Merged!

PR #768, opened 2022-11-22
libcloud: only attempt cloud upload if config section exists

dustymabe:
We shoudn't even drop into the section if the configuration for it doesn't exist because it will fail anyway.


libcloud.groovy
```diff
@@ -119,7 +121,9 @@ def upload_to_clouds(pipecfg, basearch, buildID, stream) {
     }
     credentials = [file(variable: "AWS_BUILD_UPLOAD_CONFIG",
                         credentialsId: "aws-build-upload-config")]
-    if (artifacts.contains("aws") && utils.credentialsExist(credentials)) {
+    if (pipecfg.clouds?.aws
```
jlebon, line 124: Missing &&

jlebon (approved):
Yeah, I think this makes way more sense that way.

libcloud.groovy
dustymabe: fixed!

jlebon (approved):

PR #769, opened 2022-11-22
HACKING.md: add docs for triggering builds remotely

jlebon:
This will be used by the ART workflow at least.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -598,6 +598,23 @@ value from the `manifests/pipeline.yaml` OpenShift template. This is
 currently as designed (see
 [#65](https://github.com/coreos/fedora-coreos-pipeline/issues/65)).
 
+### Triggering builds remotely
+
+We use the OpenShift Login plugin for authentication. This plugin maps
+OpenShift users to Jenkins users, including service accounts. So we can
+use the `jenkins` service account (or really, any service account with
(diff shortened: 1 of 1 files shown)


jlebon:
A more restricted variation of this is to create a separate service account with *just* the permissions for starting builds. This would require providing a configmap to customize the OpenShift Login plugin as described in [this section](https://github.com/jenkinsci/openshift-login-plugin#openshift-role-to-jenkins-permission-mapping).

HACKING.md
```diff
@@ -598,6 +598,23 @@ value from the `manifests/pipeline.yaml` OpenShift template. This is
...
+the `edit` role):
+
+```
+token_secret=$(oc get sa jenkins -o json | jq -r '.secrets[] | select(.name | contains("token")) | .name')
+token_data=$(oc get secret "${token_secret}" -o=jsonpath={.data.token} | base64 -d)
+curl -H "Authorization: Bearer ${token_data}" $JENKINS_URL/job/build/buildWithParameters --data STREAM=4.13
+```
+
+See
+https://github.com/jenkinsci/openshift-login-plugin#non-browser-access
```
dustymabe, line 615: since this is markdown we could make this prettier behing a `[link](http://)`

PR #770, opened 2022-11-22
Add temporary hack for skipping AWS GovCloud uploads

jlebon:
RHCOS 4.9 and earlier did not upload to GovCloud so we need to temporarily support a hack to skip it on those streams.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -91,6 +91,10 @@ streams:
       # OPTIONAL: additional architectures to skip building for
       skip_additional_arches:
         - s390x
+      # OPTIONAL/TEMPORARY: don't upload to govcloud even if cloud is
+      # configured. Can be dropped when RHCOS 4.9 (last release that doesn't
+      # support GovCloud) is EOL.
+      skip_govcloud_hack: true
 
 # REQUIRED: architectures to build for other than x86_64
 additional_arches: [aarch64, ppc64le, s390x]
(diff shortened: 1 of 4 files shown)


libcloud.groovy
```diff
@@ -30,7 +30,11 @@ def replicate_to_clouds(pipecfg, basearch, buildID) {
     if (meta.amis) {
         replicators["☁️ 🔄:aws"] = {
             def replicated = false
-            def ids = ['aws-build-upload-config', 'aws-govcloud-image-upload-config']
+            def ids = ['aws-build-upload-config']
+            if (pipecfg.clouds?.aws?.govcloud &&
+                pipecfg[stream]?.skip_govcloud_hack != true) {
+                ids += 'aws-govcloud-image-upload-config'
+            }
```
dustymabe, line 37: I'm touching this code too over in https://github.com/coreos/fedora-coreos-pipeline/pull/771

We can probably merge this and then I'll rebase that.

docs/config.yaml
```diff
@@ -91,6 +91,8 @@ streams:
       # OPTIONAL: additional architectures to skip building for
       skip_additional_arches:
         - s390x
+      # OPTIONAL/TEMPORARY: don't upload to govcloud even if cloud is configured
```
dustymabe, line 94: would be nice to mention here when we can drop it. I know we'd be including information about RHCOS streams here, but it's always nice to leave ourselves breadcrumbs.

dustymabe (approved):
LGTM - a few comments

jlebon:
Updated!

dustymabe (approved):
LGTM

PR #771, opened 2022-11-23
libcloud: fixes for aws uploads

dustymabe:
```
commit 5a0a12039a6b126cf368643d185a051af166d747
Date:   Wed Nov 23 15:19:40 2022 -0500

    libcloud: specify the source region for AWS replication
    
    Since we can upload to both AWS and AWS Govcloud we don't need to
    just let the code pick the first AMI in the list to use as the
    source AMI. We need to tell it which one is the source.

commit 8e58bea89ccfe89c01e2fa5cc904636b2010a9a9
Date:   Wed Nov 23 15:15:36 2022 -0500

    libcloud: create a separate replicator for AWS govcloud
    
    We can run the replication for govcloud in parallel. Let's
    use the same model we're using for the uploaders where we
    define a closure and then call that closure for primary
    AWS and for AWS Govcloud.

commit 6ad8e2d0b7073218717dd0542a50771a1286245f
Date:   Wed Nov 23 15:14:47 2022 -0500

    libcloud: dedupe AWS upload calls
    
    Let's define a closure and then call that for AWS and AWS Govcloud.

```

### DIFF
--- a/libcloud.groovy
+++ b/libcloud.groovy
@@ -123,67 +123,53 @@ def upload_to_clouds(pipecfg, basearch, buildID, stream) {
             }
         }
     }
-    credentials = [file(variable: "AWS_BUILD_UPLOAD_CONFIG",
-                        credentialsId: "aws-build-upload-config")]
-    if (pipecfg.clouds?.aws &&
-        artifacts.contains("aws") &&
-        utils.credentialsExist(credentials)) {
-        def creds = credentials
-        uploaders["☁️ ⬆️ :aws"] = {
-            withCredentials(creds) {
-                utils.syncCredentialsIfInRemoteSession(["AWS_BUILD_UPLOAD_CONFIG"])
-                def c = pipecfg.clouds.aws
-                def extraArgs = []
-                if (c.test_accounts) {
-                    extraArgs += c.test_accounts.collect{"--grant-user=${it}"}
-                }
-                if (c.public) {
-                    extraArgs += "--public"
-                }
-                shwrap("""
-                cosa buildextend-aws \
-                    --upload \
-                    --arch=${basearch} \
-                    --build=${buildID} \
-                    --region=${c.primary_region} \
-                    --bucket=s3://${c.bucket} \
-                    --credentials-file=\${AWS_BUILD_UPLOAD_CONFIG} \
-                    ${extraArgs.join(' ')} 
-                """)
+
+    // For AWS we need to consider the primary AWS partition and the
+    // GovCloud partition. Define a closure here that we'll call for both.
+    def awsUploadClosure = { config, credentialId ->
+        def creds = [file(variable: "AWS_CONFIG_FILE", credentialsId: credentialId)]
+        withCredentials(creds) {
+            utils.syncCredentialsIfInRemoteSession(["AWS_CONFIG_FILE"])
+            def c = config
+            def extraArgs = []
+            if (c.test_accounts) {
+                extraArgs += c.test_accounts.collect{"--grant-user=${it}"}
+            }
+            if (c.public) {
+                extraArgs += "--public"
+            }
+            shwrap("""
+            cosa buildextend-aws \
+                --upload \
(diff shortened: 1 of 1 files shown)


dustymabe:
I still need to figure out a way to test this.

libcloud.groovy
```diff
@@ -30,22 +30,31 @@ def replicate_to_clouds(pipecfg, basearch, buildID) {
...
-                    replicated = true
+            def aws_info = [[pipecfg.clouds.aws, 'aws-build-upload-config'],
+                            [pipecfg.clouds.aws.govcloud, 'aws-govcloud-image-upload-config']]
+            for (item in aws_info) {
+                def (c, id) = item
+                // Only try to replicate if the config exists
+                if (c) {
+                    tryWithCredentials([file(variable: 'AWS_CONFIG_FILE', credentialsId: id)]) {
+                        def extraArgs = []
+                        if (c.primary_region) {
```
jlebon, line 41: This is a required field, so we can just always do it, no?

libcloud.groovy
```diff
@@ -30,22 +30,31 @@ def replicate_to_clouds(pipecfg, basearch, buildID) {
...
-                        --credentials-file=\${AWS_CONFIG_FILE} \
-                        --log-level=INFO
-                    """)
-                    replicated = true
+            def aws_info = [[pipecfg.clouds.aws, 'aws-build-upload-config'],
+                            [pipecfg.clouds.aws.govcloud, 'aws-govcloud-image-upload-config']]
+            for (item in aws_info) {
+                def (c, id) = item
+                // Only try to replicate if the config exists
+                if (c) {
```
jlebon, line 38: Minor: could invert this and `continue` early to avoid an indentation level.

jlebon (commented):
Are you sure the second commit is required? The meta merging logic supports adding subkeys to the same parent key. We rely on this already for `images` during the `buildextend` calls.

dustymabe:
added a new commit to fix `skip_govcloud_hack`.

libcloud.groovy
```diff
@@ -27,32 +27,42 @@ def replicate_to_clouds(pipecfg, basearch, buildID, stream) {
...
+    def awsReplicateClosure = { config, credentialId ->
+        def creds = [file(variable: "AWS_CONFIG_FILE", credentialsId: credentialId)]
+        withCredentials(creds) {
+            def c = config
+            shwrap("""
+            cosa aws-replicate \
+                --log-level=INFO \
+                --build=${buildID} \
+                --arch=${basearch} \
+                --credentials-file=\${AWS_CONFIG_FILE} \
```
jlebon, line 43: Extra \ here

jlebon (approved):

libcloud.groovy
dustymabe: ok fixed that and added one more commit!

libcloud.groovy
dustymabe: fixed

libcloud.groovy
dustymabe: see https://github.com/coreos/fedora-coreos-pipeline/pull/771#issuecomment-1325617342

libcloud.groovy
```diff
@@ -97,7 +97,7 @@ def replicate_to_clouds(pipecfg, basearch, buildID, stream) {
 def upload_to_clouds(pipecfg, basearch, buildID, stream) {
 
     // Get a list of the artifacts that are currently built.
-    def images_json = readJSON(text: shwrapCapture("cosa meta --get-value images"))
+    def images_json = readJSON(text: shwrapCapture("cosa meta --arch=${basearch} --get-value images"))
```
jlebon, line 100: Really, we should also include `--build ${buildID}` in this call since the function is structured as looking generic over any build. And similarly for the replication code. Not new, so not a blocker!

dustymabe: added a commit for that

jlebon (approved):

PR #772, opened 2022-11-23
libcloud: also grant `createVolumePermission` to grant_users

jlebon:
We do want this for AWS Marketplace accounts at least. For test
accounts, it's not necessary but doesn't hurt either.


dustymabe (approved):
LGTM

PR #773, opened 2022-11-24
config: drop `skip_plume_release_task` hack, add `misc.generate_release_index`

jlebon:
Rather than `plume release`, use the new `plume make-amis-public` and
`plume update-release-index` commands directly. Gate the former on
whether the AMIs were already made public on first upload. Gate the
latter on a new `misc.generate_release_index` since we only need it for
FCOS. Finally, get rid of the `skip_plume_release_task` knob.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -32,8 +32,6 @@ hotfix:
 hacks:
   # OPTIONAL: skip upgrade kola tests
   skip_upgrade_tests: true
-  # OPTIONAL: skip calling plume in the release job
-  skip_plume_release_task: true
   # OPTIONAL: skip UEFI on older RHCOS
   skip_uefi_tests_on_older_rhcos: true
   # OPTIONAL: implement the yumrepos branch workaround
@@ -142,7 +140,8 @@ clouds:
     # REQUIRED (if AWS build upload credentials provided): the bucket+prefix
     # where image files are uploaded to then initiate an image import.
     bucket: fedora-coreos-image-uploads/ami-import
-    # OPTIONAL: boolean: whether or not to set the image/snapshot as public
+    # OPTIONAL: boolean: whether or not to set the image/snapshot as public on
+    # initial upload (AMIs of released builds are always made public)
     public: true
     # OPTIONAL: list of architectures to limit cloud testing to.
     test_architectures: [x86_64]
@@ -229,3 +228,8 @@ clouds:
       - jp-osa
       - jp-tok
       - us-south
+
+# OPTIONAL: miscellaneous options
+misc:
+  # OPTIONAL: whether to generate a release index
+  generate_release_index: true
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -298,19 +298,25 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
                     --acl=${acl} ${s3_stream_dir}/builds
                 """)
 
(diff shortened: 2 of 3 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/3240

jobs/release.Jenkinsfile
```diff
@@ -298,19 +298,25 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
-                    // We'll make this more explicit in the future.
+                def bucket_prefix = "${pipecfg.s3.bucket}/${pipecfg.s3.builds_key}"
+
+                // make AMIs public if not already the case
+                if (!pipecfg.clouds?.aws?.public) {
+                    shwrap("""
+                    cosa shell -- plume make-amis-public \
+                        --version ${params.VERSION} \
+                        --stream ${params.STREAM} \
+                        --bucket-prefix ${bucket_prefix} \
```
dustymabe, line 309: I still need to review the COSA PR, but just from scrolling this it seems odd that a call to make AMIs public needs info about an s3 bucket.

jlebon: It reads the release metadata directly from the bucket. `plume` commands traditionally are not "cosa workdir-aware", so I'd prefer not to have it start looking for a `builds/` dir there. But what we could do is hand it the release metadata file directly ourselves. That should actually allow us to drop the `--version` and `--stream` switches too.

I'll see about enhancing it for that, but will do that as a follow-up to https://github.com/coreos/coreos-assembler/pull/3240 so that we can get rid of `plume release` first to not support the old way anymore.

docs/config.yaml
```diff
@@ -110,6 +108,9 @@ s3:
   # OPTIONAL: object ACL to use when uploading
   acl: private
 
+# OPTIONAL: whether to generate a release index
+release_index: true
```
dustymabe, line 112: maybe we can add this under a new `misc` section and collect other loose items under that? I had started to do this in https://github.com/coreos/fedora-coreos-pipeline/pull/759 (see [this](https://github.com/coreos/fedora-coreos-pipeline/compare/35cfa7dc3cd8c7052eedf7eb27079b654f60c1bb..cb1cfa71d8cc56f8fba0fe79c87d8a451f0937e6) force push), but we ended up going a different way.

TL;DR - I think things are starting to get a bit disorganized and having a misc "things that don't fit anywhere else" section might be nice.

jobs/release.Jenkinsfile
```diff
@@ -298,19 +298,25 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
+                        --aws-credentials \${AWS_BUILD_UPLOAD_CONFIG}
```
dustymabe, line 310: do we need to consider govcloud here?

jlebon: Yeah, good point. I guess we will have to enhance `make-amis-public` to also take a GovCloud credentials and use it when iterating on a GovCloud region. It's not immediately necessary since we already make the AMIs public at upload so we won't fall in this branch, but if that ever changes we would. Will track that in a cosa issue so we don't forget it.

jlebon:
Updated!

dustymabe (approved):
LGTM

jobs/release.Jenkinsfile
jlebon: Opened https://github.com/coreos/coreos-assembler/issues/3259.

PR #775, opened 2022-11-25
Bake additional cert into Jenkins controller image

jlebon:
Just like we did for the agent image, play a similar trick to have the
additional cert baked into the controller image. This is required if we
need to fetch resources from HTTPS URLs covered by the root CA in code
that runs directly on the controller. This situation comes up in RHCOS
where we want to be able to load the pipecfg from an internal repo early
on in jobs.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -563,10 +563,10 @@ oc start-build --follow jenkins-with-cert
 oc start-build --follow jenkins-agent-base-with-cert
 ```
 
-We can now start a build of the Jenkins controller:
+We can now start an S2I build of the Jenkins controller:
 
 ```
-oc start-build --follow jenkins
+oc start-build --follow jenkins-s2i
 ```
 
 Once the Jenkins controller image is built, Jenkins should start up (verify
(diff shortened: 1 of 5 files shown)


jlebon:
I like the `with-cert` idea. It's a bit wordy, but clear. I find the `layer` wording confusing though. I think it'd be clearer if we remain consistent with the canonical OpenShift images? The two images are `jenkins` and `jenkins-agent-base`. So WDYT about:
- `jenkins-with-cert`
- `jenkins-agent-base-with-cert`
?

IOW, this is saying "these are the same as the regular OpenShift images, except they have a cert added". And then we rename our S2I buildconfig and output to `jenkins-s2i`. It's confusing to have it named just `jenkins`.

And I'll add a diagram somewhere showing how they're connected.

dustymabe:
yeah let's try that and see what it looks like in the next upload.

jlebon:
Updated this now with better naming and a diagram!

dustymabe:
Looks much better now. Thanks for the diagram!

dustymabe (approved):
LGTM

PR #776, opened 2022-11-30
jobs/build: adapt to new hashlist command

jlebon:
We've made hashlist generation look more like just another buildextend command. This allows us to clean up our code related to this.

See: https://github.com/coreos/coreos-assembler/pull/3248

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -52,6 +52,7 @@ default_artifacts:
     - metal4k
     - live
     - openstack
+    - hashlist-experimental
   aarch64:
     - azure
   ppc64le:
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -276,29 +276,10 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
             }
         }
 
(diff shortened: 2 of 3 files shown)


jobs/build-arch.Jenkinsfile
```diff
@@ -376,14 +357,16 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
             }
         }
 
+        // A few independent tasks that can be run in parallel
+        def parallelruns = [:]
+
```
dustymabe, line 362: can probably even bury this inside the if statement below since the usage is only in that scope now.

config.yaml
```diff
@@ -52,6 +52,7 @@ default_artifacts:
     - metal4k
     - live
     - openstack
+    - hashlist-exp
```
dustymabe, line 55:
```suggestion
    - experimental-hashlist
```

If we end up doing https://github.com/coreos/coreos-assembler/pull/3248#issuecomment-1332758075

dustymabe (approved):
LGTM

dustymabe (approved):
LGTM

PR #777, opened 2022-12-01
manifests/jenkins: bump memory request to 3Gi

jlebon:
We're hitting https://issues.jenkins.io/browse/JENKINS-55287 which according to some reports can be memory-related. Let's try bumping to 3Gi and see if it helps.


dustymabe (approved):
LGTM

jlebon:
I've rolled this out to the FCOS pipeline now.

PR #779, opened 2022-12-01
HACKING.md: tweak remote triggering instructions

jlebon:
For some reason, sometimes the token secret isn't in the `oc get sa` output. Instead, we need to do what `oc describe sa` does and filter the list of secrets by type and name and uid to get the list of token secrets we can use.

I know the `jq` command there is a mouthful. There is a shorter `oc describe | grep | cut` version available but I'd like this to also serve as documentation for how a compiled tool could e.g. fetch the list of secrets itself and do the same filtering that we do here with `jq` without resorting to hacky string manipulations.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -608,8 +608,12 @@ use the `jenkins` service account (or really, any service account with
 the `edit` role):
 
 ```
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
Trivial LGTM

PR #781, opened 2022-12-02
manifests/jenkins: set resource limits and requests for jnlp container

jlebon:
Right now, we're letting it default to the default values of the limitrange, which is 4 for CPUs and 10G for RAMs. That's vastly overkill given that we don't actually do any work in them but rather in the cosa containers colocated in them.

The limits chosen here match the builtin defaults in the kubernetes-plugin.


dustymabe (approved):
LGTM

PR #783, opened 2022-12-06
jenkins/config: define kubernetes cloud ourselves and bump maxRequestsPerHost to 64

jlebon:
```
jenkins/config: define kubernetes cloud ourselves

Rather than relying on the S2I run script to define the kubernetes
cloud, do it ourselves. This will allow us to further customize it for
our needs, which isn't currently possible with the S2I setup:

https://github.com/openshift/jenkins/issues/1505

The two major differences with what the S2I run script would've defined
are:
1. no defining of pod templates; we don't use those anyway, so this is a
   nice cleanup
2. no setting of the kube certificate; the kubernetes plugin is smart
   enough to get it from the mounted serviceaccount secret, which is
   what the S2I run script does anyway
```
---
```
jenkins/config: bump maxRequestsPerHost to 64

This is a conceptual revert of the revert in #583. Now that we own the
cloud config, we can safely do this.
```

### DIFF
--- a/jenkins/config/k8s.yaml
+++ b/jenkins/config/k8s.yaml
@@ -1,7 +1,10 @@
 jenkins:
   clouds:
   - kubernetes:
-      # The keys below match
+      # mitigation for
+      # https://docs.cloudbees.com/docs/cloudbees-ci-kb/latest/client-and-managed-masters/considerations-for-kubernetes-clients-connection-when-using-kubernetes-plugin
+      maxRequestsPerHost: 64
+      # The remaining keys below match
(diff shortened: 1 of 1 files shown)


jlebon:
Tested working in stg.

jlebon:
Just did a minor `s/JENKINS_JNLP_SERVICE_PORT_AGENT/JENKINS_JNLP_SERVICE_PORT/`. The previous one was correct too, but the new one is what the S2I run script uses, so let's match it.

jlebon:
I've rolled this out to FCOS stg/prod and RHCOS devel/prod.

Issue #784, opened 2022-12-07
Consider adding "Warning" to release job slack notification if 1 or more expected arches are missing


jschintag:
Currently the release job reports which arches where picked up during the job (`[aarch64 ppc64le s390x x86_64]`).

Since it is possible that one or more arches fail during the build-job, the release job should emit a warning that something is wrong, should it not pick up all 4 arches. This would enable us to faster spot Problems through the slack notifications, as we could see the status of all arch builds in the notification of the release job notification.

PR #786, opened 2022-12-12
jobs/release: show warning status if missing architectures

dustymabe:
This will make the job (and slack message) show up as yellow (warning) if there were missing architectures.

Fixes https://github.com/coreos/fedora-coreos-pipeline/issues/784

### DIFF
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -375,17 +375,5 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
     currentBuild.result = 'FAILURE'
     throw e
 } finally {
-    def color
-    switch(currentBuild.result) {
-        case 'SUCCESS':
-            color = 'good';
-            break;
-        case 'UNSTABLE':
-            color = 'warning';
-            break;
-        default:
-            color = 'danger';
-            break;
-    }
-    pipeutils.trySlackSend(color: color, message: ":bullettrain_front: release <${env.BUILD_URL}|#${env.BUILD_NUMBER}> [${params.STREAM}][${basearches.join(' ')}] (${params.VERSION})")
+    pipeutils.trySlackSend(message: ":bullettrain_front: release <${env.BUILD_URL}|#${env.BUILD_NUMBER}> [${params.STREAM}][${basearches.join(' ')}] (${params.VERSION})")
 }}} // try-catch-finally, cosaPod and lock finish here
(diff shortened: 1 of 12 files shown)


dustymabe:
testing looks good!

jobs/release.Jenkinsfile
```diff
@@ -111,7 +111,11 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
         assert builtarches.contains("x86_64"): "The x86_64 architecture was not in builtarches."
         if (!builtarches.containsAll(basearches)) {
             if (params.ALLOW_MISSING_ARCHES) {
-                echo "Some requested architectures did not successfully build! Continuing."
+                // Display a warning, but continue.
+                def message = "Some requested architectures did not successfully build!"
+                warnError(message: message) {
+                    error(message)
+                }
```
jlebon, line 118: Feels like we're using this API wrong. I think this is equivalent to just:

```suggestion
                echo "Some requested architectures did not successfully build!"
                currentBuild.result = 'UNSTABLE'
```

jobs/release.Jenkinsfile
```diff
@@ -371,6 +375,17 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
     currentBuild.result = 'FAILURE'
     throw e
 } finally {
-    def color = currentBuild.result == 'SUCCESS' ? 'good' : 'danger'
+    def color
+    switch(currentBuild.result) {
```
jlebon, line 379: I wonder if we should fold this into `trySlackSend()`. I think we want this in every call site. If we ever have a caller in the future that wants to control the colour not based on build status, it could override the color parameter.

jobs/release.Jenkinsfile
dustymabe: Not exactly. With `warnError` the message gets highlighted in the web view versus you having to search for it. See the following examples:

- `warnError()`: [link](https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/blue/organizations/jenkins/release/detail/release/271/pipeline)
- `result='UNSTABLE'`: [link](https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/blue/organizations/jenkins/release/detail/release/272/pipeline/)

jobs/release.Jenkinsfile
dustymabe: let me see what I can do

dustymabe: added a new commit on top (and one other one too)

jlebon (approved):

PR #787, opened 2022-12-12
jobs/build: add support for reading variant from stream/pipeline config

marmijo:
Add support in `build.jenkinsfile` and `build-arch.jenkinsfile` for reading variant information from the pipeline config/stream and passing it into `cosa init`.

xref: https://gitlab.cee.redhat.com/coreos/rhcos-devel-pipecfg/-/merge_requests/3  &  https://gitlab.cee.redhat.com/coreos/rhcos-art-pipecfg/-/merge_requests/9

See [COS-1925](https://issues.redhat.com/browse/COS-1925)

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -172,8 +172,9 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
                 """)
             } else {
                 def yumrepos = pipecfg.source_config.yumrepos ? "--yumrepos ${pipecfg.source_config.yumrepos}" : ""
+                def variant = stream_info.variant ? "--variant ${stream_info.variant}" : ""
                 shwrap("""
-                cosa init --force --branch ${ref} --commit=${src_config_commit} ${yumrepos} ${pipecfg.source_config.url}
+                cosa init --force --branch ${ref} --commit=${src_config_commit} ${yumrepos} ${variant} ${pipecfg.source_config.url}
                 """)
             }
 
--- a/jobs/cloud-replicate.Jenkinsfile
+++ b/jobs/cloud-replicate.Jenkinsfile
@@ -87,8 +87,9 @@ lock(resource: "cloud-replicate-${params.VERSION}") {
         // Fetch metadata files for the build we are interested in
         stage('Fetch Metadata') {
             def ref = pipeutils.get_source_config_ref_for_stream(pipecfg, params.STREAM)
+            def variant = stream_info.variant ? "--variant ${stream_info.variant}" : ""
             pipeutils.shwrapWithAWSBuildUploadCredentials("""
-            cosa init --branch ${ref} ${pipecfg.source_config.url}
+            cosa init --branch ${ref} ${variant} ${pipecfg.source_config.url}
             cosa buildfetch --build=${params.VERSION} \
                 --arch=all --url=s3://${s3_stream_dir}/builds \
                 --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG}
--- a/jobs/kola-openstack.Jenkinsfile
+++ b/jobs/kola-openstack.Jenkinsfile
@@ -41,6 +41,8 @@ properties([
 
 currentBuild.description = "[${params.STREAM}][${params.ARCH}] - ${params.VERSION}"
 
+def stream_info = pipecfg.streams[params.STREAM]
+
 // Use ca-ymq-1 for everything right now since we're having trouble with
 // image uploads to ams1.
 def region = "ca-ymq-1"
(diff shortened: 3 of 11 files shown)


travier:
Let's not mention any specific variant name in this PR. Something like "Add support for reading variant from the stream/pipecfg config"

travier (approved):

jobs/build-arch.Jenkinsfile
```diff
@@ -172,8 +172,9 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
                 """)
             } else {
                 def yumrepos = pipecfg.source_config.yumrepos ? "--yumrepos ${pipecfg.source_config.yumrepos}" : ""
+                def variant = stream_info.variant ? "--variant ${stream_info.variant}" : ""
                 shwrap("""
-                cosa init --force --branch ${ref} --commit=${src_config_commit} ${yumrepos} ${pipecfg.source_config.url}
+                cosa init --force --branch ${ref} --commit=${src_config_commit} ${yumrepos} ${variant} ${pipecfg.source_config.url}
```
jlebon, line 177: I think we need this also at least in all the kola jobs, since it can affect denylisting. But we might as well add it in all the jobs where we do `cosa init` for consistency (e.g. I don't *think* it currently would matter for the `release` job, but it may in the future and we'd easily regress on this).

jlebon: Actually, looking at the kola harness code, I think it would do the right thing since it'd look at the variant of the build and not the currently configured variant.

I think if we do https://github.com/coreos/coreos-assembler/pull/3274/files#r1047415430, then it's easy to tell that it only matters in jobs that actually `cosa build`. This would include `bump-lockfile` too. It's not relevant for RHCOS yet, but (hopefully) will be in the future.

travier: Not sure I understand. You don't need to specify the variant in kola as it picks it up from the `src/config.json` file

jlebon: We run follow-up kola cloud tests in separate jobs so this initialization does not carry. But yes, indeed I misread the harness code. So then we definitely do also need this in the kola jobs too in the short-term. Updated proposal above!

But I think the right thing to do for kola there is also to read it from the build metadata (and doing https://github.com/coreos/coreos-assembler/pull/3274#discussion_r1047415430).

travier: https://github.com/coreos/coreos-assembler/pull/2934/files#diff-9db99bb7fefa41cf89f75a965efa109bc8349ab0a55fdb093fa46775724791cbR354 > Kola looks at `config.json` right now so we'll need to fix this for kola if we don't pass the init'ed repo to kola runs.

travier: Proposal in https://github.com/coreos/coreos-assembler/issues/3276

travier (approved):

jobs/cloud-replicate.Jenkinsfile
```diff
@@ -35,6 +35,8 @@ properties([
 
 def build_description = "[${params.STREAM}]"
 
+def stream_info = pipecfg.streams[params.STREAM]
```
jlebon, line 38: This is already defined lower down.

jobs/kola-openstack.Jenkinsfile
```diff
@@ -41,6 +41,8 @@ properties([
 
 currentBuild.description = "[${params.STREAM}][${params.ARCH}] - ${params.VERSION}"
 
+def variant = stream_info.variant ? "--variant ${stream_info.variant}" : ""
```
jlebon, line 44: I suppose you meant to define `stream_info` here?

jobs/release.Jenkinsfile
```diff
@@ -42,6 +42,8 @@ properties([
 
 def build_description = "[${params.STREAM}]"
 
+def stream_info = pipecfg.streams[params.STREAM]
```
jlebon, line 45: This is already defined lower down.

marmijo: Fixed

jobs/kola-openstack.Jenkinsfile
marmijo: Fixed

jobs/cloud-replicate.Jenkinsfile
marmijo: Fixed

jlebon (approved):
LGTM! (Just pushed a small update to avoid some unnecessary whitespace changes.)

travier:
🎉

PR #788, opened 2022-12-13
jobs/build: don't include build result as string in Slack msg

jlebon:
It's redundant with the emojis and colour and isn't consistent with the other job notifications.


dustymabe (approved):
LGTM

PR #789, opened 2022-12-15
jobs/release: don't block release if some AWS regions are down

jlebon:
Use the new exit code added in
https://github.com/coreos/coreos-assembler/pull/3277 to detect the case where some regions are down. If that happens, continue with the release, but issue a warning, which will mark the build as unstable.

### DIFF
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -307,13 +307,23 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
 
                 // make AMIs public if not already the case
                 if (!pipecfg.clouds?.aws?.public) {
-                    shwrap("""
+                    def rc = shwrapRc("""
                     cosa shell -- plume make-amis-public \
                         --version ${params.VERSION} \
                         --stream ${params.STREAM} \
                         --bucket-prefix ${bucket_prefix} \
                         --aws-credentials \${AWS_BUILD_UPLOAD_CONFIG}
                     """)
+                    // see https://github.com/coreos/coreos-assembler/pull/3277
+                    if (rc == 77) {
+                        // Display a warning, but continue.
+                        def message = "Failed to make AMIs public in some regions"
+                        warnError(message: message) {
+                            error(message)
+                        }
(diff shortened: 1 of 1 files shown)


jobs/release.Jenkinsfile
```diff
@@ -307,13 +307,21 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
                         --aws-credentials \${AWS_BUILD_UPLOAD_CONFIG}
                     """)
+                    // see https://github.com/coreos/coreos-assembler/pull/3277
+                    if (rc == 77) {
+                        // Display a warning, but continue.
+                        def message = "Failed to make AMIs public in some regions"
+                        warnError(message: message) {
+                            error(message)
+                        }
+                    }
```
dustymabe, line 324: I feel like we're missing a case here. For example, what if the rc isn't `77` or `0`? We want to `error()` in that case.

jlebon:
> With your https://github.com/coreos/coreos-assembler/pull/3277 the public AMIs bit is best effort so at least only the failing regions will be the ones missing and we can decide to comment out the public region bits in the followup release job if we want to override.

Right, that'd be the base workflow. I was trying to aim for something a little less reactive. :)

> That being said, I'm not against this change, I just want to make sure someone actually takes action if a region is missed and if we continue the rest of the job it's possible they won't.

Right, yeah. So this is relying more on the "unstable" convention we've recently established. The release job at this point is mostly a series of unrelated tasks to update different endpoints. Once we commit to releasing a version, I think we should try to run as much of it as possible to be as consistent as we can across these endpoints.

I assumed the release job being unstable (and showing up as yellow rather than red) would be enough for releasers to question if something went wrong. I guess we could be more explicit in the release checklist about this? Alternatively though, we could also keep going but mark the build as failed at the end.

dustymabe:
I think i'm OK with it as written. Just wanted to have the discussion.

Still interested in your comments on https://github.com/coreos/fedora-coreos-pipeline/pull/789#discussion_r1049772725

dustymabe (approved):
LGTM

PR #790, opened 2022-12-15
release: Add `time -v` as part of fetch cmd for debugging

ravanelli:
- We are having trouble detecting why fetch is failing from time to time. To track it a little better, let's add it for a while.


jlebon (approved):
Thanks!

dustymabe:
do all the older RHCOS branches have `time` in COSA?

jlebon:
> do all the older RHCOS branches have `time` in COSA?

Good catch. Looking at older branches, it doesn't seem like it. Let's temporarily revert this for now so it doesn't break the release job on older RHCOS: https://github.com/coreos/fedora-coreos-pipeline/pull/791.

jlebon:
This is re-added in https://github.com/coreos/fedora-coreos-pipeline/pull/837 now that it's in all cosa branches.

PR #791, opened 2022-12-15
Revert "release: Add `time -v` as part of fetch cmd for debugging"

jlebon:
This reverts commit 0358b7624c09c30ae111819fa70af86f6764c3bd.

We need to backport https://github.com/coreos/coreos-assembler/pull/3268 to older cosa first.


cgwalters (approved):
Looks fine.  Though, couldn't we just be using the `time` builtin bash function instead of `/usr/bin/time`?

jlebon:
> Looks fine. Though, couldn't we just be using the `time` builtin bash function instead of `/usr/bin/time`?

What we're investigating is peak memory usage because sometimes `buildfetch` seems like it gets OOM-killed even though it should have sufficient memory. Bash's `time` builtin doesn't support printing memory stats.

PR #794, opened 2022-12-22
jobs/build-arch: use new autolock feature

jlebon:
This makes use of the new autolock feature in cosa to force multi-arch
builds to be consistent with the x86_64 build. This takes effect when
building a stream which doesn't currently use lockfiles.

The primary use case is RHCOS, but even once we start using lockfiles
there, this will still be useful for the FCOS rawhide stream.

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -11,7 +11,6 @@ node {
 // Base URL through which to download artifacts
 BUILDS_BASE_HTTP_URL = "https://builds.coreos.fedoraproject.org/prod/streams"
 
-
 properties([
     pipelineTriggers([]),
     parameters([
@@ -78,7 +77,7 @@ def stream_info = pipecfg.streams[params.STREAM]
 def cosa_controller_img = stream_info.cosa_controller_img_hack ?: cosa_img
 
 // If we are a mechanical stream then we can pin packages but we
-// don't maintin complete lockfiles so we can't build in strict mode.
+// don't maintain complete lockfiles so we can't build in strict mode.
 def strict_build_param = stream_info.type == "mechanical" ? "" : "--strict"
 
 // Note that the heavy lifting is done on a remote node via podman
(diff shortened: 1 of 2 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/3297

jobs/build-arch.Jenkinsfile
```diff
@@ -187,6 +193,17 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
...
+                // Also fetch the x86_64 lockfile for the build we're
+                // complementing, if any. Usually, this is the same build we
+                // just fetched above, but not necessarily.
+                pipeutils.shwrapWithAWSBuildUploadCredentials("""
+                cosa buildfetch --arch=x86_64 \
+                    --build ${newBuildID} \
+                    --file manifest-lock.generated.x86_64.json \
+                    --url s3://${s3_stream_dir}/builds \
+                    --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG}
+                cosa shell -- ls -lR builds/
```
dustymabe, line 205: Do we intend to remove the `cosa shell -- ls -lR builds/` before merging this?

dustymabe (commented):
I know this is still in draft, but this mostly LGTM

jobs/build-arch.Jenkinsfile
jlebon: Yeah, leftover from debugging. :)

jlebon:
This requires rpm-ostree patches and we won't/can't update rpm-ostree in older cosa. So we'll have to do this for the latest RHCOS only.

Rather than adding another config knob, I'm thinking we auto-detect if autolocking is supported by checking the output of `cosa fetch --help`.

dustymabe:
> Rather than adding another config knob, I'm thinking we auto-detect if autolocking is supported by checking the output of `cosa fetch --help`.

WFM

jlebon:
This is ready now!

PR #795, opened 2022-12-22
jobs/build: add `WAIT_FOR_RELEASE_JOB` parameter

jlebon:
In RHCOS, the pipeline is mainly driven by higher-level tools that
orchestrate a larger pipeline. In that context, it's not helpful to have
the `build` job just trigger the `build-arch` jobs and `release` jobs
and exit. Automation would have to independently track the state of the
triggered jobs and wait for them. It's much simpler to be able to wait
for just the `build` job.

This patch adds a new `WAIT_FOR_RELEASE_JOB` parameter which makes the
`build` job wait for the `release` job to finish, propagating failures
accordingly.

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -459,22 +459,8 @@ lock(resource: "build-${params.STREAM}") {
 
         // For now, we auto-release all non-production streams builds. That
         // way, we can e.g. test testing-devel AMIs easily.
-        //
-        // Since we are only running this stage for non-production (i.e. mechanical
-        // and development) builds we'll default to not doing AWS AMI replication.
-        // We'll also default to allowing failures for additonal architectures.
         if (uploading && stream_info.type != "production") {
-            stage('Publish') {
-                build job: 'release', wait: false, parameters: [
-                    string(name: 'STREAM', value: params.STREAM),
-                    string(name: 'ADDITIONAL_ARCHES', value: params.ADDITIONAL_ARCHES),
-                    string(name: 'VERSION', value: newBuildID),
-                    booleanParam(name: 'ALLOW_MISSING_ARCHES', value: true),
-                    booleanParam(name: 'CLOUD_REPLICATION', value: params.CLOUD_REPLICATION),
-                    string(name: 'PIPECFG_HOTFIX_REPO', value: params.PIPECFG_HOTFIX_REPO),
-                    string(name: 'PIPECFG_HOTFIX_REF', value: params.PIPECFG_HOTFIX_REF)
-                ]
-            }
+            run_release_job(newBuildID)
         }
 
         currentBuild.result = 'SUCCESS'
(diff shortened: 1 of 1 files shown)


jlebon:
Requires: https://github.com/coreos/fedora-coreos-pipeline/pull/794

jobs/build.Jenkinsfile
```diff
@@ -458,25 +480,23 @@ lock(resource: "build-${params.STREAM}") {
...
-                    string(name: 'ADDITIONAL_ARCHES', value: params.ADDITIONAL_ARCHES),
-                    string(name: 'VERSION', value: newBuildID),
-                    booleanParam(name: 'ALLOW_MISSING_ARCHES', value: true),
-                    booleanParam(name: 'CLOUD_REPLICATION', value: params.CLOUD_REPLICATION),
-                    string(name: 'PIPECFG_HOTFIX_REPO', value: params.PIPECFG_HOTFIX_REPO),
-                    string(name: 'PIPECFG_HOTFIX_REF', value: params.PIPECFG_HOTFIX_REF)
-                ]
-            }
-        }
+        publish(newBuildID)
```
dustymabe, line 499: did we want to wait on this one too?

dustymabe (commented):
I'm a bit lukewarm on this proposal. I'm not convinced the added complexity is really worth the changes. As you mention in the commit message we also have to handle all the slack messages from the various jobs.

My interpretation is that if you want this functionality you essentially only want to get one notification (slack message) about a build and it needs to be either success or failed.

Maybe let's discuss more in IRC on the merits here.

cgwalters (commented):
(I don't know this code, but superfically) LGTM

jlebon:
Rebased and updated this to a simpler approach that doesn't require https://github.com/coreos/coreos-ci-lib/pull/140.

dustymabe:
> Rebased and updated this to a simpler approach that doesn't require [coreos/coreos-ci-lib#140](https://github.com/coreos/coreos-ci-lib/pull/140).

waiting on the release job in the main job like this should work but there is some subtle behavior going on underneath that we have to be careful with.

For example currently the release job takes locks on:

```
def locks = basearches.collect{[resource: "release-${params.VERSION}-${it}"]}
```

which includes `x86_64`. Now if the `build` job also took the `release-VERSION-x86_64` lock like the `build-arch` jobs do with `lock(resource: "release-${params.VERSION}-${basearch}")` then the release job would wait forever.

We'll just keep this stuff in mind as we go.

dustymabe (approved):
LGTM

jlebon:
> For example currently the release job takes locks on:
> 
> ```
> def locks = basearches.collect{[resource: "release-${params.VERSION}-${it}"]}
> ```
>
> which includes `x86_64`. Now if the `build` job also took the `release-VERSION-x86_64` lock like the `build-arch` jobs do (with `lock(resource: "release-${params.VERSION}-${basearch}")`) then the release job would wait forever.

Indeed. I don't expect us to change this behaviour anytime soon but we should probably add a comment anyway to make that assumption explicit.

dustymabe:
One thing I just thought of.. This is probably going to affect drastically how much time we need to alot for timing out.

jlebon:
Updated to document this and also document a race I think is unlikely.

> One thing I just thought of.. This is probably going to affect drastically how much time we need to alot for timing out.

Good point. The container part of the release job itself isn't slow, but waiting for the multi-arch jobs is. I guess we need to adjust the timeout when the parameter is set.

dustymabe:
> I guess we need to adjust the timeout when the parameter is set.

or somehow move the release job to outside the timeout

jlebon:
> > I guess we need to adjust the timeout when the parameter is set.
> 
> or somehow move the release job to outside the timeout

Yeah, thought of that too though it makes things harder to follow. I'd also prefer still keeping it all under a timeout if possible. I'd say let's start with the simple approach and adjust as needed in follow-ups? (Relatedly, I noticed the release job doesn't currently have a timeout. It probably should, even if it's a really high number to account for optional cloud replication.)

jlebon:
Folding this into https://github.com/coreos/fedora-coreos-pipeline/pull/795 since I'm testing it all together.

dustymabe:
> Folding this into #795 since I'm testing it all together.

you mean #796?

PR #796, opened 2022-12-22
jobs/build: rerun `build-arch` if previous build is incomplete

jlebon:
Currently, if the x86_64 build succeeds, but not e.g. the aarch64 build,
a rerun of the `build` job will no-op. Salvaging the build requires
rerunning just the `build-arch` job for aarch64, and then rerunning the
`release` job. This is relatively easy for a human, but isn't very
automation-friendly.

With this patch, if we no-op but the latest build is missing builds for
some of the architectures, we automatically rerun the `build-arch` job
for those missing arches and then rerun the `release` job. This allows
the interface for automation to solely be the `build` job, and allows
transparently salvaging up-to-date but incomplete builds rather than
paying for a much costlier full rebuild on all arches.

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -49,6 +49,9 @@ properties([
       booleanParam(name: 'NO_UPLOAD',
                    defaultValue: false,
                    description: 'Do not upload results to S3; for debugging purposes.'),
+      booleanParam(name: 'WAIT_FOR_RELEASE_JOB',
+                   defaultValue: false,
+                   description: 'Wait for the release job and propagate errors.'),
     ] + pipeutils.add_hotfix_parameters_if_supported()),
     buildDiscarder(logRotator(
         numToKeepStr: '100',
@@ -491,7 +494,7 @@ def run_release_job(buildID) {
         // Since we are only running this stage for non-production (i.e.
         // mechanical and development) builds we'll default to allowing failures
         // for additional architectures.
-        build job: 'release', wait: wait, parameters: [
+        build job: 'release', wait: params.WAIT_FOR_RELEASE_JOB, parameters: [
             string(name: 'STREAM', value: params.STREAM),
             string(name: 'ADDITIONAL_ARCHES', value: params.ADDITIONAL_ARCHES),
             string(name: 'VERSION', value: buildID),
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -109,8 +109,20 @@ currentBuild.description = "${build_description} Waiting"
 // declare these early so we can use them in `finally` block
 def newBuildID, basearch
 
+// matches between build/build-arch job
+def timeout_mins = 240
+
+if (params.WAIT_FOR_RELEASE_JOB) {
+    // Waiting for the release job effectively means waiting for all the build-
+    // arch jobs we trigger to finish. While we do overlap in execution (by
+    // a lot when EARLY_ARCH_JOBS is set), let's just simplify and add its
+    // timeout value to ours to account for this. Add 30 minutes more for the
+    // release job itself.
+    timeout_mins += timeout_mins + 30
+}
+
 lock(resource: "build-${params.STREAM}") {
-    timeout(time: 240, unit: 'MINUTES') {
+    timeout(time: timeout_mins, unit: 'MINUTES') {
     cosaPod(cpu: "${ncpus}",
             memory: "${cosa_memory_request_mb}Mi",
             image: cosa_img,
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -1,4 +1,5 @@
 import org.yaml.snakeyaml.Yaml;
+import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 
 node {
     checkout scm
@@ -518,6 +519,14 @@ def run_multiarch_jobs(arches, src_commit, version, cosa_img) {
                 string(name: 'PIPECFG_HOTFIX_REPO', value: params.PIPECFG_HOTFIX_REPO),
                 string(name: 'PIPECFG_HOTFIX_REF', value: params.PIPECFG_HOTFIX_REF)
             ]
+            // Wait until the locks taken by the `build-arch` jobs are taken
+            // before continuing. This closes a potential race in which once we
+            // trigger the `release` job afterwards, it could end up taking the
+            // locks before the multi-arch jobs.
+            // This really should never take more than 5 minutes. Having a
+            // timeout ensures we don't wait for a long time if we somehow
+            // missed the transition.
+            wait_until_locked_or_continue("release-${version}-${arch}", 5)
         }]}
     }
 }
@@ -536,5 +545,26 @@ def run_release_job(buildID) {
             string(name: 'PIPECFG_HOTFIX_REPO', value: params.PIPECFG_HOTFIX_REPO),
             string(name: 'PIPECFG_HOTFIX_REF', value: params.PIPECFG_HOTFIX_REF)
         ]
+    }
+}
+
+// XXX: generalize and put in coreos-ci-lib eventually
+def wait_until_locked_or_continue(resource, timeout_mins) {
+    try {
+        timeout(time: timeout_mins, unit: 'MINUTES') {
+            waitUntil {
+                lock(resource: resource, skipIfLocked: true) {
+                    return false
+                }
+                return true
+            }
+        }
+    } catch (FlowInterruptedException e) {
+        // If the lock was still not taken, then something went wrong. For
+        // example, the job might've failed during the initial `git clone`. The
+        // timeout is to ensure we don't wait forever and here we continue to
+        // try to at least release for the arches that did succeed. We may be
+        // able to salvage the failed arch in the next run.
+        echo "Timed out waiting for lock ${resource} to be taken. Continuing..."
     }
 }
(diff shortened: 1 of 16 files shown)


jlebon:
Requires: https://github.com/coreos/fedora-coreos-pipeline/pull/795

jlebon:
I need to retest this.

jobs/build.Jenkinsfile
```diff
@@ -274,6 +274,24 @@ lock(resource: "build-${params.STREAM}") {
         if (prevBuildID == buildID) {
             currentBuild.result = 'SUCCESS'
             currentBuild.description = "${build_description} 💤 (no new build)"
+
+            def builds = readJSON file: "builds/builds.json"
```
dustymabe, line 278: should we do a sanity check here to confirm 100% that builds.builds[0] matches the build ID we expect it to?

jobs/build.Jenkinsfile
```diff
@@ -243,10 +246,52 @@ lock(resource: "build-${params.STREAM}") {
             }
         }
 
+        // Define closure for forking off multi-arch builds.
+        def fork_mArch_jobs = { arches, commit, version, wait ->
```
dustymabe, line 250: i'm not as happy about the wait here in the case where `WAIT_FOR_RELEASE_JOB` isn't set. Wish there was a better way to make sure the release job didn't get the lock before the build-arch jobs.

dustymabe: Maybe we could do something like:

```
while true:
    lock(resource: 'some_resource', skipIfLocked: true) {
      continue
    }
    break
}
```

dustymabe: we could put that behind some kind of `waitForLockToBeTaken()` function so it's not as ugly

jlebon: > Maybe we could do something like:

I'll note this isn't foolproof, but will give it a try and we'll see how it looks.

jobs/build.Jenkinsfile
```diff
@@ -274,6 +274,24 @@ lock(resource: "build-${params.STREAM}") {
         if (prevBuildID == buildID) {
             currentBuild.result = 'SUCCESS'
             currentBuild.description = "${build_description} 💤 (no new build)"
+
+            def builds = readJSON file: "builds/builds.json"
+            def missing_arches = additional_arches - builds.builds[0].arches
```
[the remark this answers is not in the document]
jlebon: We're subtracting from the set, so `x86_64` in `builds.builds[0].arches` is harmless.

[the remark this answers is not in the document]
dustymabe: Ahh. I had it flipped in my head. 👍

jlebon:
OK updated this now! I ended up reworking things a bit to use separate functions rather than closures since they're easier to handle. Also implemented the "lock probing" discussed above.

jobs/build.Jenkinsfile
```diff
@@ -552,3 +557,22 @@ def run_release_job(buildID) {
...
+                lock(resource: resource, skipIfLocked: true) {
+                    return false
+                }
+                return true
+            }
+        }
+    } catch (FlowInterruptedException e) {
+        // If the lock was still not taken, then something went wrong. We
+        // continue to try to at least release for the arches that did succeed.
+        // We may be able to salvage the failed arch in the next run.
```
dustymabe, line 575: This comment is leaking context in here about why it's OK for the caller to continue, which shouldn't matter inside this function (i.e. what if this function was called by more than one caller.

i.e. this comment should be something more like:

```
// The lock wasn't taken in the allotted time, but the caller
// wants us to continue anyway.

File: jobs/build.Jenkinsfile [hunk 5]
```diff
@@ -532,6 +533,10 @@ def run_multiarch_jobs(arches, src_commit, version, cosa_img) {
                 string(name: 'PIPECFG_HOTFIX_REPO', value: params.PIPECFG_HOTFIX_REPO),
                 string(name: 'PIPECFG_HOTFIX_REF', value: params.PIPECFG_HOTFIX_REF)
             ]
+            // This really should never take more than 5 minutes. Having a
+            // timeout ensures we don't wait for a long time if we somehow
+            // missed the transition.
```
dustymabe, line 538: This should probably have just a little bit more of the context from the commit message in it.

jobs/build.Jenkinsfile
```diff
@@ -552,3 +557,22 @@ def run_release_job(buildID) {
         ]
     }
 }
+
+// XXX: generalize and put in coreos-ci-lib eventually
+def wait_until_locked_or_continue(resource, timeout_mins) {
+    try {
+        timeout(time: timeout_mins, unit: 'MINUTES') {
+            waitUntil {
```
dustymabe, line 565: had to look this up.. it's nice that this has a builtin backoff

jobs/build.Jenkinsfile
```diff
@@ -552,3 +557,22 @@ def run_release_job(buildID) {
         ]
     }
 }
+
+// XXX: generalize and put in coreos-ci-lib eventually
+def wait_until_locked_or_continue(resource, timeout_mins) {
```
dustymabe, line 562: maybe `wait_until_locked_or_timeout`?

jobs/build.Jenkinsfile
```diff
@@ -516,3 +457,62 @@ lock(resource: "build-${params.STREAM}") {
...
+            ]
+        }
+    }
+}
+
+def run_release_job(buildID) {
+    stage('Publish') {
+        // Since we are only running this stage for non-production (i.e. mechanical
+        // and development) builds we'll default to not doing AWS AMI replication.
+        // We'll also default to allowing failures for additional architectures.
```
dustymabe, line 548: We should update this comment. Suggestion:

```
// Since we are only running this stage for non-production (i.e.
// mechanical and development) builds we'll default to allowing
// failures for additional architectures.
```

dustymabe:
FYI: needs a rebase on top of latest `main` branch.

jobs/build.Jenkinsfile
dustymabe: or `wait_until_locked_with_timeout`.. I'm not crazy about any of these. They're just suggestions.

jobs/build.Jenkinsfile
```diff
@@ -109,8 +109,20 @@ currentBuild.description = "${build_description} Waiting"
...
+// matches between build/build-arch job
+def timeout_mins = 240
+
+if (params.WAIT_FOR_RELEASE_JOB) {
+    // Waiting for the release job effectively means waiting for all the build-
+    // arch jobs we trigger to finish. While we do overlap in execution (by
+    // a lot when EARLY_ARCH_JOBS is set), let's just simplify and add its
+    // timeout value to ours to account for this. Add 30 minutes more for the
+    // release job itself.
+    timeout_mins += timeout_mins + 30
```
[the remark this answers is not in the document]
jlebon: Definitely feels like it. The lack of a timeout on the release job I think is an oversight. Was thinking of tackling this in a separate PR for now. WDYT?

jobs/build.Jenkinsfile
jlebon: This is why there's a comment above the helper calling to generalize it and moving it to coreos-ci-lib. :) Just didn't want to block on that since I think we should gain experience with it first and see if there are any issues.

jobs/build.Jenkinsfile
jlebon: Though `timeout` makes it sound like it will error out. I like that `continue` communicates that it's non-fatal.

jobs/build.Jenkinsfile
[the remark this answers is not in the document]
dustymabe: Yeah that works if you want to do it later.

jobs/build.Jenkinsfile
jlebon: Added some. Is that better?

jlebon:
> FYI: needs a rebase on top of latest `main` branch.

Done!

jobs/build.Jenkinsfile
dustymabe: :) - not really. You added more comments to the `wait_until_locked_or_continue` function rather than adding the context here.

What I'm looking for right here is more along the lines of "we need to wait here until the lock is taken because if we don't then we introduce a race between the mArch job and the release job and we don't want the release job to win that race."

Another thing.. Maybe we just move this function call up into the only place it's needed, which is in the `params.EARLY_ARCH_JOBS` section. It's not really needed in the non `params.EARLY_ARCH_JOBS` case, but I'm OK with leaving it here to keep the code more generic.

dustymabe: > Another thing.. Maybe we just move this function call up into the only place it's needed, which is in the `params.EARLY_ARCH_JOBS` section. It's not really needed in the non `params.EARLY_ARCH_JOBS` case, but I'm OK with leaving it here to keep the code more generic.

I guess the problem with that is we have to loop over the arches again. Wouldn't be terrible, but not great either.

jlebon: Yeah, I'd rather just leave it there. I like that it solves the theoretical case too. (And I think it's not that theoretical if you're only building the QEMU image. Compressing just that image takes a few minutes and uploading it just a few seconds; if the `git clone` is being slow for whatever reason in the `build-arch` job, I can imagine it getting overtaken by the release job.)

dustymabe: new comment LGTM

dustymabe (approved):
LGTM

PR #800, opened 2023-01-18
jobs/build-arch: only fetch generated lockfile if autolocking

jlebon:
Otherwise there's no point in fetching it. This also fixes the fact that
the `cosa buildfetch --file` switch hasn't been backported to older cosa
branches, where we don't support autolocking anyway.


dustymabe (approved):
LGTM

PR #802, opened 2023-01-18
jobs/build[-arch]: fix `cosa buildupload` argument ordering

jlebon:
The `--arch` switch is a top-level switch, not an `s3` subcommand
switch.

Fixes e70fb4b ("jobs/build[-arch]: only buildupload for given arch").


dustymabe (approved):
LGTM

marmijo (approved):

PR #804, opened 2023-01-23
bump-lockfile: Update base-url to get builds and artifacts from s3 bucket

gursewak1997:
The initial step to moving RHCOS to use lockfiles.

### DIFF
--- a/jobs/bump-lockfile.Jenkinsfile
+++ b/jobs/bump-lockfile.Jenkinsfile
@@ -8,9 +8,6 @@ node {
 repo = "coreos/fedora-coreos-config"
 botCreds = "github-coreosbot-token-username-password"
 
-// Base URL through which to download artifacts
-BUILDS_BASE_HTTP_URL = "https://builds.coreos.fedoraproject.org/prod/streams"
-
 properties([
     // we're only triggered by bump-lockfiles
     pipelineTriggers([]),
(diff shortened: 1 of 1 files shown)


jobs/bump-lockfile.Jenkinsfile
```diff
@@ -132,15 +131,15 @@ lock(resource: "bump-${params.STREAM}") {
                     if (arch == "x86_64") {
                         shwrap("""
                         cosa buildfetch --arch=${arch} --find-build-for-arch \
-                            --url=${BUILDS_BASE_HTTP_URL}/${branch}/builds
+                            --url=s3://${s3_stream_dir}/builds
```
dustymabe, line 134: now that we're accessing s3 directly we'll need credentials in our environment. We'd need something like: 

https://github.com/coreos/fedora-coreos-pipeline/blob/eb6806cc69472cd52c69bf0aa3e89b765766c998/jobs/kola-aws.Jenkinsfile#L59-L60

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -128,22 +127,25 @@ lock(resource: "bump-${params.STREAM}") {
...
-                        shwrap("""
-                        cosa buildfetch --arch=${arch} --find-build-for-arch \
-                            --url=${BUILDS_BASE_HTTP_URL}/${branch}/builds
-                        cosa fetch --update-lockfile --dry-run
-                        """)
-                    } else {
-                        pipeutils.withExistingCosaRemoteSession(
-                            arch: arch, session: archinfo[arch]['session']) {
+                withCredentials([file(variable: 'AWS_CONFIG_FILE',
+                              credentialsId: 'aws-build-upload-config')]) {
```
dustymabe, line 131: for reasons that aren't obvious I think this needs to go inside the `parallelruns[arch] = {` scope below. 

OR...

even better - let's move it to directly under the `stage("Fetch Metadata") {` scope above. 

basically it needs to not go out of scope before the function gets called (the `parallel parallelruns`)

dustymabe (approved):
LGTM

PR #809, opened 2023-02-07
jobs: include hotfix name in Slack notifications

jlebon:
Otherwise, the Slack notifications look just like the regular ones for the official streams the hotfix build is based on. We noticed this when testing the hotfix workflow recently.


dustymabe (approved):
trivial LGTM

PR #810, opened 2023-02-08
RFC: config: add `streams[].skip_cloud_images`

jlebon:
In RHCOS, we're often in a state where we don't care about regenerating all the disk images and instead just want the containers. We could do this by manually adding all the images we don't care about as `skip_artifacts` entries, but that's cumbersome and error-prone.

Add a `skip_cloud_images` knob for this. We still generate the QEMU and live artifacts for kola tests to run and because they're useful to debug, but we drop everything else.

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -355,6 +355,13 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
         artifacts -= pipecfg.streams[stream].skip_artifacts?.all ?: []
         artifacts -= pipecfg.streams[stream].skip_artifacts?."${basearch}" ?: []
     }
+    if (pipecfg.streams[stream].skip_cloud_images) {
+        // Only keep containers and live artifacts. Note that the ostree
+        // container and QEMU image are always built and not skippable
+        // artifacts.
+        artifacts = artifacts.intersect(["extensions-container", "legacy-oscontainer",
+                                         "live", "metal", "metal4k"])
+    }
     return artifacts.unique()
 }
 
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -64,6 +64,9 @@ streams:
       type: production
       # OPTIONAL: override cosa image to use for this stream
       cosa_image: "quay.io/jlebon/coreos-assembler:stable"
+      # OPTIONAL: Whether to skip cloud image generation and upload. This is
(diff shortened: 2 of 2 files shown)


cgwalters (approved):
Awesome, yes.

Though I think this isn't just RHCOS; as I've said in the past if we e.g. added a continuous stream that rebuilt on every git push, we wouldn't also want to blast images to all the public clouds for every build either.

dustymabe (approved):
I wish we had better terminology than `cloud` here because it includes skipping things that are more hypervisor and not cloud, but honestly I can't think of anything better than `skip-non-core-artifacts: true`.

I would like to maybe expand this in the future to not just `skip_cloud_images: true`, but maybe something more like "build cloud images once every two weeks", in which case we'd be updating the git repo less often because they would still be getting built periodically.
OR maybe we just make this a parameter that can be passed at build time.

utils.groovy
```diff
@@ -355,6 +355,13 @@ def get_artifacts_to_build(pipecfg, stream, basearch) {
         artifacts -= pipecfg.streams[stream].skip_artifacts?.all ?: []
         artifacts -= pipecfg.streams[stream].skip_artifacts?."${basearch}" ?: []
     }
+    if (pipecfg.streams[stream].skip_cloud_images) {
+        // Only keep containers and live artifacts. Note that the ostree
+        // container and QEMU image are always built and not skippable
+        // artifacts.
+        artifacts = artifacts.intersect(["extensions-container", "legacy-oscontainer",
+                                         "live", "metal", "metal4k"])
```
dustymabe, line 363: should we include `hashlist-experimental` here?

jlebon:
@mike-nguyen brought up that if we do this, it breaks the current flow of retroactively selecting cloud images from an already released z-stream when preparing a bootimage bump and replicating that. So either we'd need to change that process so that we turn it back on in time for the z-stream release when we know bootimage bumps are coming, or e.g. we add a new job (or enhance `cloud-replicate`) to "complete the build" with all the cloud images. The complexity there doesn't seem worth it.

jlebon:
I've opened https://github.com/coreos/fedora-coreos-pipeline/pull/1088 closely related to this but with a different rationale. Let's close this one for now.

PR #812, opened 2023-02-13
build: Fix kolatestiso workaround check

ravanelli:
- Update kolatestiso workaround to reflect changes done in coreos-assembler


dustymabe:
hmm. This change isn't necessary is it? i.e. in https://github.com/coreos/coreos-ci-lib/pull/143 we could just continue to put the contents in the `tmp/kolaTestIso-XXXXX` directory?

Is there a good reason to make that change? If not then maybe we keep it the same.

ravanelli:
> hmm. This change isn't necessary is it? i.e. in [coreos/coreos-ci-lib#143](https://github.com/coreos/coreos-ci-lib/pull/143) we could just continue to put the contents in the `tmp/kolaTestIso-XXXXX` directory?
> 
> Is there a good reason to make that change? If not then maybe we keep it the same.

Yes, we won't have this path anymore, and we can't change in the in the ci-lib, since it is part of the build config here.
`tmp/kolaTestIso-*/kola-testiso-uefi/insecure/`
We wil have kola- (or if we want to keep kolaTestIso) + test names

dustymabe:
Let's try to just drop this code: https://github.com/coreos/fedora-coreos-config/pull/2230

Then you can modify this commit to just remove the workaround detection/warning.

PR #815, opened 2023-02-16
Add hyperv to default artifacts

baude:
Supports:

  * https://github.com/coreos/fedora-coreos-tracker/issues/1411
  * https://github.com/coreos/fedora-coreos-tracker/issues/1421

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -68,6 +68,7 @@ default_artifacts:
     - digitalocean
     - exoscale
     - gcp
+    - hyperv
     - ibmcloud
     - nutanix
     - virtualbox


dustymabe (approved):
This change LGTM, but I'll let @jlebon and @bgilbert merge when all the other pieces are in place and we're ready for it.

jlebon:
Discussed this with @baude a bit today. Once Ignition support is in, if we feel like we want to spend more time testing it before making it official, we could skip building it on the official streams for now.

dustymabe:
This was done in https://github.com/coreos/fedora-coreos-pipeline/pull/891 sorry I forgot this change was already proposed in this PR.

PR #817, opened 2023-02-20
streams: enable branched stream

aaradhak:
As F38 has branched from rawhide, we are building the branched stream again

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -21,8 +21,8 @@ streams:
       # type: development # do not touch; line managed by `next-devel/manage.py`
     rawhide:
       type: mechanical
-    # branched:
-    #  type: mechanical
+   branched:
+      type: mechanical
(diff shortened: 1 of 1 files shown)


config.yaml
```diff
@@ -21,12 +21,12 @@ streams:
...
-    # bodhi-updates:
-    #   type: mechanical
-    # bodhi-updates-testing:
-    #   type: mechanical
+    branched:
+      type: mechanical
+    bodhi-updates:
+      type: mechanical
+    bodhi-updates-testing:
+      type: mechanical
```
dustymabe, line 29: Leave these commented out. We're only going to enable `branched`, not `bodhi-updates` and `bodhi-updates-testing`.

dustymabe (approved):
LGTM

PR #819, opened 2023-02-23
config: skip hashlist generation on rawhide/branched

dustymabe:
A newer systemd has introduced a file that is causing the command to fail. Let's skip it on rawhide/branched while we investigate. See: https://github.com/coreos/fedora-coreos-tracker/issues/1429


bgilbert (approved):
LGTM

PR #820, opened 2023-02-23
jobs/build: wait when re-running mArch jobs

dustymabe:
In the case that we are re-running a mArch job I'd prefer we "wait" so that if the mArch job fails we don't re-run the release job.


jlebon (approved):
WFM, thanks! I was thinking about whether we wanted to rerun the release job if some of the missing arches did succeed. But from a releng POV, you really just want the complete build; a slightly more complete but still incomplete build isn't worth the Quay.io churn.

PR #821, opened 2023-02-23
jobs/build: stop waiting for multi-arch jobs to take lock

jlebon:
The main reason we added that was because in the new "rerun build-arch and release jobs" path, there was a higher likelihood that the release job could in theory take the locks before the build-arch jobs. But with 0664cd6 ("jobs/build: wait when re-running mArch jobs"), this is no longer a concern.

There's still the theoretical possibility the race happens even in the regular path (especially when `EARLY_ARCH_JOBS` is unset), but (1) something must be really slow in the multi-arch jobs for that to happen (in which case, it might end up taking more than our 5 minute timeout anyway) and (2) the worst case is that we release without that arch before it's built, which is salvageable (by rerunning the release job).

So overall, IMO maintaining this code is not worth the complexity. We can always bring it back and adjust the timeout if this is a recurring issue.

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -1,5 +1,4 @@
 import org.yaml.snakeyaml.Yaml;
-import org.jenkinsci.plugins.workflow.steps.FlowInterruptedException;
 
 node {
     checkout scm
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

It will be nice to have this code in the future if we ever do want to wait for locks to be taken before continuing, though.

PR #822, opened 2023-02-23
jobs/build: don't try to complete ongoing builds

jlebon:
Since 973bcf9 ("jobs/build: rerun `build-arch` if previous build is
incomplete"), there is a race possible where the `build` job rerun logic
could kick in before the `release` jobs initially triggered for that
build has finished. We don't want to queue builds in that case.

Add an arch-independent lock to the `release` job, and then gate the
rerun logic on whether it's taken. Note this implicitly also covers
waiting for previous `build-arch` jobs.

It's theoretically possible but highly unlikely that we probe the lock
status before the previous `release` job takes it. Ideally, we would
have a way to directly take the lock and "transfer" its ownership to the
job we trigger. Anyway, if that somehow happens, it would result in the
job being run twice, which is safe.


dustymabe:
alternative in https://github.com/coreos/fedora-coreos-pipeline/pull/823

jobs/build.Jenkinsfile
```diff
@@ -265,9 +265,10 @@ lock(resource: "build-${params.STREAM}") {
             // Nothing changed since the latest build. Check if it's missing
             // some arches and retrigger `build-arch` only for the missing
             // arches, and the follow-up `release` job. Match the exact src
-            // config commit that was used. Skip if not uploading since it's
-            // required for multi-arch.
-            if (uploading) {
+            // config commit that was used. But only do this if there isn't
+            // already a release job pending for that build ID. Skip if not
+            // uploading since it's required for multi-arch.
+            if (uploading && !is_locked("release-${buildID}")) {
```
dustymabe, line 271: I think this doesn't really give us what you want because of 0664cd6. i.e. the release job could be not running yet because we're waiting on the mArch jobs. I guess we could check all of those locks here.

jlebon: Good catch!

PR #825, opened 2023-02-27
manifests/jenkins: bump jnlp memory to 512Mi

jlebon:
By default, the jnlp container has a memory request of 256Mi and no
memory limit. In 9eed927 ("manifests/jenkins: set resource limits and
requests for jnlp container"), we added an explicit memory limit so that
we don't inherit the default 10G memory limit from the limitrange in
the RHCOS OpenShift project. We set it to 256Mi since that's what the
default memory *request* was, but it's too little as a hard limit. Most
of the time, the jnlp unexplicably dies, but at least once, I saw it
fail with:

```
java.lang.OutOfMemoryError: unable to create native thread: possibly out of memory or process/resource limits reached
```

Bump it to 512Mi. And also bump the request since in this project we try
to have requests match what the pod actually needs.

While we're here, add a comment about why we set these values.

Fixes https://github.com/coreos/coreos-assembler/issues/3372.


jlebon:
I've already rolled this out in CoreOS CI to test it. If https://github.com/coreos/coreos-assembler/pull/3375 and https://github.com/coreos/coreos-assembler/pull/3374 come back happy, let's consider it fixed and merge this.

dustymabe (approved):
LGTM

PR #827, opened 2023-03-03
coreos-s390x-rhcos-builder: Fix problems with create-secex-data.sh

jschintag:
Add a delay in `create-secex-data.sh` to allow the disk to be fully ready before trying to mount it.
Change the script location to `/usr/local/bin`

Signed-off-by: Jan Schintag <jan.schintag@de.ibm.com>

### DIFF
--- a/multi-arch-builders/coreos-s390x-rhcos-builder.bu
+++ b/multi-arch-builders/coreos-s390x-rhcos-builder.bu
@@ -24,7 +24,8 @@ systemd:
         Description=Create secex-data volume
         [Service]
         Type=oneshot
-        ExecStart=/home/core/create-secex-data.sh
+        RemainAfterExit=yes
+        ExecStart=/usr/local/bin/create-secex-data.sh
         [Install]
         WantedBy=multi-user.target
 storage:
@@ -35,12 +36,12 @@ storage:
       group:
         name: builder
   files:
-    - path: /home/core/create-secex-data.sh
-      mode: 0744
+    - path: /usr/local/bin/create-secex-data.sh
+      mode: 0755
       user:
-        name: core
+        name: root
       group:
-        name: core
+        name: root
       contents:
         inline: |
             #!/bin/bash
@@ -62,6 +63,20 @@ storage:
             if ! $(lszdev | grep -q "${DISK_LUN}"); then
                 echo "Adding LUN to system"
                 echo "${DISK_LUN}" > /sys/bus/ccw/drivers/zfcp/${DISK_FCP}/${DISK_WWWN}/unit_add
+                echo "Waiting for disk to show up"
+                i=0
+                while [ ${i} -lt 5 ]; do
+                  sleep 5
+                  if [ -e "${DISK_PART}" ]; then
+                    echo "Disk is available"
+                    break
+                  fi
+                  ((i++))
+                done
+                if [ ${i} -ge 5 ]; then
+                  echo "Disk failed to show up"
+                  exit 1
+                fi
             fi
 
             if ! $(mountpoint -q "${MNTP}"); then


jschintag:
One Problem right now is that systemd is unable to execute the `create-secex-data.sh` script.
```
[core@coreos-s390x-builder ~]$ sudo systemctl status secex-data-volume.service
× secex-data-volume.service - Create secex-data volume
     Loaded: loaded (/etc/systemd/system/secex-data-volume.service; disabled; preset: enabled)
     Active: failed (Result: exit-code) since Fri 2023-03-03 09:26:41 UTC; 12min ago
   Main PID: 1124 (code=exited, status=203/EXEC)
        CPU: 763us

Mar 03 09:26:41 coreos-s390x-builder systemd[1]: Starting secex-data-volume.service - Create secex-data volume...
Mar 03 09:26:41 coreos-s390x-builder systemd[1124]: secex-data-volume.service: Failed to locate executable /usr/local/bin/create-secex-data.sh: Permission denied
Mar 03 09:26:41 coreos-s390x-builder systemd[1124]: secex-data-volume.service: Failed at step EXEC spawning /usr/local/bin/create-secex-data.sh: Permission denied
Mar 03 09:26:41 coreos-s390x-builder systemd[1]: secex-data-volume.service: Main process exited, code=exited, status=203/EXEC
Mar 03 09:26:41 coreos-s390x-builder systemd[1]: secex-data-volume.service: Failed with result 'exit-code'.
Mar 03 09:26:41 coreos-s390x-builder systemd[1]: Failed to start secex-data-volume.service - Create secex-data volume.

[core@coreos-s390x-builder ~]$ ll /usr/local/bin/create-secex-data.sh
-rwxr-xr-x. 1 root root 1439 Mar  3 09:21 /usr/local/bin/create-secex-data.sh
```
I do not think the service is running before the filesystem is fully mounted and the permissions should be correct.

multi-arch-builders/coreos-s390x-rhcos-builder.bu
```diff
@@ -62,6 +62,20 @@ storage:
             if ! $(lszdev | grep -q "${DISK_LUN}"); then
                 echo "Adding LUN to system"
                 echo "${DISK_LUN}" > /sys/bus/ccw/drivers/zfcp/${DISK_FCP}/${DISK_WWWN}/unit_add
+                echo "Waiting for disk to show up"
```
dustymabe, line 65: I feel like we should be able to use systemd unit primitives to wait for this device to exist before running the `secex-data-volume.service` service

multi-arch-builders/coreos-s390x-rhcos-builder.bu
```diff
@@ -24,7 +24,7 @@ systemd:
         Description=Create secex-data volume
         [Service]
         Type=oneshot
-        ExecStart=/home/core/create-secex-data.sh
+        ExecStart=/usr/local/bin/create-secex-data.sh
```
dustymabe, line 27: maybe we should add `RemainAfterExit=yes` here since this is a oneshot unit.

dustymabe:
> I do not think the service is running before the filesystem is fully mounted and the permissions should be correct.

Hmm. I would think the filesystems would be mounted up before the switchroot into the real root.

multi-arch-builders/coreos-s390x-rhcos-builder.bu
jschintag: We could split the service into 2 parts, the first one adding the device, the 2nd one mounting it. However I don't really think that makes sense.
The main issue with this is that the LPAR does not scan for "new" LUNs automatically, since I do not want to install FCOS to the wrong disk.

jschintag:
> Hmm. I would think the filesystems would be mounted up before the switchroot into the real root.

I think so too, which is my current problem. According to the error, it either can't find the file or does not have sufficient permission to execute it.
But the file should be present and with `755` it should also have sufficient permissions to run it...

multi-arch-builders/coreos-s390x-rhcos-builder.bu
jschintag: Well it does change the state of the system, so it should be the best practice. I will update the PR.

multi-arch-builders/coreos-s390x-rhcos-builder.bu
```diff
@@ -35,12 +36,12 @@ storage:
       group:
         name: builder
   files:
-    - path: /home/core/create-secex-data.sh
-      mode: 0744
+    - path: /usr/local/bin/create-secex-data.sh
+      mode: 0755
```
jlebon, line 40: Following up on our call, YAML does support octals and the generated Ignition does have mode 493 for this file, which is correct.

jschintag: Yes, but it seems I found the root cause. Since it was maintenance window, I was running out of time and tried debugging this by rebooting the machine instead of re-installing because it was faster.
Now with more time, I managed to recreate the error with a local QEMU instance. What I found is this:
1. Creating the script with ignition under `/home/core` or `/var/home/core` -> failure
2. Moving the script on an already running system from `/var/home/core` to `/usr/local/bin`, updating the service and rebooting -> failure
3. Directly creating the script with ignition under `/usr/local/bin` -> success

So the good news is, this PR should work as is.
The bad news, I don't know if the behavior above is a bug or a feature. Since you are not really supposed to change the system after firstboot manually, it might be a feature, at least number 2.

multi-arch-builders/coreos-s390x-rhcos-builder.bu
dustymabe: Oh I see. The `echo "${DISK_LUN}"` is what causes the device to be added to the system?

dustymabe: Is there a way to do that with a configuration file (static) versus an `echo` at runtime (dynamic)?

dustymabe: Either way I'll resolve this conversation.

jschintag: The official docs for RHEL are: https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/7/html/installation_guide/sect-post-installation-fcp-attached-luns-s390

multi-arch-builders/coreos-s390x-rhcos-builder.bu
jlebon: This is likely an SELinux issue, e.g. systemd prevented from running scripts from the home directory. You can verify this by looking for AVC denials in the journal/audit logs.

jschintag:
Since the reprovisioning of the builder on Friday was successful, this PR is good to go from my side. It reflects the current configuration of the s390x RHCOS builder and it is proven to work.

PR #828, opened 2023-03-03
jobs/build: verify kernel + kernel-rt versions match in rhcos

marmijo:
The old RHCOS pipeline had a mechanism for checking if the kernel and kernel-rt packages were built from the same version and failing if they were not. The check was never migrated over to the new RHCOS pipeline. Add the check to verify that the kernel and kernel-rt are of the same version and stop the pipeline if there is a version mismatch. Also add an optional field in the stream section of `config.yaml`  to enable, disable this check. 

By default, the check will not run unless the optional field `check_kernel_rt_mismatch` is set to true. This will ensure the check will not run against FCOS builds.

See: https://issues.redhat.com/browse/COS-2021

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -231,3 +231,5 @@ clouds:
 misc:
   # OPTIONAL: whether to generate a release index
   generate_release_index: true
+  # OPTIONAL: require kernel + kernel-rt versions to match
+  check_kernel_rt_mismatch_rhcos: true
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -347,6 +347,19 @@ lock(resource: "build-${params.STREAM}") {
         // Build the remaining artifacts
         stage("Build Artifacts") {
             pipeutils.build_artifacts(pipecfg, params.STREAM, basearch)
+
+            // Stop the build if the kernel + kernel-rt versions do not match.
+            // This check runs on x86_64 RHCOS builds only.
+            // NOTE: This approach only checks the legacy extensions and not the new extensions
+            // container. This check can be removed for 9.3+ builds when we drop the legacy
+            // oscontainer as the versions will be matched using `match-base-evr` in `extensions.yaml`.
+            if (pipecfg.misc?.check_kernel_rt_mismatch_rhcos) {
+                echo("Verifying kernel + kernel-rt versions match")
+                def build_meta = [readJSON(file: "builds/latest/${basearch}/commitmeta.json"), readJSON(file: "builds/latest/${basearch}/meta.json")]
+                def kernel_version = build_meta[0]['ostree.linux'].split('.el')[0]
+                def kernel_rt_version = build_meta[1]['extensions']['manifest']['kernel-rt-core'].split('.rt')[0]
+                assert kernel_version == kernel_rt_version : "kernel-rt version: ${kernel_rt_version} does not match kernel version: ${kernel_version}"
+            }
         }
 
         // Run Kola TestISO tests for metal artifacts


jlebon:
As mentioned in the ticket, starting from 9.3 the kernel and kernel-rt packages will be built together and won't really need this. Though we'll need to keep it for the older streams that will remain on el8.

An alternative approach is to make the check part of the `cosa buildextend-extensions` command. A tricky bit though is wiring through the knob to disable the check. Our `buildextend` invocations are templated so it can't easily be a switch. We could use an env var, though that's awkward too since we'd set it on all `buildextend` commands and would also have to survive through `cosa remote-session`.

@marmijo I know we landed on this when discussing alternatives, though I another idea I had is to e.g. `touch ${workdir}/tmp/skip-kernel-rt-version-check`. And then both `buildextend-extension` and `buildextend-extensions-container` (but really the `Dockerfile` code) could check that file to know if to skip the check. WDYT? It's not super clean and it's more work, but it keeps the logic in cosa where it'll benefit local devs and would also allow us to cover `buildextend-extensions-container` more easily. @dustymabe Thoughts?

Maybe let's chat about this on Monday.

docs/config.yaml
```diff
@@ -60,6 +60,8 @@ streams:
       source_config_ref: main
       # OPTIONAL: override OS variant to use for this stream
       variant: rhcos-9.0
+      # OPTIONAL: require kernel + kernel-rt versions to match
+      check_kernel_rt_mismatch_rhcos: true
```
dustymabe, line 64: Would we want this only for specific streams? Would we ever want it to apply globally to all streams?

jobs/build.Jenkinsfile
```diff
@@ -347,6 +347,17 @@ lock(resource: "build-${params.STREAM}") {
...
+            // Stop the build if the kernel + kernel-rt versions do not match
+            if (pipecfg.streams[params.STREAM]?.check_kernel_rt_mismatch_rhcos == true) {
+                echo("Verifying kernel + kernel-rt versions match")
+                build_meta = [readJSON(file: "builds/latest/${basearch}/commitmeta.json"), readJSON(file: "builds/latest/${basearch}/meta.json")]
+                kernel_version = build_meta[0]['ostree.linux'].split('.el')[0]
+                kernel_rt_version = [build_meta[1]['extensions']['manifest']['kernel-rt-core'].split('.rt')[0]
+                assert kernel_version == kernel_rt_version : "kernel-rt version: ${kernel_rt_version} does not match kernel version: ${kernel_version}"
+            } else {
+                echo("Skipping kernel + kernel-rt mismatch check")
+            }
```
dustymabe, line 360:
```suggestion
            }
```

I would probably just skip saying anything if that was the case.

dustymabe:
> As mentioned in the ticket, starting from 9.3 the kernel and kernel-rt packages will be built together and won't really need this. Though we'll need to keep it for the older streams that will remain on el8.

This is a great point, which means optimizing this further has diminishing returns.

> 
> An alternative approach is to make the check part of the `cosa buildextend-extensions` command. A tricky bit though is wiring through the knob to disable the check. Our `buildextend` invocations are templated so it can't easily be a switch. We could use an env var, though that's awkward too since we'd set it on all `buildextend` commands and would also have to survive through `cosa remote-session`.

Yep exactly. Because of the way we build "artifacts" it doesn't make it easy to pass through extra arguments unless we use something like an environment variable, which means we'd have to modify the pipeline anyway to achieve this goal, so I just told him to implement the logic in the pipeline which is the place where we have all of the information readily available to us.

> 
> @marmijo I know we landed on this when discussing alternatives, though I another idea I had is to e.g. `touch ${workdir}/tmp/skip-kernel-rt-version-check`. And then both `buildextend-extension` and `buildextend-extensions-container` (but really the `Dockerfile` code) could check that file to know if to skip the check. WDYT? It's not super clean and it's more work, but it keeps the logic in cosa where it'll benefit local devs and would also allow us to cover `buildextend-extensions-container` more easily. @dustymabe Thoughts?

I think we just keep it simple here. Since we already have to touch the pipeline to decide *IF* to do the check, then let's just put the code to perform the check there too.

> 
> Maybe let's chat about this on Monday.

WFM

jobs/build.Jenkinsfile
```diff
@@ -347,6 +347,17 @@ lock(resource: "build-${params.STREAM}") {
         // Build the remaining artifacts
         stage("Build Artifacts") {
             pipeutils.build_artifacts(pipecfg, params.STREAM, basearch)
+
+            // Stop the build if the kernel + kernel-rt versions do not match
```
dustymabe, line 351: A good bit of context to add to this comment is that this should only apply to x86_64 builds (IIUC).

marmijo: That's correct. I thought about adding that, but it's in the x86_64 file, so I opted to not add the information. I'll update though comment though to be clear about what's going on.

marmijo: It'll also help if we decide to change the way we build x86_64 using `build.jenkinsfile`

jobs/build.Jenkinsfile
marmijo: Works for me. That way we dont see this on every FCOS build.

docs/config.yaml
marmijo: If we applied it to all streams, then we would have to disable the check for all streams just to turn this off for e.g. `4.11` only. I considered both options, but figured this gave us more granular control over the check. WDYT?

marmijo:
thanks @dustymabe @jlebon for the conversation and suggestions. We'll chat about this on Monday.

jobs/build.Jenkinsfile
marmijo: updated

jobs/build.Jenkinsfile
marmijo: Removed with latest push

jlebon:
Had an OOB chat with @marmijo and @dustymabe about this. This approach is only checking the legacy extensions and not the new extensions container image. The likelihood that there is a race between the two when querying repodata seems quite low however. To keep it simple, we'll just keep the check here for now, but note that once we drop the legacy oscontainer, we'll have to tweak this to e.g. run the extensions container to get the kernel-rt version (or extend cosa to write that metadata somewhere). Though if the kernel-rt merge happens before that, then we don't have any additional work since we'll be able to use `match-base-evr` directly in the `extensions.yaml`.

jobs/build.Jenkinsfile
```diff
@@ -347,6 +347,16 @@ lock(resource: "build-${params.STREAM}") {
...
+            // This check runs on x86_64 RHCOS builds only
+            if (pipecfg.streams[params.STREAM]?.check_kernel_rt_mismatch_rhcos == true) {
```
jlebon, line 353:
```suggestion
            if (pipecfg.streams[params.STREAM]?.check_kernel_rt_mismatch_rhcos) {
```

jobs/build.Jenkinsfile
```diff
@@ -347,6 +347,16 @@ lock(resource: "build-${params.STREAM}") {
...
+                echo("Verifying kernel + kernel-rt versions match")
+                build_meta = [readJSON(file: "builds/latest/${basearch}/commitmeta.json"), readJSON(file: "builds/latest/${basearch}/meta.json")]
+                kernel_version = build_meta[0]['ostree.linux'].split('.el')[0]
+                kernel_rt_version = build_meta[1]['extensions']['manifest']['kernel-rt-core'].split('.rt')[0]
```
jlebon, line 357: `def` for all of these

docs/config.yaml
jlebon: I think having it be global would probably be fine. AIUI, the use case for temporarily disabling it is for hotfix builds. In that workflow, ART creates a separate pipecfg with the hotfix details, and can turn off the knob if necessary without affecting the prod streams. We can always make it stream-level in the future if needed.

cgwalters:
https://gitlab.com/redhat/centos-stream/src/kernel/centos-stream-9/-/merge_requests/1743 just landed

docs/config.yaml
marmijo: Alright, I can make that change. Where do we want that to live? Just a general `pipecfg.check_kernel_mismatch_rhcos` or add in another layer such as e.g. FCOS's `pipecfg.misc` layer?

jlebon: Under `misc` SGTM!

marmijo:
> To keep it simple, we'll just keep the check here for now, but note that once we drop the legacy oscontainer, we'll have to tweak this to e.g. run the extensions container to get the kernel-rt version (or extend cosa to write that metadata somewhere). Though if the kernel-rt merge happens before that, then we don't have any additional work since we'll be able to use `match-base-evr` directly in the `extensions.yaml`.

I've updated this PR in response to the reviews. I added a note summarizing this to the code. Could I please get some feedback on that? Thanks!

jlebon (approved):
LGTM. Have you been able to test this?

dustymabe (approved):

dustymabe:
@marmijo feel free to merge if it's ready.

PR #830, opened 2023-03-07
config: re-enable next-devel

aaradhak:
The Fedora Linux 38 beta release is next week so `next-devel` and `next` are moving over to Fedora 38 content. Re-enable `next-devel` to allow it to differ from `testing-devel`.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -17,8 +17,8 @@ streams:
     testing-devel:
       type: development
       default: true
-    # next-devel:         # do not touch; line managed by `next-devel/manage.py`
-      # type: development # do not touch; line managed by `next-devel/manage.py`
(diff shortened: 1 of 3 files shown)


jlebon (approved):
It's in two weeks, but worth the respin. :)

dustymabe:
Thanks! Can we disabled `branched` now?

jlebon:
> Thanks! Can we disabled `branched` now?

SGTM! Staged in https://github.com/coreos/fedora-coreos-pipeline/pull/831.

jlebon:
Sorry, I got confused when looking at the dates in [the schedule](https://fedorapeople.org/groups/schedule/f-38/f-38-key-tasks.html). Your commit message is right that it's next week.

PR #831, opened 2023-03-07
config: disable `branched` stream

jlebon:
`next-devel` is now on Fedora 38 so there's no point in also tracking it in the `branched` stream.


jlebon:
Let's wait until https://github.com/coreos/fedora-coreos-config/pull/2254 is merged before merging this one.

dustymabe (approved):
LGTM

PR #832, opened 2023-03-08
jenkins: Add splunk plugins

ravanelli:
- Add the splunk plugging needed in order to be compliant with ProdSec

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -489,6 +489,29 @@ oc annotate secret/additional-root-ca-cert \
 The description can be customized. All other fields must be
 exactly as is.
 
+### [OPTIONAL] Create splunk secret
+
+The Splunk secret is actually a configuration stored in Bitwarden as
+secret. It is optional and must be used for the internal RHCOS cluster.
+Here you can see an example of the secret:
+
+```
+unclassified:
+  splunkJenkinsInstallation:
+    enabled: true
+    globalPipelineFilter: true
+    host: ***
+    port: 8088
+    splunkAppUrl: ***
+    token: ***
+    useSSL: true
+```
+To create it use:
+```
+oc create secret generic splunk-casc-cfg \
+    --from-file=splunk.yaml
+```
+
 ### Create a Jenkins instance with a persistent volume backing store
 
 ```
--- a/jenkins/controller/plugins.txt
+++ b/jenkins/controller/plugins.txt
@@ -28,6 +28,8 @@ kubernetes-credentials-provider:1.199.v4a_1d1f5d074f
 pipeline-github:2.8-138.d766e30bb08b
 slack:625.va_eeb_b_168ffb_0
 timestamper:1.20
+splunk-devops-extend:1.9.9
+splunk-devops:1.9.9
 
 # The below list are plugins that are also in base-plugins.txt but for
 # some reason or other we need to temporarily freeze or fast-track.
--- a/manifests/jenkins.yaml
+++ b/manifests/jenkins.yaml
@@ -136,8 +136,11 @@ objects:
             name: ${JENKINS_SERVICE_NAME}-data
           # DELTA: mount c-as-c config map
           - name: ${JENKINS_SERVICE_NAME}-casc-cfg
-            mountPath: /var/lib/jenkins/configuration-as-code
+            mountPath: /var/lib/jenkins/configuration-as-code/base
             readOnly: true
+          # DELTA: mount c-as-c secret for Splunk
+          - name: splunk-casc-cfg
+            mountPath: /var/lib/jenkins/configuration-as-code/splunk
         dnsPolicy: ClusterFirst
         restartPolicy: Always
         serviceAccountName: ${JENKINS_SERVICE_NAME}
@@ -149,6 +152,11 @@ objects:
         - name: ${JENKINS_SERVICE_NAME}-casc-cfg
           configMap:
             name: jenkins-casc-cfg
+        # DELTA: add a splunk secret
+        - name: splunk-casc-cfg
+          secret:
+            secretName: splunk-casc-cfg
+            optional: true
     triggers:
     - imageChangeParams:
         automatic: true


jenkins/controller/plugins.txt
```diff
@@ -28,6 +28,9 @@ kubernetes-credentials-provider:1.199.v4a_1d1f5d074f
 pipeline-github:2.8-138.d766e30bb08b
 slack:625.va_eeb_b_168ffb_0
 timestamper:1.20
+splunk-devops
+splunk-devops-extend
```
jlebon, line 32: Can we add version tags to these like the other ones?

jlebon (approved):

jlebon:
Did you want to add the JCasC config in this PR or in a follow-up? I guess we need to figure out how to wire up the secrets too.

dustymabe:
`s/pluggings/pluggins` ?

ravanelli:
> Did you want to add the JCasC config in this PR or in a follow-up? I guess we need to figure out how to wire up the secrets too.

Yes, I'm trying to figure out now. The config itself is really simple:
```
unclassified:
  SPLUNKJenkinsInstallation:
    enabled: true
    globalPipelineFilter: true
    host: "${SPLUNK-HOST}"
    port: 8088
    SPLUNKAppUrl: "${SPLUNK-URL}"
    token: "${SPLUNK-TOKEN}"
    useSSL: true
```
I couldn't find a way yet to tie the secrets without mounting it.

jlebon:
While we're here, `s/pluggins/plugins/`? :)

> I couldn't find a way yet to tie the secrets without mounting it.

Yeah, it seems like it doesn't support passing a credential ID. Maybe simplest is to keep in BitWarden the JCasC dropin containing just the Splunk configuration with the secrets in it. And then we change the Jenkins DeploymentConfig to mount it in.

OK, something like this _should_ work:

```diff
diff --git a/manifests/jenkins.yaml b/manifests/jenkins.yaml
index c1d9fce..46b0682 100644
--- a/manifests/jenkins.yaml
+++ b/manifests/jenkins.yaml
@@ -140,7 +140,11 @@ objects:
             name: ${JENKINS_SERVICE_NAME}-data
           # DELTA: mount c-as-c config map
           - name: ${JENKINS_SERVICE_NAME}-casc-cfg
-            mountPath: /var/lib/jenkins/configuration-as-code
+            mountPath: /var/lib/jenkins/configuration-as-code/base
+            readOnly: true
+          # DELTA: mount c-as-c secret for Splunk
+          - name: splunk-casc-cfg
+            mountPath: /var/lib/jenkins/configuration-as-code/splunk
             readOnly: true
         dnsPolicy: ClusterFirst
         restartPolicy: Always
@@ -153,6 +157,11 @@ objects:
         - name: ${JENKINS_SERVICE_NAME}-casc-cfg
           configMap:
             name: jenkins-casc-cfg
+        # DELTA: add a splunk secret
+        - name: splunk-casc-cfg
+          secret:
+            secretName: splunk-casc-cfg
+            optional: true
     triggers:
     - imageChangeParams:
         automatic: true
```

The base JCasC config is now in the `base/` subdirectory. The JCasC plugin should recurse into all directories when looking for dropins under `/var/lib/jenkins/configuration-as-code/`.

ravanelli:
> Hmm, we shouldn't need to move `jenkins/config` to a `base` subdir in this repo itself. Just having it get mounted at `base` in the YAML file should be enough.

A ok, I probably got it wrong  =). Fixing it

HACKING.md
```diff
@@ -489,6 +489,18 @@ oc annotate secret/additional-root-ca-cert \
 The description can be customized. All other fields must be
 exactly as is.
 
+### [OPTIONAL] Create splunk secret
+
```
jlebon, line 493: For consistency with the other sections, can we add some context here? At least let's mention that the secret is located in BitWarden.

jlebon (approved):
One comment, LGTM otherwise!

manifests/jenkins.yaml
```diff
@@ -136,8 +136,11 @@ objects:
             name: ${JENKINS_SERVICE_NAME}-data
           # DELTA: mount c-as-c config map
           - name: ${JENKINS_SERVICE_NAME}-casc-cfg
-            mountPath: /var/lib/jenkins/configuration-as-code
+            mountPath: /var/lib/jenkins/configuration-as-code/base
```
dustymabe, line 139: is this a harmless change? we're changing the mountpath of something, but I don't see another corresponding change somewhere else to adjust for it (in this same commit).

HACKING.md
dustymabe: It would also be nice to show an example of what that secret looks like (without any secret data in the example, of course).

manifests/jenkins.yaml
jlebon: Yeah, it should be harmless. JCasC recurses into directories. So now it'll look in both the `base` and the `splunk` subdirs.

HACKING.md
```diff
@@ -489,6 +489,29 @@ oc annotate secret/additional-root-ca-cert \
...
+    port: 8088
+    splunkAppUrl: ***
+    token: ***
+    useSSL: true
+```
+To create it use:
+```
+oc create secret generic splunk-casc-cfg \
+    --from-literal=splunk.yaml \
+    --from-file=splunk.yaml
```
jlebon, line 512: This looks off. I don't think we want the `--from-literal` switch? The `--from-file` switch should do what we want (create a key named `splunk.yaml` with the contents of the file).

HACKING.md
```diff
@@ -489,6 +489,29 @@ oc annotate secret/additional-root-ca-cert \
...
+    port: 8088
+    splunkAppUrl: ***
+    token: ***
+    useSSL: true
+```
+To create it use:
+```
+oc create secret generic splunk-casc-cfg \
+    --from-file=splunk.yaml \
+    --from-file=splunk.yaml
```
jlebon, line 512:
```suggestion
    --from-file=splunk.yaml
```

jlebon (approved):
One comment, LGTM otherwise!

HACKING.md
dustymabe: This still needs to be addressed.. `--from-file` is specified twice.

dustymabe:
I feel like we maybe should squash all of these commits into a single commit?

HACKING.md
```diff
@@ -489,6 +489,29 @@ oc annotate secret/additional-root-ca-cert \
 The description can be customized. All other fields must be
 exactly as is.
 
+### [OPTIONAL] Create splunk secret
+
+The Splunk secret is actually a configuration stored in Bitwarden as
+secret, it is optional and must be used for the internal clusters
+when necessary, here you can see an example of the secret:
```
dustymabe, line 496: Optional wording suggestions:

```suggestion
The Splunk secret is actually a configuration stored in Bitwarden as
secret. It is optional and must be used for the internal RHCOS cluster.
Here you can see an example of the secret:
```

dustymabe (approved):
LGTM

dustymabe (approved):
LGTM

jlebon (approved):

PR #836, opened 2023-03-13
jobs: Add Brew upload job

ravanelli:
- Create a separate job to upload the metadata;
- We are not required to upload it for every single build as we did before. Only ART will be responsible for it, we are required to make Brew uploads for GAs and other images available to customers.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -509,6 +509,25 @@ To create it use:
 oc create secret generic splunk-casc-cfg \
     --from-file=splunk.yaml
 ```
+### [OPTIONAL] Create brew secrets (if uploading to Brew)
+
+Brew requires a couple of secrets in order to work. All of them are
+stored in Bitwarden.  We need the certificate (brew-ca), the keytab
+(brew-keytab), the koji configuration (koji-conf) and the kerberos
+configuration (krb5-conf). Make sure to meet the secret names, these
+names are used in the brew job.
+
+To create it use:
+```
+oc create secret generic brew-ca \
+    --from-file=ca.crt
+oc create secret generic brew-keytab \
+    --from-file=brew.keytab
+oc create secret generic koji-conf \
+    --from-file=koji.conf
+oc create secret generic krb5-conf \
+    --from-file=krb5.conf
+```
 
 ### Create a Jenkins instance with a persistent volume backing store
 
--- a/jobs/upload-brew.Jenkinsfile
+++ b/jobs/upload-brew.Jenkinsfile
@@ -1,0 +1,135 @@
+node {
+    checkout scm
+    // these are script global vars
+    pipeutils = load("utils.groovy")
+    pipecfg = pipeutils.load_pipecfg()
+}
+
+def brew_principal = pipecfg.brew.principal
+def brew_profile = pipecfg.brew.profile
+
+properties([
+    pipelineTriggers([]),
+    parameters([
+      string(name: 'ADDITIONAL_ARCHES',
+             description: "Override additional architectures (space-separated). " +
+                          "Use 'none' to only upload for x86_64. " +
+                          "Supported: ${pipeutils.get_supported_additional_arches().join(' ')}",
+             defaultValue: "",
+             trim: true),
+      string(name: 'VERSION',
+             description: 'Build version',
+             defaultValue: '',
+             trim: true),
+      choice(name: 'STREAM',
+             choices: pipeutils.get_streams_choices(pipecfg),
+             description: 'CoreOS stream to upload to brew'),
+      booleanParam(name: 'ALLOW_MISSING_ARCHES',
+                   defaultValue: false,
+                   description: 'Allow upload to continue even with missing architectures'),
+    ]),
+    buildDiscarder(logRotator(
+        numToKeepStr: '100',
+        artifactNumToKeepStr: '100'
+    )),
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+])
+
+def build_description = "[${params.STREAM}]"
+
+// runtime parameter always wins
+def cosa_img = params.COREOS_ASSEMBLER_IMAGE
+cosa_img = cosa_img ?: pipeutils.get_cosa_img(pipecfg, params.STREAM)
+def basearches = []
+if (params.ADDITIONAL_ARCHES != "none") {
+    basearches = params.ADDITIONAL_ARCHES.split() as List
+    basearches = basearches ?: pipeutils.get_additional_arches(pipecfg, params.STREAM)
+}
+
+// we always upload for x86_64
+basearches += 'x86_64'
+// make sure there are no duplicates
+basearches = basearches.unique()
+build_description += "[${basearches.join('')}][${params.VERSION}]"
+
+// We don't need that much memory for downloading/uploading to brew, since
+// it will be mostly metadata
+def cosa_memory_request_mb = 1 * 1024 as Integer
+
+// Same here, we don't need that much
+def ncpus = 1
+
+echo "Waiting for upload-brew lock"
+currentBuild.description = "${build_description} Waiting"
+def stream_info = pipecfg.streams[params.STREAM]
+
+lock(resource: "upload-brew") {
+    cosaPod(cpu: "${ncpus}",
+            memory: "${cosa_memory_request_mb}Mi",
+            image: cosa_img,
+            serviceAccount: "jenkins",
+            secrets: ["brew-keytab", "brew-ca:ca.crt:/etc/pki/ca.crt",
+                      "koji-conf:koji.conf:/etc/koji.conf",
+                      "krb5-conf:krb5.conf:/etc/krb5.conf"]) {
+    timeout(time: 240, unit: 'MINUTES') {
+    try {
+        stage('Fetch Metadata') {
+            def ref = pipeutils.get_source_config_ref_for_stream(pipecfg, params.STREAM)
+            def variant = stream_info.variant ? "--variant ${stream_info.variant}" : ""
+            def s3_stream_dir = pipeutils.get_s3_streams_dir(pipecfg, params.STREAM)
+            pipeutils.shwrapWithAWSBuildUploadCredentials("""
+                cosa init --branch ${ref} ${variant} ${pipecfg.source_config.url}
+                cosa buildfetch --build=${params.VERSION} \
+                    --arch=all --url=s3://${s3_stream_dir}/builds \
+                    --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG} \
+                    --file "coreos-assembler-config-git.json"
+            """)
+        }
+        def builtarches = shwrapCapture("""
+                          cosa shell -- cat builds/builds.json | \
+                              jq -r '.builds | map(select(.id == \"${params.VERSION}\"))[].arches[]'
+                          """).split() as Set
+        assert builtarches.contains("x86_64"): "The x86_64 architecture was not in builtarches."
+        if (!builtarches.containsAll(basearches)) {
+            if (params.ALLOW_MISSING_ARCHES) {
+                warn("Some requested architectures did not successfully build!")
+                basearches = builtarches.intersect(basearches)
+            } else {
+                echo "ERROR: Some requested architectures did not successfully build"
+                echo "ERROR: Detected built architectures: $builtarches"
+                echo "ERROR: Requested base architectures: $basearches"
+                currentBuild.result = 'FAILURE'
+                return
+            }
+        }
+
+        stage('Brew Upload') {
+            def tag = pipecfg.streams[params.STREAM].brew.tag
+            for (arch in basearches) {
+                shwrap("""
+                    coreos-assembler koji-upload \
+                        upload --reserve-id \
+                        --keytab "/run/kubernetes/secrets/brew-keytab/brew.keytab" \
+                        --build ${params.VERSION} \
+                        --retry-attempts 6 \
+                        --buildroot builds \
+                        --owner ${brew_principal} \
+                        --profile ${brew_profile} \
+                        --tag ${tag} \
+                        --arch "$arch"
+                """)
+            }
+        }
+        currentBuild.result = 'SUCCESS'
+
+    } catch (e) {
+        currentBuild.result = 'FAILURE'
+        throw e
+    } finally {
+        def message = "[${params.STREAM}][${basearches.join(' ')}] (${params.VERSION})"
+        echo message
+        if (currentBuild.result != 'SUCCESS') {
+            pipeutils.trySlackSend(message: ":beer: brew-upload <${env.BUILD_URL}|#${env.BUILD_NUMBER}> ${message}")
+        }
+    }
+}}} // timeout, cosaPod, and lock finish here
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -112,6 +112,15 @@ s3:
   # OPTIONAL: object ACL to use when uploading
   acl: private
 
+# OPTIONAL: Brew profile and principal to which to upload build metadata to Brew.
+brew:
+  # REQUIRED: Profile name
+  profile: brew-stage
+  # REQUIRED: Principal name
+  principal: rhcos-build/servername
+  # REQUIRED: tag
+  tag: candidate
+
 # OPTIONAL: container registry-related keys
 registry_repos:
   # OPTIONAL: repo and tags to which to push oscontainer


ravanelli:
+ We can also add an option for all (all arches) in the future.

cgwalters:
> /hold Merging, we also need a PR in coreos-assembler granting access under /etc for koji and brew configuration files. As a PR for config.yaml

Can't we just mount these as ConfigMap/Secret objects?

ravanelli:
> > /hold Merging, we also need a PR in coreos-assembler granting access under /etc for koji and brew configuration files. As a PR for config.yaml
> 
> Can't we just mount these as ConfigMap/Secret objects?

I think your idea for the pipeline is not to mount anything, and use the `withCredentials` pluging for the secrets we need. Basically we will need something similar to https://github.com/coreos/coreos-assembler/blob/main/Dockerfile#L36, to make brew/koji config to work

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,128 @@
...
+    checkout scm
+    pipeutils = load("utils.groovy")
+    pipecfg = pipeutils.load_pipecfg()
+    libcloud = load("libcloud.groovy")
+
+    brew_principle =  pipecfg['brew']['principle']
+    brew_profile =  pipecfg['brew']['profile']
+}
+
+arches = [ 'aarch64', 'ppc64le', 'x86_64', 's390x']
```
dustymabe, line 15: we have a `get_supported_additional_arches()` function for this, but we should probably just default to pushing all arches that the given version was built against.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,128 @@
...
+        numToKeepStr: '100',
+        artifactNumToKeepStr: '100'
+    )),
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+])
+
+def build_description = "[${params.VERSION}]"
+
+// Always use the latest
+cosa_img = "quay.io/coreos-assembler/coreos-assembler:latest"
```
dustymabe, line 42: should probably use the same machanism for this as we do the other jobs: 

https://github.com/coreos/fedora-coreos-pipeline/blob/0c6fef51773fe60d9e18083b2c6159e74ec6e3e0/jobs/build.Jenkinsfile#L75-L77

It is true that we want to use the non stream specific COSA, so maybe we modify it in some way. Either way, should still be picking this up from the pipecfg.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,128 @@
...
+// testiso: https://github.com/coreos/fedora-coreos-tracker/issues/1339
+def cosa_memory_request_mb = 10.5 * 1024 as Integer
+
+// Now that we've established the memory constraint based on xz above, derive
+// kola parallelism from that. We leave 512M for overhead and VMs are at most
+// 1.5G each (in general; root reprovisioning tests require 4G which is partly
+// why we run them separately).
+// XXX: https://github.com/coreos/coreos-assembler/issues/3118 will make this
+// cleaner
+def ncpus = ((cosa_memory_request_mb - 512) / 1536) as Integer
```
dustymabe, line 58: we need to update the comments here since we don't decompress anything OR run `kola`.. I doubt we need as much memory as we do for a full `build` job either.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,128 @@
...
+
+echo "Waiting for build-upload-to-brew lock"
```
dustymabe, line 60:
```suggestion
echo "Waiting for upload-brew lock"
```

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,128 @@
...
+currentBuild.description = "${build_description} Waiting"
+def stream_info = pipecfg.streams[params.STREAM]
+
+lock(resource: "build-upload-to-brew") {
```
dustymabe, line 64:
```suggestion
lock(resource: "upload-brew") {
```

I think we typically try to name these close to the job name.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,128 @@
...
+                file(credentialsId: 'brew-cfg', variable: 'BREW'),
+                     file(credentialsId: 'brew-krb-config', variable: 'KRB5'),
+                     file(credentialsId: 'brew-ca', variable: 'CA'),
+                     file(credentialsId: 'brew-keytab', variable: 'KEYTAB'),
+             ]) {
+                shwrap("""
+                    cp $BREW /etc/koji.conf
+                    cp $KRB5 /etc/krb5.conf
+                    cp $KEYTAB /tmp/keytab
+                    cp $CA /etc/pki/brew/brew-ca
```
dustymabe, line 95: I assume there's no way to override the place to look for these files when invoking `brew`? That might be easier than modifying `cosa`.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,128 @@
...
+        }
+        currentBuild.result = 'SUCCESS'
+
+    // main timeout and try {} finish here
+    } catch (e) {
+             currentBuild.result = 'FAILURE'
+             throw e
+         } finally {
+             def color
+             def message = "[${params.VERSION}][${params.ARCH}]"
```
dustymabe, line 116: probably want the stream name in here too and also the name of the job. we want this message to be similar to the kola jobs.. i.e. something like:

```
:fcos: :gcp: kola-gcp [#394](https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/job/kola-gcp/394/) [rawhide][x86_64] (39.20230313.91.0)
```

jobs/upload-brew.Jenkinsfile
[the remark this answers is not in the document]
dustymabe: Since we'd only run this when we're doing actual RHCOS releases we could probably get by with having it single threaded.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,128 @@
...
+
+arches = [ 'aarch64', 'ppc64le', 'x86_64', 's390x']
+
+properties([
+    pipelineTriggers([]),
+    parameters([
+      choice(name: 'ARCH',
+             description: 'The target architecture',
+             choices: (arches),
+             trim: true),
```
dustymabe, line 23: when we run this do we only want to upload for a given arch, or do we essentially want all arches for that given build ID to be uploaded to brew?

Assuming we want all arches, can we do it in a single brew call or do we need to do separate brew calls (one for each arch)?

If we need to do separate brew calls I would still consider making this job do them all rather than doing it one by one (mentioned this in https://github.com/coreos/fedora-coreos-pipeline/pull/836/files#r1134578302 too).

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,128 @@
...
+        stage('Fetch Metadata') {
+            def ref = pipeutils.get_source_config_ref_for_stream(pipecfg, params.STREAM)
+            def variant = stream_info.variant ? "--variant ${stream_info.variant}" : ""
+            def s3_stream_dir = pipeutils.get_s3_streams_dir(pipecfg, params.STREAM)
+            pipeutils.shwrapWithAWSBuildUploadCredentials("""
+                cosa init --branch ${ref} ${variant} ${pipecfg.source_config.url}
+                cosa buildfetch --build=${params.VERSION} \
+                    --arch=${params.ARCH} --url=s3://${s3_stream_dir}/builds \
+                    --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG} \
+                    --artifact=qemu \
```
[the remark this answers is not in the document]
jlebon: If we need to have one single canonical artifact for metadata purposes, I'd go for the ociarchive instead.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,117 @@
+import org.yaml.snakeyaml.Yaml;
+
+def pipeutils, pipecfg, libcloud, uploading, jenkins_agent_image_tag
```
jlebon, line 3: Let's do the same thing as https://github.com/coreos/fedora-coreos-pipeline/commit/e1fdff84e42cd2b6c91bac176b180ebef9a571a3 here for consistency.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,117 @@
...
+// runtime parameter always wins
+def cosa_img = params.COREOS_ASSEMBLER_IMAGE 
+cosa_img = cosa_img ?: pipeutils.get_cosa_img(pipecfg, params.STREAM) 
+
+// We don't need that much memory for downloading/uploading to brew, since
+// it will be mostly metadata
+def cosa_memory_request_mb = 5 * 1024 as Integer
+
+// Same here, we don't need that much 
+def ncpus = 4
```
jlebon, line 49: This still feels a bit overprovisioned. Wouldn't 1 CPU and 1G of RAM be enough?

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,117 @@
...
+
+lock(resource: "upload-brew") {
+    timeout(time: 240, unit: 'MINUTES') {
+    cosaPod(cpu: "${ncpus}",
+            memory: "${cosa_memory_request_mb}Mi",
+            image: cosa_img,
+            serviceAccount: "jenkins", 
+            secrets: ["brew-keytab", "brew-ca:ca.crt:/etc/pki/ca.crt",
+                      "koji-conf:koji.conf:/etc/koji.conf",
+                      "krb5-conf:krb5.conf:/etc/krb5.conf"]) {
```
jlebon, line 63: Nice!

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,117 @@
...
+            }
+        }
+        currentBuild.result = 'SUCCESS'
+
+    // main timeout and try {} finish here
+    } catch (e) {
+             currentBuild.result = 'FAILURE'
+             throw e
+         } finally {
+             def color
```
jlebon, line 101: We don't need the color stuff anymore (it's handled by `trySlackSend`).

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,117 @@
...
+             def message = "[${params.STREAM}][${params.VERSION}][${params.ARCH}]"
+
+             if (currentBuild.result == 'SUCCESS') {
+                 message = "brew :sparkles: ${message}"
```
jlebon, line 105: Is it even worth sending a notification on success? Maybe let's just send one if something fails.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,117 @@
...
+                 color = 'good';
+             } else {
+                 message = "brew :fire: ${message}"
```
jlebon, line 108: Bikeshed:

```suggestion
                 message = ":beer: :sparkles: ${message}"
```

(Matches the Brew favicon) :)

jobs/upload-brew.Jenkinsfile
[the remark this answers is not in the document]
ravanelli: I think we can work on it as a follow up, it would also requires some change here: https://github.com/coreos/coreos-assembler/blob/main/src/cmd-koji-upload#L585C21-L587

Either way I will need to touch again this rpm creation part to fix the host rpms, we will need to upload it to S3 too.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,115 @@
...
+node {
+    checkout scm
+    // these are script global vars
+    pipeutils = load("utils.groovy")
+    pipecfg = pipeutils.load_pipecfg()
+    libcloud = load("libcloud.groovy")
+
+}
+
+def brew_principle = pipecfg.brew.principle
```
jlebon, line 12: Mentioned this in the internal MR using this too but surfacing it here too: I think this is probably supposed to be `principal` instead?

jlebon: Also, let's document these new knobs in `docs/config.yaml`.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,115 @@
...
+         } finally {
+             def message = "[${params.STREAM}][${params.VERSION}][${params.ARCH}]"
+
+             if (currentBuild.result != 'SUCCESS') {
+                 message = ":beer: :sparkles: ${message}"
+             }
+
+             echo message
+             pipeutils.trySlackSend(message: message)
+         }
```
jlebon, line 112: Indentation is off here.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,115 @@
...
+     }
+   }
```
jlebon, line 114: Likely here it would look something like:

```groovy
}}} // try-catch-finally, cosaPod and timeout finish here
```

like we do in other jobs

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,115 @@
...
+    // main timeout and try {} finish here
+    } catch (e) {
+             currentBuild.result = 'FAILURE'
+             throw e
+         } finally {
+             def message = "[${params.STREAM}][${params.VERSION}][${params.ARCH}]"
+
+             if (currentBuild.result != 'SUCCESS') {
+                 message = ":beer: :sparkles: ${message}"
+             }
```
jlebon, line 108:
```suggestion
             if (currentBuild.result != 'SUCCESS') {
                 pipeutils.trySlackSend(message: ":beer: brew-upload <${env.BUILD_URL}|#${env.BUILD_NUMBER}> ${message}")
             }
```

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,115 @@
...
+                        --owner ${brew_principle} \
+                        --profile ${brew_profile} \
+                        --tag ${params.TAG} \
+                        --arch "$arch"
+                """)
+            }
+        }
+        currentBuild.result = 'SUCCESS'
+
+    // main timeout and try {} finish here
```
jlebon, line 99: The timeout doesn't finish here. For consistency, we can drop this comment and only have the one at the bottom as suggested.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,115 @@
...
+
+lock(resource: "upload-brew") {
+    timeout(time: 240, unit: 'MINUTES') {
+    cosaPod(cpu: "${ncpus}",
+            memory: "${cosa_memory_request_mb}Mi",
+            image: cosa_img,
+            serviceAccount: "jenkins", 
+            secrets: ["brew-keytab", "brew-ca:ca.crt:/etc/pki/ca.crt",
+                      "koji-conf:koji.conf:/etc/koji.conf",
+                      "krb5-conf:krb5.conf:/etc/krb5.conf"]) {
```
jlebon, line 66: Should invert the order of these. See https://github.com/coreos/fedora-coreos-pipeline/pull/836.

HACKING.md
```diff
@@ -509,6 +509,22 @@ To create it use:
 oc create secret generic splunk-casc-cfg \
     --from-file=splunk.yaml
 ```
+### [OPTIONAL] Create brew secrets
+
+Brew requires a couple of secrets in order to work. All of them are stored in Bitwarden.
+We need the certificate (brew-ca), the keytab (brew-keytab), the koji configuration (koji-conf) and the kerberos configuration (krb5-conf). Make sure to meet the secret names, these names are used in the brew job.
```
jlebon, line 515: For consistency, would be good to wrap this to ~72 columns.

HACKING.md
```diff
@@ -509,6 +509,22 @@ To create it use:
 oc create secret generic splunk-casc-cfg \
     --from-file=splunk.yaml
 ```
+### [OPTIONAL] Create brew secrets
```
jlebon, line 512: Maybe something like

```suggestion
### [OPTIONAL] Create Brew secrets (if uploading to Brew)
```

?

HACKING.md
```diff
@@ -509,6 +509,22 @@ To create it use:
...
+To create it use:
+```
+oc create secret generic brew-ca \
+    --from-literal=filename=ca.crt --from-file=data=ca.crt
+oc create secret generic brew-keytab \
+    --from-literal=filename=brew-keytab --from-file=data=brew-keytab
+oc create secret generic koji-conf \
+    --from-literal=filename=koji-conf --from-file=data=koji-conf
+oc create secret generic krb5-conf \
+    --from-literal=filename=krb5-conf --from-file=data=krb5-conf
```
jlebon, line 526: Hmm, these look off. It's creating a secret with keys `filename` and `data`, but I think based on the code below, the key names should be e.g. `ca.crt`, `koji.conf`, `krb5.conf`, no?

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,115 @@
...
+    // these are script global vars
+    pipeutils = load("utils.groovy")
+    pipecfg = pipeutils.load_pipecfg()
+    libcloud = load("libcloud.groovy")
+
+}
+
+def brew_principle = pipecfg.brew.principle
+def brew_profile = pipecfg.brew.profile
+def arches = pipeutils.get_supported_additional_arches()
```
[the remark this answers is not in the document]
ravanelli: We will always have a build with all arches, so I think in this case we can assume use all arches, instead of checking it, WDYT?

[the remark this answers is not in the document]
dustymabe: Ideally we always ship with all arches, but one day there may be a need to exclude one arch when doing a release so it would probably be better to check rather than assume.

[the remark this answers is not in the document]
jlebon: Hotfix builds is an example where we likely wouldn't have all arches (I'm not sure if the Brew job would be used in that case, but it should do the right thing if it were).

jobs/upload-brew.Jenkinsfile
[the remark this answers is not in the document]
dustymabe: Though, as she said before maybe we can leave it as a followup to drop the qemu upload.

HACKING.md
```diff
@@ -509,6 +509,25 @@ To create it use:
...
+To create it use:
+```
+oc create secret generic brew-ca \
+    --from-literal=ca.crt
+oc create secret generic brew-keytab \
+    --from-literal=brew.keytab
+oc create secret generic koji-conf \
+    --from-literal=koji.conf
+oc create secret generic krb5-conf \
+    --from-literal=krb5.conf
```
jlebon, line 529: I think these need to be

```suggestion
oc create secret generic brew-ca \
    --from-file=ca.crt
oc create secret generic brew-keytab \
    --from-file=brew.keytab
oc create secret generic koji-conf \
    --from-file=koji.conf
oc create secret generic krb5-conf \
    --from-file=krb5.conf
```

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,107 @@
...
+
+    } catch (e) {
+             currentBuild.result = 'FAILURE'
+             throw e
+    } finally {
+        def message = "[${params.STREAM}][${params.VERSION}][${params.ARCH}]"
+        echo message
+       // if (currentBuild.result != 'SUCCESS') {
+       //          pipeutils.trySlackSend(message: ":beer: brew-upload <${env.BUILD_URL}|#${env.BUILD_NUMBER}> ${message}")
+       // }
```
jlebon, line 105: Meant to uncomment this?

jobs/upload-brew.Jenkinsfile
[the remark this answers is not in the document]
ravanelli: https://github.com/coreos/coreos-assembler/pull/3488
simpler than I thought ;)

jobs/upload-brew.Jenkinsfile
ravanelli: Yes, opis!

jobs/upload-brew.Jenkinsfile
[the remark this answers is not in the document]
ravanelli: Ok, how about we let all arches as default,  but an option to overwrite it?

ravanelli:
About the qemu artifact see: https://github.com/coreos/coreos-assembler/pull/3488

jobs/upload-brew.Jenkinsfile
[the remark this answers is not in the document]
jlebon: I think if the goal is to eventually make it part of the release job, then we should just match the same semantics as that. I.e. let's have an `ADDITIONAL_ARCHES` and a `ALLOW_MISSING_ARCHES` parameter like in the release job?

jlebon (commented):
Discussed this with @ravanelli OOB. We're going to get this in as a separate job for now. It's not currently being triggered by the build or release jobs, so it's purely manual.

Once we make `cosa koji-upload` idempotent, we can merge it into the release job.

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,136 @@
+node {
+    checkout scm
+    // these are script global vars
+    pipeutils = load("utils.groovy")
+    pipecfg = pipeutils.load_pipecfg()
+    libcloud = load("libcloud.groovy")
+
```
dustymabe, line 7:
```suggestion
```

Unused

jobs/upload-brew.Jenkinsfile
```diff
@@ -0,0 +1,136 @@
...
+                """)
+            }
+        }
+        currentBuild.result = 'SUCCESS'
+
+    } catch (e) {
+        currentBuild.result = 'FAILURE'
+        throw e
+    } finally {
+        def message = "[${params.STREAM}][${params.VERSION}][${params.ARCH}]"
```
dustymabe, line 130:
```suggestion
        def message = "[${params.STREAM}][${basearches.join(' ')}] (${params.VERSION})"
```

Let's make this more like the release job.

dustymabe:
I would possibly just squash the two commits down into one.

dustymabe (approved):
LGTM

PR #837, opened 2023-03-13
jobs: do `cosa buildfetch` with `time -v` when fetching large objects

jlebon:
On `cosa buildfetch` invocations that fetch large objects, add `time -v` so we get a picture of how much memory the operation took. We're occasionally hitting what looks like OOM issues, but before either increasing our memory requests or tweaking S3 download settings, let's sanity-check first that's what's happening.


dustymabe (approved):
LGTM

PR #838, opened 2023-03-14
jobs: nest timeout within `cosaPod`

jlebon:
Getting the resources for the cosa pod might take time if there's lots of concurrent builds in progress filling our quota. That wait time shouldn't count towards the timeout. Flip the scoping order so that the timeout starts only when we actually have the pod.

This is most problematic in the `build` job where we allocate pods with lots of resources. But for consistency, do this in all our jobs.


dustymabe (approved):
LGTM

cgwalters (approved):
Thanks for taking out the time to fix this.

PR #839, opened 2023-03-20
config: support per-stream `skip_cloud_uploads` knob

jlebon:
While working with unreleased content, we don't want any publicly accessible artifacts. This includes cloud images. Add a `skip_cloud_uploads` key for this.

This is temporary to unblock the RHEL 9.2 work. I think what would be nicer here instead is to still upload to clouds but keep the images private. However, we need to add a `--public` knob to some of the cloud commands we use first.

Another option worth investigating is support for overriding the base `cloud` key entirely like we have for some other keys.

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -316,7 +316,7 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
 
         // Upload to relevant clouds
         // XXX: we don't support cloud uploads yet for hotfixes
-        if (uploading && !pipecfg.hotfix) {
+        if (uploading && !pipecfg.hotfix && !stream_info.skip_cloud_uploads) {
             stage('Cloud Upload') {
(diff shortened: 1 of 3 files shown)


cgwalters (approved):

jobs/build-arch.Jenkinsfile
```diff
@@ -316,7 +316,7 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
 
         // Upload to relevant clouds
         // XXX: we don't support cloud uploads yet for hotfixes
-        if (uploading && !pipecfg.hotfix) {
+        if (uploading && !pipecfg.hotfix && !stream_info.skip_cloud_uploads) {
```
jlebon, line 319: I wasn't sure if `!stream_info.skip_cloud_uploads` would work in the case that it's undefined. But we clearly assume it does elsewhere and haven't had any issues, so simplified this!

PR #840, opened 2023-03-20
jobs: pass fully qualified ref name to `git ls-remote`

jlebon:
Otherwise we risk getting multiple refs matching the given pattern (i.e. the branch name showing up in more branches than just the canonical branch). We hit this today with a revert branch that was auto-created accidentally by GitHub through the web UI:

```
$ git ls-remote https://github.com/coreos/fedora-coreos-config next
b9f8903339c142b00105bd0fb97c3fcc21e1502a        refs/heads/next
9007e5804e41097afaa7e42d13455d8a8cd87375        refs/heads/revert-2307-bootloader/next
```

And then the `src_config_commit` variable had multiple lines which messed up parameter substitution lower down.


dustymabe (approved):
LGTM

PR #841, opened 2023-03-22
config: Start producing kubevirt artifact for x86_64

gursewak1997:
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1126#issuecomment-1452628138

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -69,6 +69,7 @@ default_artifacts:
     - exoscale
     - gcp
     - ibmcloud
+    - kubevirt
     - nutanix
     - virtualbox
     - vmware


dustymabe:
LGTM - but let's hold this until after the triple release going out later this week.

dustymabe:
Also, let's get the `stream-metadata-{go,rust}` library updates that we found were needed in place before we merge this.

gursewak1997:
> Also, let's get the `stream-metadata-{go,rust}` library updates that we found were needed in place before we merge this.

Those are already merged in:
- https://github.com/coreos/stream-metadata-rust/pull/39
- https://github.com/coreos/stream-metadata-go/pull/57

dustymabe:
oh nice!

dustymabe:
Though I do wonder if we also need a release to happen in those projects or not cc @jlebon 

Either way this LGTM.

dustymabe (approved):

jlebon:
Yes, stream-metadata-go at least needs to be tagged for Dependabot to bump it in cosa. I'm not sure where stream-metadata-rust is used (I thought in coreos-installer, but that's not the case). Though it looks like the PRs are just changing test fixtures. Since the schema itself hasn't changed, we don't need to get releases out or block this PR.

Issue #803, opened 2023-01-19
create a job to aid in multi-arch builder access


dustymabe:
We recently discussed our limited access to multi-arch for most of our team members. We have our multi-arch builders, but managing access can be complicated. We could just use the pipeline itself to spawn a remote session on a builder and the user can then grab a shell via OpenShift to do testing. The remote sessions would be auto-garbage collected after a period of time.

aaradhak:
@dustymabe 
Summarizing the requirements as discussed in the meeting. Kindly review if there are any missing points in the requirements.

Requirements:
- pod name includes the name of the person who ran the job
       - i.e. user-devpod-2023101abc
- pod/job times out after X hours (configurable?)
- starts remote session on multi-arch builder and does a few convenience things
       - runs tmux (so that the session can be connected to by the user)
           - user uses openshift webui or CLI to connect to pod and re-attach to tmux
           - tmux window offers two panes
                  - one where the remote session is already set up. 
                  - running cosa just works in tmux. 
- Run cosa init

aaradhak:
Completed the task of adding debug pod job .

PR #843, opened 2023-03-24
jobs:debug-pod to aid in multi-arch builder access

aaradhak:
In order to resolve the limited access to multi-arch for most of our team members,we have the multi-arch builders where we could just use the pipeline itself to spawn a remote session on a builder and the user can then grab a shell via OpenShift to do testing. cosa-init & sleep stages are added as part of creating a job for multi-arch builder access.

Ref: https://github.com/coreos/fedora-coreos-pipeline/issues/803


dustymabe:
closing in favor of #844

PR #844, opened 2023-03-27
jobs:debug-pod for multi-arch builder access

aaradhak:
In order to resolve the limited access to multi-arch for most of our team members,we have the multi-arch builders where we could just use the pipeline itself to spawn a remote session on a builder and the user can then grab a shell via OpenShift to do testing. cosa-init & sleep stages are added as part of creating a job for multi-arch builder access.

Ref: https://github.com/coreos/fedora-coreos-pipeline/issues/803

### DIFF
--- a/jobs/debug-pod.Jenkinsfile
+++ b/jobs/debug-pod.Jenkinsfile
@@ -1,0 +1,135 @@
+import org.yaml.snakeyaml.Yaml;
+
+node {
+    checkout scm
+    // these are script global vars
+    pipeutils = load("utils.groovy")
+    pipecfg = pipeutils.load_pipecfg()
+}
+
+properties([
+    pipelineTriggers([]),
+    parameters([
+      choice(name: 'STREAM',
+             choices: pipeutils.get_streams_choices(pipecfg),
+             description: 'CoreOS stream to build'),
+      choice(name: 'ARCH',
+             description: 'The target architecture',
+             choices: pipeutils.get_supported_additional_arches()),
+      string(name: 'COREOS_ASSEMBLER_IMAGE',
+             description: 'Override coreos-assembler image to use',
+             defaultValue: "",
+             trim: true),
+      string(name: 'TIMEOUT',
+             description: 'Timeout value (in hours)',
+             defaultValue: "8",
+             trim: true),  
+    ]),
+    buildDiscarder(logRotator(
+        numToKeepStr: '15',
+        artifactNumToKeepStr: '15'
+    )),
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+])
+
+// Retrieve the username logged in from Jenkins 
+def userName = currentBuild.getBuildCauses('hudson.model.Cause$UserIdCause')[0]?.userName
+def build_description = "[${userName}][${params.STREAM}][${params.ARCH}]"
+
+// runtime parameter always wins
+def cosa_img = params.COREOS_ASSEMBLER_IMAGE
+cosa_img = cosa_img ?: pipeutils.get_cosa_img(pipecfg, params.STREAM)
+
+def stream_info = pipecfg.streams[params.STREAM]
+
+// Note that the heavy lifting is done on a remote node via podman
+// --remote so we shouldn't need much memory or CPU.
+def cosa_memory_request_mb = 512
+def ncpus = 1
+
+// Give ourselves a unique pod name that the user can identify
+def podName = "debug-pod-$userName-${params.ARCH}-"
+podName += UUID.randomUUID().toString().substring(0, 8)
+
+cosaPod(cpu: "${ncpus}",
+        memory: "${cosa_memory_request_mb}Mi",
+        image: cosa_img,
+        serviceAccount: "jenkins" ,
+        name : podName) {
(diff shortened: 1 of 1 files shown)


jlebon:
I presume this replaces https://github.com/coreos/fedora-coreos-pipeline/pull/843 ? Can we close that one then?

jobs/debug-pod.Jenkinsfile
```diff
@@ -0,0 +1,187 @@
...
+// release job would win this race.
+lock(resource: "release-${params.VERSION}-${basearch}") {
+// build lock: we don't want multiple concurrent builds for the same stream and
+// arch (though this should work fine in theory)
+lock(resource: "build-${params.STREAM}-${basearch}") {
+    cosaPod(cpu: "${ncpus}",
+            memory: "${cosa_memory_request_mb}Mi",
+            image: cosa_controller_img,
+            serviceAccount: "jenkins" ,
+            user : "${userName}") {
```
dustymabe, line 98: for this argument to `cosaPod` I would just name the argument `name`, as in the name of the pod. i.e. your coreos-ci-lib PR would be to add a `name` argument to `pod.groovy`.

jobs/debug-pod.Jenkinsfile
```diff
@@ -0,0 +1,187 @@
...
+
+// runtime parameter always wins
+def cosa_img = params.COREOS_ASSEMBLER_IMAGE
+cosa_img = cosa_img ?: pipeutils.get_cosa_img(pipecfg, params.STREAM)
+
+def stream_info = pipecfg.streams[params.STREAM]
+
+def cosa_controller_img = stream_info.cosa_controller_img_hack ?: cosa_img
+cosa_controller_img = 'quay.io/dustymabe/coreos-assembler-staging:cosa-tmux'
+// DELETE ^^
```
dustymabe, line 56: would be nice if we could go ahead and get the pieces in COSA that we need and stop using this image I built.

jobs/debug-pod.Jenkinsfile
aaradhak: I have renamed the "user" to be "name"

jobs/debug-pod.Jenkinsfile
aaradhak: Just to be clear, we have added tmux in [coreos-assembler](https://github.com/coreos/coreos-assembler/blob/main/src/build-deps.txt), in that case does it need to be pointed to the latest image or do we delete this?
`cosa_controller_img = http://quay.io/coreos-assembler/coreos-assembler:latest`

dustymabe: oh good. Forgot we did that in https://github.com/coreos/coreos-assembler/pull/3404

then just delete line 55 and 56.

honestly lets just delete line 54 too and change line 97 to use `cosa_img`.

jobs/debug-pod.Jenkinsfile
```diff
@@ -0,0 +1,188 @@
...
+// declare these early so we can use them in `finally` block
+assert params.VERSION != ""
+def newBuildID = params.VERSION
+def basearch = params.ARCH
+
+// matches between build/build-arch job
+def timeout_mins = 240
+
+// Retrieve the username logged in from Jenkins 
+def userName = currentBuild.getBuildCauses('hudson.model.Cause$UserIdCause')[0]?.userName ?: 'default'
```
dustymabe, line 82: is there a case here where the username won't be retrieved (i.e. why do we need the `default` fallback?)

jobs/debug-pod.Jenkinsfile
aaradhak: done

jobs/debug-pod.Jenkinsfile
aaradhak: hmm right, jira would have a username always. we can remove the 'default'

jobs/debug-pod.Jenkinsfile
```diff
@@ -0,0 +1,139 @@
...
+node {
+    checkout scm
+    // these are script global vars
+    pipeutils = load("utils.groovy")
+    pipecfg = pipeutils.load_pipecfg()
+    libcloud = load("libcloud.groovy")
+}
+
+// Base URL through which to download artifacts
+BUILDS_BASE_HTTP_URL = "https://builds.coreos.fedoraproject.org/prod/streams"
```
dustymabe, line 12: delete unused.

jobs/debug-pod.Jenkinsfile
```diff
@@ -0,0 +1,139 @@
+import org.yaml.snakeyaml.Yaml;
+
+node {
+    checkout scm
+    // these are script global vars
+    pipeutils = load("utils.groovy")
+    pipecfg = pipeutils.load_pipecfg()
+    libcloud = load("libcloud.groovy")
```
dustymabe, line 8:
```suggestion
```

delete - unused

dustymabe (changes requested):

dustymabe (approved):
LGTM - thanks for working on this!

aaradhak:
> LGTM - thanks for working on this!

Thank you, Dusty!

PR #846, opened 2023-03-29
Revert "config: Start producing kubevirt artifact for x86_64"

dustymabe:
Reverts coreos/fedora-coreos-pipeline#841

Apparently `cosa buildextend-kubevirt` works fine when running locally
but runs into failures in the pipeline. Let's revert for now until
we get fixes into coreos-assembler so that it can run fine inside
of OpenShift.


gursewak1997 (approved):
Lgtm

PR #847, opened 2023-03-29
jobs: add kola-upgrade test

dustymabe:
This will test upgrades from older releases to verify old systems can be updated to the latest versions. For our production FCOS streams we'll kick off a full matrix to try to test from different points in the history of the stream. For development/mechanical streams we'll just kick of a single modulo $date test per build.

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -379,6 +379,12 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
                 pipeutils.run_cloud_tests(pipecfg, params.STREAM, newBuildID,
                                           s3_stream_dir, basearch, src_config_commit)
             }
+            if (pipecfg.misc?.run_extended_upgrade_test_fcos) {
+                stage('Upgrade Tests') {
+                    pipeutils.run_fcos_upgrade_tests(pipecfg, params.STREAM,
+                                                     newBuildID, basearch, src_config_commit)
+                }
+            }
         }
 
         stage('Destroy Remote') {
--- a/jobs/kola-upgrade.Jenkinsfile
+++ b/jobs/kola-upgrade.Jenkinsfile
@@ -1,0 +1,244 @@
+node {
+    checkout scm
+    // these are script global vars
+    pipeutils = load("utils.groovy")
+    pipecfg = pipeutils.load_pipecfg()
+}
+
+properties([
+    pipelineTriggers([]),
+    parameters([
+      choice(name: 'STREAM',
+             choices: pipeutils.get_streams_choices(pipecfg),
+             description: 'CoreOS stream to test'),
+      string(name: 'START_VERSION',
+             description: 'CoreOS Build ID or Fedora major version to start from',
+             trim: true),
+      string(name: 'TARGET_VERSION',
+             description: 'Final CoreOS Build ID that passes test',
+             defaultValue: '',
+             trim: true),
+      string(name: 'ARCH',
+             description: 'Target architecture',
+             defaultValue: 'x86_64',
+             trim: true),
+      string(name: 'COREOS_ASSEMBLER_IMAGE',
+             description: 'Override the coreos-assembler image to use',
+             defaultValue: "quay.io/coreos-assembler/coreos-assembler:main",
+             trim: true),
+      string(name: 'SRC_CONFIG_COMMIT',
+             description: 'The exact config repo git commit to run tests against',
+             defaultValue: '',
+             trim: true),
+    ]),
+    buildDiscarder(logRotator(
+        numToKeepStr: '100',
+        artifactNumToKeepStr: '100'
+    )),
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+])
+
+
+// Map the FCOS stream name to the production stream we should
+// use for the test.
+def start_streams = [
+    'stable':         'stable',
+    'testing':        'testing',
+    'next':           'next',
+    'testing-devel':  'testing',
+    'next-devel':     'next',
+    'branched':       'next',
+    'rawhide':        'next',
+]
+
+// Deadend releases as found in
+// https://github.com/coreos/fedora-coreos-streams/tree/main/updates
+def deadends = ['38.20230310.1.0','30.20190716.1']
+
+// Vars for start/target versions
+def target_version = params.TARGET_VERSION
+def start_version = params.START_VERSION
+def start_stream = start_streams[params.STREAM]
+
+// Because of an x86_64 bootloader issue the oldest that is supported
+// on x86_64 is 32.x: https://github.com/coreos/fedora-coreos-tracker/issues/1448
+if (start_version == '' && params.ARCH == 'x86_64') {
+    start_version = '32'
+}
+
+currentBuild.description = "[${params.STREAM}][${params.ARCH}] - ${start_version}->${target_version}"
+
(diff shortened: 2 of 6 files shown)


dustymabe:
requires https://github.com/coreos/fedora-coreos-config/pull/2335 to merge first

jobs/build-arch.Jenkinsfile
```diff
@@ -379,6 +379,10 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
                 pipeutils.run_cloud_tests(pipecfg, params.STREAM, newBuildID,
                                           s3_stream_dir, basearch, src_config_commit)
             }
+            stage('Upgrade Tests') {
+                pipeutils.run_upgrade_tests(pipecfg, params.STREAM,
+                                            newBuildID, basearch, src_config_commit)
+            }
```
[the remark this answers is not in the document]
dustymabe: ok I added a knob for this.

utils.groovy
```diff
@@ -482,6 +496,65 @@ def run_cloud_tests(pipecfg, stream, version, s3_stream_dir, basearch, commit) {
...
+    }
+
+    // If this is a non-production build we'll run a single test from
+    // a start version determined by a modulo of the day of the month.
+    // This prevents running the full matrix for each development
+    // build, but gradually covers the full matrix over a period of time.
+    if (stream_info.type in ['development', 'mechanical']) {
+        def now = new Date()
+        def dayOfMonth = now.getAt(Calendar.DAY_OF_MONTH)
+        def offset = dayOfMonth % (min_version - max_version)
```
jlebon, line 538: Shouldn't this be

```suggestion
        def offset = dayOfMonth % (max_version - min_version)
```

?

(Or maybe it doesn't matter in Groovy? But better to flip it anyway so we don't have to wonder.)

Assuming we want `max_version` to be one of the possible choices, I think also there's an off-by-one bug here and so this would be

```suggestion
        def offset = dayOfMonth % (max_version - min_version + 1)
```

jlebon, line 538: Using days for this means that the same starting point will be used for multiple builds in the same day. Also if the pipeline is broken for a day, we won't get coverage for that starting point until the next cycle.

WDYT instead of something like:


```suggestion
        def builds = readJSON(file: "builds/builds.json")
        def offset = builds.size % (max_version - min_version + 1)
```

?

jobs/kola-upgrade.Jenkinsfile
```diff
@@ -0,0 +1,242 @@
...
+        numToKeepStr: '100',
+        artifactNumToKeepStr: '100'
+    )),
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+])
+
+
+// runtime parameter always wins
+def cosa_img = params.COREOS_ASSEMBLER_IMAGE
+cosa_img = cosa_img ?: pipeutils.get_cosa_img(pipecfg, params.STREAM)
```
jlebon, line 44: We don't typically use `get_cosa_img` for kola tests. For consistency, let's not do it here either? It doesn't work anyway as is since the parameter has a default value.

jobs/kola-upgrade.Jenkinsfile
```diff
@@ -0,0 +1,242 @@
...
+}
+
+properties([
+    pipelineTriggers([]),
+    parameters([
+      choice(name: 'STREAM',
+             choices: pipeutils.get_streams_choices(pipecfg),
+             description: 'CoreOS stream to test'),
+      string(name: 'START_VERSION',
+             description: 'CoreOS Build ID to start from',
```
jlebon, line 15: Should this include something like ` (or Fedora major version)`?

jobs/kola-upgrade.Jenkinsfile
```diff
@@ -0,0 +1,242 @@
...
+                    k2.extraArgs += " --qemu-firmware=bios"
+                    parallelruns['Kola:BIOS'] = { kola(k2) }
+                    break;
+                case 'aarch64':
+                    k1 = kolaparams.clone()
+                    k1.extraArgs += " --qemu-firmware=uefi"
+                    parallelruns['Kola:UEFI'] = { kola(k1) }
+                    break;
+                case 's390x':
+                    parallelruns['Kola:S390FW'] = { kola(kolaparams) }
```
jlebon, line 209: Minor: personally not a fan of the S390FW and PPC64FW naming. We used it in the testiso case because we had to come up with something for easier pattern matching. But here, just "Kola" seems good enough?

jobs/kola-upgrade.Jenkinsfile
```diff
@@ -0,0 +1,242 @@
...
+            // If we're on mArch and using QEMU then initialize the
+            // session on the remote builder
+            stage("Initialize Remote") {
+                // Wrap a bunch of commands now inside the context of a remote
+                // session. All `cosa` commands, other than `cosa remote-session`
+                // commands, should get intercepted and executed on the remote.
+                // We set environment variables that describe our remote host
+                // that `podman --remote` will transparently pick up and use.
+                // We set the session to time out after 4h. This essentially
+                // performs garbage collection on the remote if we fail to clean up.
```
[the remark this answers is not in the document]
dustymabe: just deleted it. It's probably not needed.

utils.groovy
```diff
@@ -482,6 +496,65 @@ def run_cloud_tests(pipecfg, stream, version, s3_stream_dir, basearch, commit) {
...
+    def start_versions = []
+
+    // If this is a production build we'll run a test from the
+    // earliest release of each Fedora major we've supported for this
+    // stream. Iterate over the possibilities and kick off a job for
+    // each.
+    if (stream_info.type == 'production') {
+        for (start_version in min_version..max_version) {
+            start_versions += start_version
+        }
```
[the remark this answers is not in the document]
dustymabe: yeah. I was thinking a bit more about this and I think it would be good to go ahead and run the full matrix for `testing` and `next` for the same reason as we want to run tests against `rawhide`. i.e. if there were a breaking change we'd want to know it when we were getting ready to release `testing` rather than later when we promote that content to stable.

I'm not super firm on this, but since we only run new builds on average every two weeks, it seemed like a little additional insurance for not a lot of drawback.

dustymabe:
Ok I think I addressed most comments.

dustymabe:
pushed a `testing` commit here while I test a few things further.

utils.groovy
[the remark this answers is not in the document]
jlebon: SGTM, we can always adjust in the future too.

dustymabe:
ok this should be ready to go now.

jobs/kola-upgrade.Jenkinsfile
```diff
@@ -0,0 +1,244 @@
...
+            ]
+            def k1, k2, k3
+
+            switch(params.ARCH) {
+                case 'x86_64':
+                    k1 = kolaparams.clone()
+                    k1.extraArgs += " --qemu-firmware=uefi"
+                    k1.marker = "uefi"
+                    parallelruns['Kola:UEFI'] = { kola(k1) }
+                    if ((start_version[0..1] as Integer) < 34) {
```
jlebon, line 187: Shouldn't this be

```suggestion
                    if ((start_version[0..1] as Integer) >= 34) {
```

?

dustymabe:
Not sure I understand the CI failure.

jlebon (approved):

PR #848, opened 2023-04-01
jobs/kola-upgrade: fix deadend checks; disallow uefi-secure on aarch64

dustymabe:
```
commit a6b6946e5623a128512d307ef9abc4fa220e9182
Date:   Fri Mar 31 20:35:52 2023 -0400

    jobs/kola-upgrade: fix deadend checks
    
    In would case we should have used `start_version` and in the other case
    we should have used `release["version"]` and not `release`.

commit 3fede5fc8a9a2f5e5bf208ca103a49e5c603ce8c
Date:   Fri Mar 31 20:29:50 2023 -0400

    jobs/kola-upgrade: SecureBoot on aarch64 isn't supported
    
    Looks like this isn't supported in Fedora. When running the tests
    we ran into:
    
    ```
     2023-03-31T23:49:51Z kola: retryloop: failed to bring up machines:
     architecture aarch64 doesn't have support for secure boot in kola
    ```
    
    I think this migth be the blocker for Fedora right now:
    https://pagure.io/fedora-infrastructure/issue/7361
```

### DIFF
--- a/jobs/kola-upgrade.Jenkinsfile
+++ b/jobs/kola-upgrade.Jenkinsfile
@@ -102,13 +102,21 @@ lock(resource: "kola-upgrade-${params.ARCH}") {
         } else {
             shwrap("curl -LO https://builds.coreos.fedoraproject.org/prod/streams/${start_stream}/releases.json")
             def releases = readJSON file: "releases.json"
+            def newest_version = releases["releases"][-1]["version"]
             for (release in releases["releases"]) {
                 def has_arch = release["commits"].find{ commit -> commit["architecture"] == params.ARCH }
                 if (release["version"] in deadends || has_arch == null) {
                     continue // This release has been disqualified
                 }
                 if (start_version.length() == 2) {
-                    if ((release["version"][0..1] as Integer) > (start_version as Integer)) {
+                    if (release["version"] == newest_version) {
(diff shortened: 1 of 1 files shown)


dustymabe:
added another commit:

```
commit 1c304aaafc5714be205ec059f452a17204aa9758
Date:   Fri Mar 31 21:08:00 2023 -0400

    jobs/kola-upgrade: set target_version based on target stream
    
    Since we can have a different start stream verses target stream
    here we need to pick up the target version from the target stream's
    builds.json, not the start stream's builds.json.

```

dustymabe:
added another commit:

```
commit ab2996a3119598695b23a534bb73e3a9a0a35cb2
Date:   Sat Apr 1 14:13:59 2023 -0400

    jobs/kola-upgrade: use latest in start_stream if necessary
    
    We hit this today in the pipeline. The builds.size() % max-min+1
    code specified 39 as the start_version (rawhide is currently on 39)
    but the `next` stream is on 38. So the releases.json traversal code
    wasn't finding and builds in the next stream start were on 39 and
    the test ultimately failed.
    
    If we encounter a case like this, let's just use the latest build
    from the start_stream.

```

jlebon (approved):
LGTM. The force-push there is just for some typo fixes in the commit messages and code.

Issue #833, opened 2023-03-13
Add Secure Execution builder host key to enable reboots in tests


jlebon:
From https://github.com/coreos/coreos-assembler/pull/3384#discussion_r1124933887:

> Hmm, in a follow-up, we could do something similar on the s390x builder we do for the genprotimg qcow2 where we put the builder's hostkey in a volume and then mount it into the cosa container at e.g. `/data.secex-hostkey` or something. Then, our test would even be able to reboot and test upgrades and verify that the system comes up fine.

jlebon:
Bundling it in the existing volume SGTM. :+1:

jschintag:
I have added the hostkey of the LPAR into the volume:
```
$ tree /data.secex/
/data.secex/
├── genprotimgvm.qcow2
└── hostkeys
    └── HKD-<serial>.crt

2 directories, 2 files
```
So for the future the syntax would be `hostkeys/HKD-*.crt` for adding new Hostkeys.

jlebon:
Thanks! Let's close this then.

PR #849, opened 2023-04-03
jobs: add kola-secex test

nikita-dubrovskii:
This will test qemu-secex artifact on s390x bulder machine.

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -306,6 +306,14 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
             pipeutils.build_artifacts(pipecfg, params.STREAM, basearch)
         }
 
+        // secex specific tests
+        if (shwrapCapture("cosa meta --get-value images.qemu-secex") != "None") {
+            stage("Kola:Secex") {
+                kola(cosaDir: env.WORKSPACE, arch: basearch, skipUpgrade: true,
+                     extraArgs: "--qemu-secex --tag secex --qemu-secex-hostkey /data.secex/hostkeys/HKD-*.crt")
+            }
+        }
+
         // Run Kola TestISO tests for metal artifacts
         if (shwrapCapture("cosa meta --get-value images.live-iso") != "None") {
             stage("Kola:TestISO") {


jobs/build-arch.Jenkinsfile
```diff
@@ -306,6 +306,15 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
             pipeutils.build_artifacts(pipecfg, params.STREAM, basearch)
         }
 
+        // secex specific tests
+        if (shwrapCapture("cosa meta --get-value images.qemu-secex") != "None") {
+            stage("Kola:TestSecex") {
+                // FIXME: Once we have valid HKD for the builder we should add `--qemu-secex-hostkey /path/to/key` switch and enable reboot tests
```
jschintag, line 312: I already added the certificate to the builder:
https://github.com/coreos/fedora-coreos-pipeline/issues/833#issuecomment-1494220283
I think we should implement in the test some option to import all certificates from `/data.secex/hostkeys/HKD-*.crt` by default and only take `--qemu-secex-hostkey /path/to/key` as an override.

jobs/build.Jenkinsfile
```diff
@@ -362,6 +362,15 @@ lock(resource: "build-${params.STREAM}") {
...
 
+        // secex specific tests
+        if (shwrapCapture("cosa meta --get-value images.qemu-secex") != "None") {
+            stage("Kola:TestSecex") {
+                // FIXME: Once we have valid HKD for the builder we should add `--qemu-secex-hostkey /path/to/key` switch and enable reboot tests
+                kola(cosaDir: env.WORKSPACE, arch: basearch, skipBasicScenarios: true, skipSecureBoot: true, skipUpgrade: true,
+                     extraArgs: "--qemu-secex --tag secex")
+            }
+        }
+
```
jschintag, line 373: This should not be necessary, as `build` only runs `x86`

jobs/build-arch.Jenkinsfile
```diff
@@ -306,6 +306,15 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
...
+                kola(cosaDir: env.WORKSPACE, arch: basearch, skipBasicScenarios: true, skipSecureBoot: true, skipUpgrade: true,
+                     extraArgs: "--qemu-secex --tag secex")
```
jschintag, line 314: I'm not that into the implementation of the kola function, but would it not make more sense to implement a setting here like `runOnlySecex: true`, to specify that we only want secex tests to run? This way, with all the skips seems more prone to error when future tests are added.

jobs/build.Jenkinsfile
nikita-dubrovskii: Thx, will drop

jobs/build-arch.Jenkinsfile
nikita-dubrovskii: good to know, so now we can mount them `-v /data.secex/hostkeys/key:/secex-hostkey`

jobs/build-arch.Jenkinsfile
nikita-dubrovskii: maybe, but than i have to update https://github.com/coreos/coreos-ci-lib/blob/main/vars/kola.groovy

nikita-dubrovskii: @jlebon wdyt?

jobs/build-arch.Jenkinsfile
jschintag: It is already mounted in the CI by default, since it is a volume.

nikita-dubrovskii: done

jobs/build-arch.Jenkinsfile
```diff
@@ -306,6 +306,14 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
             pipeutils.build_artifacts(pipecfg, params.STREAM, basearch)
         }
 
+        // secex specific tests
+        if (shwrapCapture("cosa meta --get-value images.qemu-secex") != "None") {
+            stage("Kola:TestSecex") {
```
jlebon, line 311: Minor: maybe just `Kola:Secex`? "Test" is implied by kola. "Kola:TestISO" below is a reference to the command name.

jobs/build-arch.Jenkinsfile
```diff
@@ -306,6 +306,14 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
...
+                kola(cosaDir: env.WORKSPACE, arch: basearch, skipBasicScenarios: true, skipSecureBoot: true, skipUpgrade: true,
```
jlebon, line 312: Passing `extraArgs` already implies that we won't run basic scenarios, so we can simplify this to

```suggestion
                kola(cosaDir: env.WORKSPACE, arch: basearch, skipUpgrade: true,
```

jobs/build-arch.Jenkinsfile
jlebon: Probably not worth having a helper knob specific to secex. I could imagine in the future though have a `tag` parameter to make this more ergonomic. Anyway, in my review below I mention we can simplify this line today.

jlebon (approved):
LGTM, though to be safe let's hold it until [the current releases](https://github.com/coreos/fedora-coreos-streams/issues?q=is%3Aissue+2023-04-03) are out.

PR #853, opened 2023-04-13
jobs/kola-upgrade: add exception for next stream start_version

dustymabe:
Early versions of `next` on F32 did not yet have Ignition 2.3.0 with spec 3.1.0 support. Let's add an exception that will hardcode the starting version for next if the specified start_version is 32.

### DIFF
--- a/jobs/kola-upgrade.Jenkinsfile
+++ b/jobs/kola-upgrade.Jenkinsfile
@@ -91,6 +91,13 @@ lock(resource: "kola-upgrade-${params.ARCH}") {
             """)
         }
         echo "Selected ${target_version} as the target version to test"
+        // EXCEPTION: If the prod stream we'll use for the upgrade test will be `next`
+        // then we need to start at 33 because the earliest 32 versions in `next`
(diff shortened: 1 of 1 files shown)


dustymabe:
> Hmm, seems like it'd be cleaner to lower the artificial merge config to use spec 3.0.0 to avoid a special case here. But cool with this too!

I did the initial spec lowering in https://github.com/coreos/coreos-assembler/pull/3416. It didn't seem obvious to me how to change the code in COSA to use spec 3.0.0 because Resources was added in 3.1.0. If you see a clear path to victory we can drop this workaround.

jlebon:
> > Hmm, seems like it'd be cleaner to lower the artificial merge config to use spec 3.0.0 to avoid a special case here. But cool with this too!
> 
> I did the initial spec lowering in [coreos/coreos-assembler#3416](https://github.com/coreos/coreos-assembler/pull/3416). It didn't seem obvious to me how to change the code in COSA to use spec 3.0.0 because Resources was added in 3.1.0. If you see a clear path to victory we can drop this workaround.

Threw up https://github.com/coreos/coreos-assembler/pull/3426!

PR #854, opened 2023-04-13
jobs/kola-gcp: rename `gce`-related switches to `gcp`

jlebon:
cosa was recently updated to use `gcp` rather than `gce`: https://github.com/coreos/coreos-assembler/pull/3424


HuijingHei:
As the dependent PR https://github.com/coreos/coreos-assembler/pull/3424 is merged, so remove the draft status.

HuijingHei (approved):
LGTM

PR #855, opened 2023-04-13
config: enable aarch64 boot images for GCP

r4f4:
Google cloud now has support for aarch64 instances https://cloud.google.com/compute/docs/instances/arm-on-compute

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -56,6 +56,7 @@ default_artifacts:
   aarch64:
     - aws
     - azure
+    - gcp
   ppc64le:
     - powervs
   s390x:


dustymabe:
This would need to be done with some thought. For example

- we'd need to enable testing for these images
- we have image families in GCP. We'd need to think about the implications of this there.

Ultimately this ties back in to: https://github.com/coreos/fedora-coreos-tracker/issues/1377

dustymabe:
This is completed now. See https://github.com/coreos/fedora-coreos-pipeline/pull/879

PR #856, opened 2023-04-13
kola-aws: Snooze xen-tests for rawhide stream

gursewak1997:
Snooze this test for the month of April while we wait for a fix for https://github.com/coreos/fedora-coreos-tracker/issues/1438

### DIFF
--- a/jobs/kola-aws.Jenkinsfile
+++ b/jobs/kola-aws.Jenkinsfile
@@ -86,23 +86,29 @@ cosaPod(memory: "512Mi", kvm: false,
                 if (tests == "") {
                     tests = "basic"
                 }
-                parallelruns['Kola:Xen'] = {
-                    // https://github.com/coreos/fedora-coreos-tracker/issues/997
-                    // Run this test on i3.large so we can also run ext.config.platforms.aws.nvme
-                    // to verify access to instance storage nvme disks works
-                    // https://github.com/coreos/fedora-coreos-tracker/issues/1306
-                    // Also add in the ext.config.platforms.aws.assert-xen test just
-                    // to sanity check we are on a Xen instance.
-                    def xen_tests = tests
-                    if (xen_tests == "basic") {
-                        xen_tests = "basic ext.config.platforms.aws.nvme ext.config.platforms.aws.assert-xen"
+                // Snooze the xen_tests for rawhide for this month while we wait
+                // for a fix for https://github.com/coreos/fedora-coreos-tracker/issues/1438
+                def now = Calendar.getInstance()
+                def month = now.get(Calendar.MONTH)
+                if (params.STREAM != "rawhide" || month != Calendar.APRIL) {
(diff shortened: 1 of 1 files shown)


jobs/kola-aws.Jenkinsfile
```diff
@@ -86,23 +86,25 @@ cosaPod(memory: "512Mi", kvm: false,
...
-                    // https://github.com/coreos/fedora-coreos-tracker/issues/997
-                    // Run this test on i3.large so we can also run ext.config.platforms.aws.nvme
-                    // to verify access to instance storage nvme disks works
-                    // https://github.com/coreos/fedora-coreos-tracker/issues/1306
-                    // Also add in the ext.config.platforms.aws.assert-xen test just
-                    // to sanity check we are on a Xen instance.
-                    def xen_tests = tests
-                    if (xen_tests == "basic") {
-                        xen_tests = "basic ext.config.platforms.aws.nvme ext.config.platforms.aws.assert-xen"
+                if (params.STREAM != "rawhide" || LocalDate.now().monthValue != 4) {
```
dustymabe, line 89: Might be worth adding a comment above this and link to the tracker issue.

jobs/kola-aws.Jenkinsfile
```diff
@@ -86,23 +86,27 @@ cosaPod(memory: "512Mi", kvm: false,
...
-                    // to verify access to instance storage nvme disks works
-                    // https://github.com/coreos/fedora-coreos-tracker/issues/1306
-                    // Also add in the ext.config.platforms.aws.assert-xen test just
-                    // to sanity check we are on a Xen instance.
-                    def xen_tests = tests
-                    if (xen_tests == "basic") {
-                        xen_tests = "basic ext.config.platforms.aws.nvme ext.config.platforms.aws.assert-xen"
+                // Snooze the xen_tests for rawhide for this month while we wait
+                // for a fix for https://github.com/coreos/fedora-coreos-tracker/issues/1438 
+                if (params.STREAM != "rawhide" || LocalDate.now().monthValue != 4) {
```
jlebon, line 91: Does this work? Testing in a throwaway job, it fails on `LocalDate`.

I think you can do something like:

```groovy
def now = Calendar.getInstance()
def month = now.get(Calendar.MONTH)
if (params.STREAM != "rawhide" || month != Calendar.APRIL) {
    ...
}
```

gursewak1997: Huh that's unexpected, as [LocalDate](https://docs.groovy-lang.org/latest/html/groovy-jdk/java/time/LocalDate.html) should work but I will edit this for now.

jlebon (approved):

jobs/kola-aws.Jenkinsfile
jlebon: It does indeed work, but requires an `import java.time.LocalDate` at the top.

PR #858, opened 2023-04-14
config.yaml: disable next-devel

dustymabe:
Now that testing-devel is moving back over to F38 we don't have any differences in testing-devel and next-devel so we'll disable it.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -17,8 +17,8 @@ streams:
     testing-devel:
       type: development
       default: true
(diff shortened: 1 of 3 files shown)


dustymabe:
need to wait on this [until Tuesday](https://github.com/coreos/fedora-coreos-config/pull/2376#issuecomment-1509223455).

jlebon (approved):

PR #859, opened 2023-04-18
jobs/bump-lockfile: run testiso in a stage

dustymabe:
Running it in a stage will cause it to show up properly in the Blue Ocean view.

### DIFF
--- a/jobs/bump-lockfile.Jenkinsfile
+++ b/jobs/bump-lockfile.Jenkinsfile
@@ -224,7 +224,9 @@ lock(resource: "bump-${params.STREAM}") {
                             // compressed one
                             shwrap("cosa compress --artifact=metal")
                         }
-                        kolaTestIso(cosaDir: env.WORKSPACE, arch: arch, marker: arch)
+                        stage("${arch}:kola:testiso") {
+                            kolaTestIso(cosaDir: env.WORKSPACE, arch: arch, marker: arch)
+                        }
                     }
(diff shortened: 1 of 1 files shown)


dustymabe:
Another option is to make the coreos-ci-lib function run it in a stage from within the function and then un-stage calls to it in this repo.

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -224,7 +224,9 @@ lock(resource: "bump-${params.STREAM}") {
                             // compressed one
                             shwrap("cosa compress --artifact=metal")
                         }
-                        kolaTestIso(cosaDir: env.WORKSPACE, arch: arch, marker: arch)
+                        stage("${arch}:Kola:TestISO") {
```
jlebon, line 227: Should it be `${arch}:kola:testiso` for consistency with the `${arch}:kola` and `${arch}:kola:upgrade`?

jlebon (approved):
No strong opinion. I'd say it's probably cleaner to keep it here rather than coreos-ci-lib, because you can't "unstage" it as a caller if you don't want it for whatever reason (e.g. you're calling it as part of parallel branches and it'd mess up the BlueOcean view).

jobs/bump-lockfile.Jenkinsfile
dustymabe: yep. updated.

PR #860, opened 2023-04-19
add support for kubevirt container image pushing

dustymabe:


jobs/release.Jenkinsfile
```diff
@@ -207,6 +207,7 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
         // OCP ART doesn't actually care what the tag name is (it's just to stop GC), we
         // hardcode it.
         def push_containers = ['oscontainer': ['ostree', 'base-oscontainer', ''],
+                               'kubevirt': ['kubevirt', 'kubevirt', ''],
```
dustymabe, line 210: I think this one is tricky because I think upstream we don't want a tag suffix, but downstream we do.

@jlebon does that match your understanding?

gursewak1997:
Not sure about the tag suffix requirements, but the rest lgtm

config.yaml
```diff
@@ -43,6 +43,7 @@ s3:
 
 registry_repos:
   oscontainer: quay.io/fedora/fedora-coreos
+  kubevirt: quay.io/fedora/fedora-coreos-kubevirt
```
jlebon, line 46: We should add an entry to `docs/config.yaml` too.

jobs/release.Jenkinsfile
jlebon: Indeed. We can't not add a suffix for this downstream. It'll conflict with the base-oscontainer. Upstream, we could push as `quay.io/fedora/fedora-coreos-kubevirt:stable-kubevirt`, but that looks awkward. Probably time to make this configurable. One idea is to make it templated, like we do for other keys. E.g.:

```yaml
registry_repos:
    kubevirt: quay.io/example/monorepo:${STREAM}
```

or

```yaml
registry_repos:
    kubevirt: quay.io/example/monorepo:${STREAM}-suffix
```

To make it backwards-compatible, we could have the third element in the tuple here represent the default tag template when none is given, so e.g.

```groovy
'extensions': ['extensions-container', 'extensions-container', '${STREAM}-extensions']
```

dustymabe: I liked this idea when I first read it, but I'm starting to think it's not complete enough.

we also have the `registry_repos.add_build_tag` piece that adds `--tag=${params.VERSION}${tag_suffix}` so we'd have to split out `tag_suffix` from `quay.io/example/monorepo:${STREAM}-suffix`

Maybe instead we should just add more flexibility in the config spec:

```
registry_repos:
    kubevirt:
        registry: quay.io/example/monorepo
        tags:
          - ${STREAM}-suffix
          - ${VERSION}-suffix
```

IIUC we don't need to preserve backwards compat here because we use the latest pipeline code and latest pipecfg (i.e. there are only one copy of them for all branches and we can update them at the same time).

jlebon: Yup, making the templating bits a list works for me too! The `add_build_tag` knob is definitely a bit of a hack, so it'd be good to get rid of it. The part about backcompat was more to provide a path to make this move more seamless, not that we can't change everything if we wanted to.

dustymabe: ok pushed a new upload that should take care of this.

dustymabe:
ok I pushed up a new upload here. This should handle all outstanding concerns. I tested this in the downstream devel pipeline.

dustymabe:
Let's hold this until this week's releases are out.

config.yaml
```diff
@@ -42,8 +42,12 @@ s3:
   acl: public-read
 
 registry_repos:
-  oscontainer: quay.io/fedora/fedora-coreos
-  kubevirt: quay.io/fedora/fedora-coreos-kubevirt
+  oscontainer:
+	registry: quay.io/fedora/fedora-coreos
```
jlebon, line 46: Tab character here

docs/config.yaml
```diff
@@ -113,19 +115,27 @@ s3:
 # OPTIONAL: container registry-related keys
 registry_repos:
   # OPTIONAL: repo to which to push oscontainer
-  oscontainer: quay.io/fedora/fedora-coreos
+  oscontainer:
+	registry: quay.io/fedora/fedora-coreos
```
jlebon, line 119: Tab character here

docs/config.yaml
```diff
@@ -113,19 +115,27 @@ s3:
...
+    tags: ["${STREAM}"]
```
jlebon, line 120: Should add `REQUIRED` markers on these and docstrings for consistency.

jlebon, line 120: I know it seems obvious since we're using it in the examples, but for completeness, can we explicitly document the supported variables?

utils.groovy
```diff
@@ -417,10 +417,23 @@ def build_artifacts(pipecfg, stream, basearch) {
...
+def get_registry_repos(pipecfg, stream, version) {
     def registry_repos = pipecfg.registry_repos ?: [:]
     // merge top-level registry_repos with stream-specific bits
     registry_repos += pipecfg.streams[stream].additional_registry_repos ?: [:]
+    for (repo in (registry_repos.keySet() as List)) {
+        if (repo == 'v2s2') {
+            // this is a boolean option, not a registry repo
+            continue
+        }
+        if (registry_repos."${repo}".tags) {
```
jlebon, line 429: Minor/optional: could also do

```groovy
registry_repos[repos].tags
```

which looks cleaner.

config.yaml
jlebon, line 46: Hmm, I guess technically this key should be called `repo` rather than `registry`?

jobs/release.Jenkinsfile
```diff
@@ -239,19 +240,19 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
+                        def repo = registry_repos[configname]['registry']
+                        def (artifact, metajsonname) = val
+                        def tag_args = []
+                        for (tag in registry_repos[configname]['tags']) {
+                            if (pipecfg.hotfix) {
+                                // this is a hotfix build; include the hotfix name
+                                // in the tag suffix so we don't clobber official
+                                // tags
+                                tag += "-hotfix-${pipecfg.hotfix.name}"
+                            }
```
jlebon, line 252: Seems like it'd be cleaner to have this be part of `get_registry_repos` now that it does tag stuff?

jobs/release.Jenkinsfile
```diff
@@ -239,19 +240,19 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
                     withCredentials([file(variable: 'REGISTRY_SECRET',
                                           credentialsId: 'oscontainer-push-registry-secret')]) {
-                        def repo = registry_repos[configname]
-                        def (artifact, metajsonname, tag_suffix) = val
-                        if (pipecfg.hotfix) {
-                            // this is a hotfix build; include the hotfix name
-                            // in the tag suffix so we don't clobber official
-                            // tags
-                            tag_suffix += "-hotfix-${pipecfg.hotfix.name}"
+                        def repo = registry_repos[configname]['registry']
```
jlebon, line 243: Hmm, interesting that here we went with the opposite notation. :) Regardless of which one we go with, let's have it match `get_registry_repos`?

jobs/release.Jenkinsfile
jlebon: And then here we'd have:

```groovy
def tag_args = registry_repos[configname].tags.collect{"--tag=$it"}
```

jlebon (commented):
Can we prefix the last commit title with e.g. `config:`?
Also, there's a "wentb" typo in the last commit message.

dustymabe:
ok I think I addressed all comments. I need to run this through another round of testing; will report back on that in an hour.

docs/config.yaml
dustymabe: added REQUIRED docstrings

dustymabe: I did in the first instance; didn't for the others.

jlebon: We could even drop the `REQUIRED` comments in the following ones, but cool as is too!

docs/config.yaml
```diff
@@ -113,17 +115,38 @@ s3:
 # OPTIONAL: container registry-related keys
 registry_repos:
   # OPTIONAL: repo to which to push oscontainer
```
jlebon, line 117: Minor: maybe this now can be:

```suggestion
  # OPTIONAL: repo and tags to which to push oscontainer
```

docs/config.yaml
```diff
@@ -113,17 +115,38 @@ s3:
...
-  oscontainer: quay.io/fedora/fedora-coreos
+  oscontainer:
+    # REQUIRED: repo to which to push oscontainer
```
jlebon, line 119: Minor: and here, just

```suggestion
    # REQUIRED: repo name
```

jlebon (approved):
Some minor comments, but LGTM as is too!

jlebon:
For the commit title, how about something like:

> config: add container image tags to push to and drop add_build_tags

?

dustymabe:
> config: add container image tags to push to and drop add_build_tags

Sorry I forgot about this in the last upload. New upload just landed with fixes. I'll start testing this now.

dustymabe:
ok this seems to pass in some basic testing.

gursewak1997 (approved):

gursewak1997:
> ok this seems to pass in some basic testing.

How you are testing this? :eyes:

dustymabe:
> > ok this seems to pass in some basic testing.
> 
> How you are testing this? eyes

Downstream devel pipeline. Look at the most recent runs of the release job.

PR #861, opened 2023-04-20
Revert "jobs: add kola-secex test"

jlebon:
This reverts commit 7f2b7d256d74076d66cc0acce6b3e55286163c30.

It's failing in the RHCOS pipeline with:

    13:42:52  --- FAIL: ext.config.shared.secex.ensure (1.00s)
    13:42:52          harness.go:1607: Cluster failed starting machines: reading hostkey: open /data.secex/hostkeys/HKD-*.crt: no such file or directory

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -306,14 +306,6 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
             pipeutils.build_artifacts(pipecfg, params.STREAM, basearch)
         }
 
-        // secex specific tests
-        if (shwrapCapture("cosa meta --get-value images.qemu-secex") != "None") {
-            stage("Kola:Secex") {
-                kola(cosaDir: env.WORKSPACE, arch: basearch, skipUpgrade: true,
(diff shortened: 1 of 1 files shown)


dustymabe (approved):

nikita-dubrovskii:
@jschintag is the hostkey there ?

jschintag:
Yes, the hostkey is there, i checked. Given that the `cosa buildextend-secex` worked the volume has to be mounted as well. I think that maybe kola is not expanding on the wildcard.

nikita-dubrovskii:
> Yes, the hostkey is there, i checked. Given that the `cosa buildextend-secex` worked the volume has to be mounted as well. I think that maybe kola is not expanding on the wildcard.

Could you please check with the full path to the hostkey? or rename it to something like `se-hostkey`

nikita-dubrovskii:
https://github.com/coreos/fedora-coreos-pipeline/pull/862

jschintag:
> Could you please check with the full path to the hostkey? or rename it to something like `se-hostkey`

I have added a link to the hostkey, so you can try with `/data.secex/hostkeys/secex-hostkey.crt`.

PR #862, opened 2023-04-21
Hostkey fix

nikita-dubrovskii:


jobs/build-arch.Jenkinsfile
```diff
@@ -309,8 +309,9 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
         // secex specific tests
         if (shwrapCapture("cosa meta --get-value images.qemu-secex") != "None") {
             stage("Kola:Secex") {
-                kola(cosaDir: env.WORKSPACE, arch: basearch, skipUpgrade: true,
-                     extraArgs: "--qemu-secex --tag secex --qemu-secex-hostkey /data.secex/hostkeys/HKD-*.crt")
+                def hostkey = shwrapCapture("realpath /data.secex/hostkeys/HKD-*.crt")
```
jlebon, line 312: I think what's going on here is that the glob isn't getting expanded because it tries to expand it on the controller rather than the builder. You'd have to do something like:

```groovy
def hostkey = shwrap("cosa shell -- 'realpath /data.secex/hostkeys/HKD-*.crt'")
```

which isn't great.

@jschintag created a stable symlink: https://github.com/coreos/fedora-coreos-pipeline/pull/861#issuecomment-1517691341. Maybe let's use that instead and avoid globs?

jlebon (commented):
Let's hold this until https://github.com/openshift/os/pull/1268 merges.

jobs/build-arch.Jenkinsfile
nikita-dubrovskii: yep, done

jlebon (approved):

jschintag:
The bump landed and latest `qemu-secex` artifact is working.

PR #863, opened 2023-04-21
jobs/build.Jenkinsfile: Implement koji-tagger delay work around

c4rt0:
See: https://github.com/coreos/fedora-coreos-tracker/issues/1185

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -121,6 +121,26 @@ if (params.WAIT_FOR_RELEASE_JOB) {
     timeout_mins += timeout_mins + 30
 }
 
+def recent_commits_to_lockfiles() {
+    shwrapRc('''
+    [ -z "$(ls ./src/config/manifest-lock.* 2>/dev/null)" ] && exit 1 # no lockfiles exist
+    # Temporary assignment to a particular file:
+    last_log=$(git -C ./src/config/ log -1 --format="%ct" manifest-lock.*)
+    echo "Last commit: $(date -ud @$last_log)"
+    # Number of minutes to check
+    n=120
+    current_time=$(date +%s)
+    minutes_ago=$(( current_time - ( 60 * n ) ))
+    # Check if git log was run within the last n minutes
+    if (( last_log > minutes_ago )); then
+        echo "Last log < $n minutes; implementing --with-cosa-overrides"
+        exit 0
+    else
+        exit 1
+    fi
+    ''')
+}
+
 lock(resource: "build-${params.STREAM}") {
     cosaPod(cpu: "${ncpus}",
             memory: "${cosa_memory_request_mb}Mi",
@@ -228,11 +248,16 @@ lock(resource: "build-${params.STREAM}") {
             new_version = shwrapCapture("/usr/lib/coreos-assembler/fcos-versionary")
         }
 
+        def overrides_fetch_param = ""
+
         // fetch from repos for the current build
         stage('Fetch') {
-            shwrap("""
-            cosa fetch ${strict_build_param}
-            """)
+            // Dont run this for production builds
+            if (recent_commits_to_lockfiles() == 0 && stream_info.type != "production" ) {
+                shwrap("python3 /usr/lib/coreos-assembler/download-overrides.py")
+                overrides_fetch_param = "--with-cosa-overrides"
+            }
+            shwrap("cosa fetch ${overrides_fetch_param} ${strict_build_param}")            
         }
 
         stage('Build OSTree') {


jobs/build.Jenkinsfile
```diff
@@ -231,8 +231,28 @@ lock(resource: "build-${params.STREAM}") {
         // fetch from repos for the current build
         stage('Fetch') {
             shwrap("""
-            cosa fetch ${strict_build_param}
-            """)
+            # Dont run this for production builds
+            if (stream_info.type != "production") {
```
dustymabe, line 235: I think we are mixing groovy and bash here.

jobs/build.Jenkinsfile
```diff
@@ -231,8 +231,28 @@ lock(resource: "build-${params.STREAM}") {
...
+                # Number of minutes to check
+                n=120
+                # Current time
+                current_time=$(date +%s)
+                # Time n minutes ago
+                n_minutes_ago=$(( current_time - ( 60 * n ) ))
+                # Last commit
+                last_git_log=$(git log -1 --follow manifest-lock* --format="%cI")
+                # Check if git log was run within the last n minutes
+                if (( last_git_log > n_minutes_ago )); then
```
dustymabe, line 247: let's export this logic into another function and just call it.. the code here would then look something like:

```
def overrides_fetch_param = ""
if (stream_info.type != "production" && recent_commits_to_lockfiles()) {
    overrides_fetch_param = "--with-cosa-overrides"
}
shwrap("""
cosa fetch ${overrides_fetch_param} ${strict_build_param}
""")
```

jobs/build.Jenkinsfile
```diff
@@ -230,9 +230,38 @@ lock(resource: "build-${params.STREAM}") {
...
+                    echo "Last commit was authored less than $n minutes ago."
+                    # Workaround implemented, use --with-cosa-overrides
+                    return 0
```
c4rt0, line 252: @dustymabe I tried to implement your suggestion, but not sure if this is going to work, so pushing it just for further discussion.

jobs/build.Jenkinsfile
```diff
@@ -230,9 +230,38 @@ lock(resource: "build-${params.STREAM}") {
...
+                # Last author time in UTC
+                echo "Last commit: $(date -ud @$(git log -1 --follow manifest-lock* --format="%ct"))" 
+                # Number of minutes to check
+                n=120
+                # Current time
+                current_time=$(date +%s)
+                # Time n minutes ago
+                n_minutes_ago=$(( current_time - ( 60 * n ) ))
+                # Last commit
+                last_git_log=$(git log -1 --follow manifest-lock* --format="%cI")
```
dustymabe, line 247: it appears we need  to use: 

```suggestion
                last_git_log=$(git log -1 --format="%cI" manifest-lock*)
```

The other ways gave errors:

```
[dustymabe@media fedora-coreos-config (upstreamtesting-devel %=)]$ git log -1 --follow manifest-lock* --format="%cI"
fatal: option '--format=%cI' must come before non-option arguments
[dustymabe@media fedora-coreos-config (upstreamtesting-devel %=)]$ git log -1 --follow --format="%cI" manifest-lock*
fatal: --follow requires exactly one pathspec
[dustymabe@media fedora-coreos-config (upstreamtesting-devel %=)]$ git log -1 --format="%cI" manifest-lock*
2023-05-05T03:03:10+00:00
```

jobs/build.Jenkinsfile
```diff
@@ -230,9 +230,38 @@ lock(resource: "build-${params.STREAM}") {
 
         // fetch from repos for the current build
         stage('Fetch') {
+            def overrides_fetch_param = ""
+            # Dont run this for production builds
+            recent_commits_to_lockfiles() {
+                // Checking if last commits meet the criteria
+                shwrap("""
```
dustymabe, line 237: I think we need to use `shwrapRc` here and then check for `0` or `1`. See other examples from this repo of using `shwrapRc`. 
```suggestion
                shwrapRc("""
```

jobs/build.Jenkinsfile
dustymabe: also the output of `git log` isn't really giving us the time in the format we need it to be in to make the comparision you are making in the if statement below.

ok yes we need to use something like:

```
$ git log -1 --date=unix --format="%cd" manifest-lock*
1683255790
```

jobs/build.Jenkinsfile
```diff
@@ -230,9 +230,38 @@ lock(resource: "build-${params.STREAM}") {
 
         // fetch from repos for the current build
         stage('Fetch') {
+            def overrides_fetch_param = ""
+            # Dont run this for production builds
+            recent_commits_to_lockfiles() {
```
dustymabe, line 235: let's move this function out of the code path here (so it doesn't cloud reading the code). For example [look](https://github.com/coreos/fedora-coreos-pipeline/blob/a31b30aaaeac99a562c6971a4a29c598b61f2231/jobs/bump-lockfile.Jenkinsfile#L48) at how `getLockfileInfo()` is above the main set of code in bump-lockfile.Jenkinsfile.

jobs/build.Jenkinsfile
c4rt0: This is now amended.

jobs/build.Jenkinsfile
c4rt0: Thanks for the above - can't find any documentation on the shwrap | shwrapRc, got any links?

jobs/build.Jenkinsfile
c4rt0: Thanks for all above @dustymabe!

c4rt0:
Ok, so the code passed the `continuous-integration/jenkins/pr-merge` test for the above after replacing the commonly used apostrophe after `shwrapRc("""` with `shwrapRc('''`. I have a feeling this isn't the correct solution to the problem.

c4rt0:
Temporarily code above considers [manifest-lock.x86_64.json](https://github.com/coreos/fedora-coreos-pipeline/pull/863/files#diff-0bcfa2f93179e91968d7396bcf4881e343ec8e12da5f886c5cdae515e93a30d3R234) as a default file for date verification (`manifest_file="manifest-lock.x86_64.json"`). Just as an update, I am wondering now how to implement platform-specific file selection for the manifest-lock*. In other words: when the date on x86_64 matters more than ppc64le or s390x.

dustymabe:
> Temporarily code above considers [manifest-lock.x86_64.json](https://github.com/coreos/fedora-coreos-pipeline/pull/863/files#diff-0bcfa2f93179e91968d7396bcf4881e343ec8e12da5f886c5cdae515e93a30d3R234) as a default file for date verification (`manifest_file="manifest-lock.x86_64.json"`). Just as an update, I am wondering now how to implement platform-specific file selection for the manifest-lock*. In other words: when the date on x86_64 matters more than ppc64le or s390x.

You need to consider all lockfiles (i.e. overrides lockfiles too) I would just glob all lockfiles (`manifest-lock.*`).

c4rt0:
> > Temporarily code above considers [manifest-lock.x86_64.json](https://github.com/coreos/fedora-coreos-pipeline/pull/863/files#diff-0bcfa2f93179e91968d7396bcf4881e343ec8e12da5f886c5cdae515e93a30d3R234) as a default file for date verification (`manifest_file="manifest-lock.x86_64.json"`). Just as an update, I am wondering now how to implement platform-specific file selection for the manifest-lock*. In other words: when the date on x86_64 matters more than ppc64le or s390x.
> 
> You need to consider all lockfiles (i.e. overrides lockfiles too) I would just glob all lockfiles (`manifest-lock.*`).

I changed the above to a single file since `git log` allows for the use of only one parameter. 
In light of your suggestion, I will now work on a solution that checks the latest date on all of the files covered by `manifest-lock*`.

c4rt0:
After final modifications above implementation successfully use the `--with-cosa-overrides` once conditions are met. 
The build however fails:
`15:20:01  Error: unknown command "tags=needs-internet" for "kola"`

dustymabe:
> After final modifications above implementation successfully use the `--with-cosa-overrides` once conditions are met. The build however fails: `15:20:01 Error: unknown command "tags=needs-internet" for "kola"`

The version of coreos-ci-lib was using a fork in the jenkins-casc-config configmap. I switched it to use the `main` branch of https://github.com/coreos/coreos-ci-lib.git

c4rt0:
> 



> > After final modifications above implementation successfully use the `--with-cosa-overrides` once conditions are met. The build however fails: `15:20:01 Error: unknown command "tags=needs-internet" for "kola"`
> 
> The version of coreos-ci-lib was using a fork in the jenkins-casc-config configmap. I switched it to use the `main` branch of https://github.com/coreos/coreos-ci-lib.git

Thank you, this build now passed. 🎊

c4rt0:
I spent some time today looking at the [coreos-koji-tagger](https://github.com/coreos/fedora-coreos-releng-automation/blob/main/coreos-koji-tagger/coreos_koji_tagger.py) code.
I'm aware of @jlebon comment on the implementation of the [download-overrides.py](https://github.com/coreos/fedora-coreos-config/blob/a85766ed7a315e3c09629dbf11cc5bdf02560cfc/.cci.jenkinsfile#L15-L16), I'm just somewhat confused :
Should I execute it the same way in [here](https://github.com/coreos/fedora-coreos-pipeline/pull/863/files#diff-0bcfa2f93179e91968d7396bcf4881e343ec8e12da5f886c5cdae515e93a30d3R255) or is there a better way to do it?

jobs/build.Jenkinsfile
```diff
@@ -228,11 +228,34 @@ lock(resource: "build-${params.STREAM}") {
...
+            # Number of minutes to check
+            n=120
+            current_time=$(date +%s)
+            minutes_ago=$(( current_time - ( 60 * n ) ))
+            last_log=$(git log -1 --format="%ct" manifest-lock.*)
+            cd ../../
+            # Check if git log was run within the last n minutes
+            if (( last_log > minutes_ago )); then
+                echo "Last log < $n minutes; implementing --with-cosa-overrides"
+                python3 download-overrides.py
```
dustymabe, line 245:
```suggestion
                python3 /usr/lib/coreos-assembler/download-overrides.py
```

dustymabe: also I would put this down on line 256.. this function just tells you whether or not there were recent commits to lockfiles. It shouldn't actually do anything.

jobs/build.Jenkinsfile
```diff
@@ -228,11 +228,34 @@ lock(resource: "build-${params.STREAM}") {
...
+            fi
+        ''')
+
         // fetch from repos for the current build
         stage('Fetch') {
-            shwrap("""
-            cosa fetch ${strict_build_param}
-            """)
+            // Dont run this for production builds
+            if (recent_commits_to_lockfiles == 0 && stream_info.type != "production" ) {
```
dustymabe, line 255: I think you need `()` since it's a function call:

```suggestion
            if (recent_commits_to_lockfiles() == 0 && stream_info.type != "production" ) {
```

dustymabe:
> I guess I'm confused in regards to what should be a follow-up task, and what's there to be done now 🤷🏼

If we get something working we can merge it and then start working on the followup.

jobs/build.Jenkinsfile
```diff
@@ -228,11 +228,34 @@ lock(resource: "build-${params.STREAM}") {
...
+            # Number of minutes to check
+            n=120
+            current_time=$(date +%s)
+            minutes_ago=$(( current_time - ( 60 * n ) ))
+            last_log=$(git log -1 --format="%ct" manifest-lock.*)
+            cd ../../
+            # Check if git log was run within the last n minutes
+            if (( last_log > minutes_ago )); then
+                echo "Last log < $n minutes; implementing --with-cosa-overrides"
+                python3 /usr/lib/coreos-assembler/download-overrides.py
```
dustymabe, line 245: This still needs to move down to line 256 (i.e. insidte the `if (recent_commits_to_lockfiles()...`).

c4rt0: Oh, ok - I was wondering if that wasn't the right place...

> This still needs to move down to line 256 (i.e. insidte the `if (recent_commits_to_lockfiles()...`).

Oh, ok - I was wondering if that wasn't the right place...

c4rt0:
I have soooo many questions related to the above 🫣

dustymabe:
pushed up some fixes so we can get this in and test it further.

c4rt0:
> pushed up some fixes so we can get this in and test it further.

Not exactly 'a few simple tweaks' but works for me 😃 
Thanks Dusty!

dustymabe (approved):

PR #866, opened 2023-04-25
jobs/build-arch: temporarily disable secex tests

jlebon:
The tests don't work on the 4.12 pipeline because cosa is missing some
patches. We need to backport at least https://github.com/coreos/coreos-
assembler/pull/3384 but it gets messy because we don't want to also
backport the Ignition config GPG key stuff. Possibly we'll just need a
temporary hack key there in the future.

Anyway, for now let's at least unblock 4.12.


marmijo (approved):
LGTM

PR #867, opened 2023-05-01
multi-arch-builders: add s390x steps for another region

dustymabe:
We've occasionally seen some failures and/or capacity issues so we'll document `us-east-1` and `ca-tor-1`.


marmijo (approved):
LGTM

dustymabe:
Thanks @marmijo - will merge tomorrow (allowing for others to review and/or ask questions).

PR #868, opened 2023-05-02
jobs/kola-{upgrade,openstack}: update --allow-rerun-success options

dustymabe:
This option to `kola run` now takes a string as input. See https://github.com/coreos/coreos-assembler/pull/3430.


nikita-dubrovskii:
LGTM!

PR #869, opened 2023-05-03
jobs/release: rework buildfetch to support kubevirt (x86_64 only)

dustymabe:
Previously we were downloading all listed artifacts for all of the architectures with a `cosa buildfetch --arch=x86_64 --arch=aarch64` etc. For kubevirt we are only building for x86_64 so we need to rework this to make separate buildfetch calls, one for each architecture.

### DIFF
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -132,21 +132,21 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
         // modify/update meta.json because buildfetch will re-download and
         // overwrite meta.json.
         stage('Fetch Artifacts') {
-            // We need to fetch a few artifacts if they were built. This assumes if
-            // it was built for one platform it was built for all.
-            def fetch_artifacts = ['ostree', 'extensions-container', 'legacy-oscontainer']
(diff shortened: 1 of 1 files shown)


cgwalters (commented):
Superficial LGTM.

Offhand it seems like the architecture iteration logic would better live directly in coreos-assembler, like `cosa buildfetch --all-architectures`.

dustymabe:
> Superficial LGTM.

Thanks!

> 
> Offhand it seems like the architecture iteration logic would better live directly in coreos-assembler, like `cosa buildfetch --all-architectures`.

`cosa buildfetch --arch=all` exists but if you specify `cosa buildfetch --arch=all --artifact=kubevirt` (since we are downloading artifacts here) it will fail. This is the correct behavior in my opinion.

jlebon (approved):
Might be cleaner to just separate out the `kubevirt` call since it's special, but LGTM as is too.

dustymabe:
> Might be cleaner to just separate out the `kubevirt` call since it's special, but LGTM as is too.

The reason I didn't is because we may have other things in the future that are specific to an architecture or a subset of architectures. This is now generic and should support non-kubevirt cases too.

Issue #842, opened 2023-03-22
Add an option to flip on kola `--allow-rerun-success`


cgwalters:
A lot of our tests contact the Internet.  These tests *will* flake (Particularly things trying to contact the Fedora yum repos, but not exclusively that of course).  It's not something which should stop us from shipping OS updates.  It's not something which should occupy any significant fraction of human time.

Today, kola has `--allow-rerun-success` that defaults to `false`.  I'd like a pipeline knob to flip this on.  And I definitely plan to do so for the RHCOS pipeline at least.  (It looks like the FCOS pipeline hardcodes this on only for OpenStack?)

dustymabe:
> A lot of our tests contact the Internet. These tests _will_ flake (Particularly things trying to contact the Fedora yum repos, but not exclusively that of course). It's not something which should stop us from shipping OS updates. It's not something which should occupy any significant fraction of human time.

Indeed. They will flake.


> 
> Today, kola has `--allow-rerun-success` that defaults to `false`. I'd like a pipeline knob to flip this on. And I definitely plan to do so for the RHCOS pipeline at least. (It looks like the FCOS pipeline hardcodes this on only for OpenStack?)

It was targeted to Openstack: https://github.com/coreos/fedora-coreos-pipeline/pull/523

We've discussed this in the past and I thought there was an issue that captured that discussion, but I can't find it right now. So we can let this issue capture the discussion.

In the past we've wanted better handling for flakes here, but we don't want to go all the way to what `--allow-rerun-success` is today. Ideas we've had:

- use `--allow-rerun-success` for tests that need internet
- rerun flakes multiple times (i.e. 3) and if it passes all (or maybe a high percentage) we allow success.

Either of those require work in COSA.

jlebon:
@nikita-dubrovskii, @dustymabe and I had a discussion about this today. There are varying hammer sizes we could use here for `--allow-rerun-success` depending on how many flakes we want to try to catch.

To start, we decided to support something like `--allow-rerun-success tag=needs-internet` to limit it to just tests that need the Internet. In the future, we may expand the syntax for other things (e.g. specific tests, test patterns, multiple reruns to consider success, etc...).

Some other things discussed were:
- for flakes that could indicate an issue in our code, ideally we'd track it in an issue
- we could then extend the `kola-denylist.yaml` schema for also expressing flakes and e.g. how many reruns we want to consider success

dustymabe:
[...]

PR #871, opened 2023-05-05
jobs/kola-{openstack,upgrade): adjust to coreos-ci-lib allowRerunSuccess changes

dustymabe:
In https://github.com/coreos/coreos-ci-lib/pull/148 we are modifying the defaults to now call kola run with `--allow-rerun-success tags=needs-internet`, which is part of https://github.com/coreos/fedora-coreos-pipeline/issues/842.

Here we drop the `--allow-rerun-success tags=all` call from the upgrade test. Since the test itself has the `needs-internet` tag then it will be covered by the new default.

For the openstack test we'll adjust to the new way to override the argument string passed to `--allow-rerun-success` inside of `kola()`.


gursewak1997 (approved):
Lgtm

PR #872, opened 2023-05-10
kola-aws: Drop snooze for xen-tests

gursewak1997:
The latest kernel version doesn't result in kernel-oops for the zen tests so dropping the snooze on the the test. Since this snooze expired last month, the test's been running successfully on latest rawhide builds.
Closes: https://github.com/coreos/fedora-coreos-tracker/issues/1438


jlebon:
Minor: can we express this as a `git revert` or does it not apply cleanly? OK as is too.

gursewak1997:
> Minor: can we express this as a `git revert` or does it not apply cleanly? OK as is too.

Done

jlebon (approved):

PR #874, opened 2023-05-23
gcp: run tests on confidential instance

HuijingHei:


HuijingHei:
Hold on until https://github.com/coreos/fedora-coreos-config/pull/2425 merged and https://issues.redhat.com/browse/OCPBUGS-7582 is fixed.

HuijingHei:
https://github.com/coreos/fedora-coreos-config/pull/2425 is merged, but it will not run by default.
Let's run basic test on confidential instance first, will add `ext.config.platforms.gcp.nvme-symlink` when https://issues.redhat.com/browse/OCPBUGS-7582 is fixed.
cc @dustymabe

dustymabe:
> [coreos/fedora-coreos-config#2425](https://github.com/coreos/fedora-coreos-config/pull/2425) is merged, but it will not run by default. Let's run basic test on confidential instance first, will add `ext.config.platforms.gcp.nvme-symlink` when [issues.redhat.com/browse/OCPBUGS-7582](https://issues.redhat.com/browse/OCPBUGS-7582) is fixed. cc @dustymabe

Yes. We should be able to start running the test when https://github.com/coreos/fedora-coreos-config/pull/2450 merges

HuijingHei:
> Yes. We should be able to start running the test when [coreos/fedora-coreos-config#2450](https://github.com/coreos/fedora-coreos-config/pull/2450) merges

Thanks.
One question, if we add the case, will it also be triggered on rhcos (might fail before backport)?

dustymabe:
> One question, if we add the case, will it also be triggered on rhcos (might fail before backport)?

Right now this test won't run in RHCOS pipeline, so we should be safe to merge after https://github.com/coreos/fedora-coreos-config/pull/2450 merges

HuijingHei:
Thanks, added the test as PR2450 is merged.

dustymabe (approved):
LGTM

PR #876, opened 2023-05-26
Enable ppc64le on testing/next

dustymabe:
Once https://github.com/coreos/fedora-coreos-config/pull/2435 merges we should be unblocked for releasing ppc64le. We'll do a soft rollout with just `testing`/`next` for now.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -6,14 +6,8 @@ streams:
         - ppc64le
     testing:
       type: production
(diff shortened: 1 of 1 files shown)


dustymabe:
This one is ready for review now that https://github.com/coreos/fedora-coreos-config/pull/2435 has merged.

gursewak1997 (approved):

PR #877, opened 2023-06-13
jobs/kola-upgrade: enable kola-upgrade test for ppc64le

dustymabe:
Let's set the min_supported_start_versions for ppc64le to F38 now that we are doing builds/releases for F38.


prestist (approved):
LGTM

PR #878, opened 2023-06-15
s390x-rhcos-builder: Remove podman volume workaround

jschintag:
With podman version 4.5.0 it is now possible to skip volumes during prune with a specific label by adding --filter="label!=persistent". Remove the workaround used before to not prune secex-data volume, as it is no longer needed.

### DIFF
--- a/multi-arch-builders/builder-common.bu
+++ b/multi-arch-builders/builder-common.bu
@@ -87,7 +87,7 @@ storage:
           ExecStart=podman image prune --force
           ExecStart=podman image prune --all --force --filter until=48h
           ExecStart=podman container prune --force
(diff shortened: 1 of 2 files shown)


jschintag:
Since we also wanted a session where someone else installs the builder, my suggestion would be to review the RP, let someone else install the builder with this PR and then only merge this after we tested it through reinstalling, since i only tested the modified command in isolation.

dustymabe:
> Since we also wanted a session where someone else installs the builder, my suggestion would be to review the RP, let someone else install the builder with this PR and then only merge this after we tested it through reinstalling, since i only tested the modified command in isolation.

Maybe @nikita-dubrovskii would be interested in going through that exercise

PR #880, opened 2023-06-23
jobs/debug-pod: use a controller image for cosa

dustymabe:
On older RHCOS releases we need to use a newer COSA (one that supports `cosa remote-session`). Let's use the `:main` COSA image always so we know we have the latest bits.

### DIFF
--- a/jobs/debug-pod.Jenkinsfile
+++ b/jobs/debug-pod.Jenkinsfile
@@ -42,6 +42,9 @@ cosa_img = cosa_img ?: pipeutils.get_cosa_img(pipecfg, params.STREAM)
 
 def stream_info = pipecfg.streams[params.STREAM]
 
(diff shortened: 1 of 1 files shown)


jobs/debug-pod.Jenkinsfile
```diff
@@ -42,6 +42,9 @@ cosa_img = cosa_img ?: pipeutils.get_cosa_img(pipecfg, params.STREAM)
 
 def stream_info = pipecfg.streams[params.STREAM]
 
+// We'll just always use latest for the controller image here.
+def cosa_controller_img = "quay.io/coreos-assembler/coreos-assembler:main",
```
aaradhak, line 46: the comma needs to be removed at the end. it is throwing an error I believe

dustymabe: fixed

PR #881, opened 2023-06-26
utils: snooze kola-azure tests on rawhide

marmijo:
Snooze the kola-azure cloud test on rawhide for the remainder of June and July while we wait for a fix for:
https://github.com/coreos/fedora-coreos-tracker/issues/1515

PR #882, opened 2023-06-26
utils: fix typo in kola-azure snooze condition

marmijo:
I made a mistake when accessing the `params` variable in my previous PR: https://github.com/coreos/fedora-coreos-pipeline/pull/881.

Update the condition to snooze kola-azure by accessing the `stream` variable directly.


aaradhak (approved):
LGTM

PR #884, opened 2023-06-27
config: enable ppc64le arch on the stable stream

dustymabe:
https://github.com/coreos/fedora-coreos-tracker/issues/987


gursewak1997 (approved):
lgtm

PR #886, opened 2023-06-30
release: Merge brew job with release job

ravanelli:
- We decided to add brew into the release job, since make sense to upload the metadata to brew everytime we release something.


ravanelli:
/hold 

We need it first https://github.com/coreos/coreos-assembler/pull/3519/files as the backports for it

jobs/release.Jenkinsfile
```diff
@@ -275,6 +282,39 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
+                    def nvr = "rhcos-${arch}-${version}-0"
+                    state = shwrapCapture("""
+                    coreos-assembler koji-upload search \
+                        --nvr ${nvr} \
+                        --keytab "/run/kubernetes/secrets/brew-keytab/brew.keytab" \
+                        --owner ${brew_principal} \
+                        --profile ${brew_profile}
+                    """)
+                    // Check if no Brew upload was done yet
+                    // State 1 means brew build complete
```
dustymabe, line 300: can we link here to documentation that describes what the different states could be?

jobs/release.Jenkinsfile
```diff
@@ -275,6 +282,39 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
+                    def state = false
+                    def version = (params.VERSION).replaceAll("-", ".")
+                    def nvr = "rhcos-${arch}-${version}-0"
+                    state = shwrapCapture("""
+                    coreos-assembler koji-upload search \
+                        --nvr ${nvr} \
+                        --keytab "/run/kubernetes/secrets/brew-keytab/brew.keytab" \
+                        --owner ${brew_principal} \
+                        --profile ${brew_profile}
+                    """)
```
dustymabe, line 298: what gets returned here if the build doesn't exist?

jobs/release.Jenkinsfile
```diff
@@ -275,6 +282,39 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
             }
         }
 
+        if (brew_profile) {
+            stage('Brew Upload') {
+                def tag = pipecfg.streams[params.STREAM].brew_tag
+                for (arch in basearches) {
+                    def state = false
+                    def version = (params.VERSION).replaceAll("-", ".")
```
dustymabe, line 290: weird. why is this needed?

ravanelli: Because we do this replace [here](https://github.com/coreos/coreos-assembler/blob/main/src/cmd-koji-upload#L437). What I think make sense to have the rhcos version without '-' in the NVR name. I think we can also remove this replace if we want to, but need to see with koji allow it too, but I guess so.

jobs/release.Jenkinsfile
ravanelli: I mean, the build state itself goes to the output and the return 0 doesn't.

dustymabe: > However, the 'shwrapCapture' needs to have something as output.

What if we made the output was just an empty string? Would that be OK?

jobs/release.Jenkinsfile
dustymabe: I don't think these "replacements" are doing anything in either python or groovy. The first argument is the string you want to replace and the second is the string you want to replace it with.

The buildid already starts out as `413.92.202305261801.0` so there are no `-` characters to replace IIUC. You can see this in the brew upload you shared with me earlier: https://brewweb.stage.engineering.redhat.com/brew/buildinfo?buildID=2292952

ravanelli: That's because it got changed  in the link I sent above before adding it to Brew, if you see the parameters from Jenkins, the Version ID is like:  '413.92.202305261801-0' and with that replace we make it as '413.92.202305261801.0'

dustymabe: oh I see. because of https://github.com/openshift/os/blob/64e21bb5eedf19cdc0b16db235e4d24ca5e86ecf/manifest-rhel-9.2.yaml#L29

so our versions are like:

```
414.92.202307122109-0
```

but it's interesting to note that the version string is [down to the minute](https://github.com/openshift/os/blob/64e21bb5eedf19cdc0b16db235e4d24ca5e86ecf/manifest-rhel-9.2.yaml#L27) which virtually guarantees there will never be anything other than `-0` in the version string. I wonder if we should pivot what we are doing in https://github.com/coreos/coreos-assembler/pull/3533 to make it such that VR represents the rhcos buildid.

i.e.:

```
$ V=414.92.202307122109
$ R=-0
$ echo "${V}${R}"
414.92.202307122109-0
```

This would make it a bit more intuitive I think.

dustymabe (changes requested):

jobs/release.Jenkinsfile
```diff
@@ -6,6 +6,9 @@ node {
     libcloud = load("libcloud.groovy")
 }
 
+def brew_principal = pipecfg.brew.principal
+def brew_profile = pipecfg.brew.profile
```
dustymabe, line 10: what happens if these aren't defined? anything fail?

dustymabe (approved):
LGTM

jobs/release.Jenkinsfile
ravanelli: Nothing fails, it will be defined as empty vars

ravanelli:
All the backport PRs are merged. We are ok to merge this PR now too.

marmijo:
When these jobs were merged, the RHCOS secrets are being passed in when creating the pod in all cases. This means that the secrets are being passed in when creating the pod in FCOS and causing the pods to fail. We'll need to determine how to differentiate between FCOS and RHCOS here.

marmijo (commented):
When these jobs were merged, the RHCOS secrets are being passed in when creating the pod in all cases. This means that the secrets are being passed in when creating the pod in FCOS and causing the pods to fail. We'll need to determine how to differentiate between FCOS and RHCOS here.
https://github.com/coreos/fedora-coreos-pipeline/pull/886/files#diff-4768730b889afbd1361511a1e552ee9a9f29b11a7e578aa3ede5e57db08762acR91-R93

PR #892, opened 2023-07-21
manifests/jenkins.yaml: raise memory_limit to 8GB

jmarrero:
rustc is being killed while building rpm-ostree deps


cgwalters:
I think we want to change https://github.com/coreos/rpm-ostree/blob/ced6de0700cced241fb116b575844d37d5414ca3/.cci.jenkinsfile#L6
(which we should be able to test in your same PR)

I believe the line being changed here is for the Jenkins instance itself, but that Jenkins instance is generally just a "coordinator" that spawns other Kubernetes pods.

jmarrero:
woopsie

PR #897, opened 2023-07-25
Fix secrets definition in release job

ravanelli:
- If a secret doesn't exist it will fail to be mounted, as volume Mounts.
 - For now let's not pass it as argument to cosaPod, if we don't need the secret


marmijo (approved):
LGTM

PR #898, opened 2023-07-27
multi-arch-builders: drop coreos-ppc64le-builder-512e.bu config

dustymabe:
This was added to force a reprovision of the root filesystem on a particular instance we were using for a RHCOS builder, but it's no longer needed. Now when a system has a disk > 100G the root filesystem will get reprovisioned automatically by the work done in https://github.com/coreos/fedora-coreos-config/pull/2320


dustymabe:
added a second commit to drop a stale comment.

marmijo (approved):
LGTM

PR #901, opened 2023-08-03
utils: remove 'snooze kola-openstack' workaround

marmijo:
The kola-openstack tests are passing again. Remove the workaround because the snooze expired on August 1. However, we still need to implement some additional clean up during garbage collection: https://github.com/coreos/coreos-assembler/issues/3550

This reverts commit 18252d4ce7e5dbbc5f1f5ce426cfe61ad7d5c793.


dustymabe (approved):
LGTM

PR #903, opened 2023-08-05
bump-lockfile: improve slack message

dustymabe:
Right now it puts the stream in between the two emojis. Let's put the stream at the end.


marmijo (approved):
LGTM, the square brackets will also keep it consistent with the other jobs too. Thanks for updating this!

PR #904, opened 2023-08-14
config: enable branched stream

aaradhak:
As F39 has branched from rawhide, we are building the branched stream again


marmijo (approved):
LGTM

PR #906, opened 2023-08-23
manifests/jenkins: set file.encoding=UTF-8

dustymabe:
This will allow the emojis that are output by the denylist processing
from kola to show up in the jenkins output rather than being shown
as weird looking binary characters.


marmijo (approved):
LGTM

dustymabe:
So.. I tested this initially on our staging Jenkins and then opened this PR. I then went ahead and deployed this on our prod Jenkins and it doesn't seem to work.

This may be due to the fact that our staging instance version of Jenkins is slightly newer. I guess it could be something else but for now I'll assume when we update the prod jenkins this will start working.

PR #907, opened 2023-08-30
Minor Jenkins plugin updates

marmijo:
The RHCOS Jenkins instance is failing to load because the fast-tracked `kubernetes` and `token-macro` plugins are now outdated. Both of these plugins have updated versions in [openshift/jenkins](https://github.com/openshift/jenkins/blob/master/2/contrib/openshift/base-plugins.txt). Let's remove these fast-tracks as they are no longer needed.

The `antisamy-markup-formatter` plugin is also listed as a dependency for the `timestamper` plugin, so add that to the list of plugins we layer on top of openshift/jenkins.

### DIFF
--- a/jenkins/controller/plugins.txt
+++ b/jenkins/controller/plugins.txt
@@ -34,16 +34,3 @@ splunk-devops:1.9.9
 # The below list are plugins that are also in base-plugins.txt but for
 # some reason or other we need to temporarily freeze or fast-track.
 #
-# NOTE: 20221004 had to fast-track token-macro to 293.v283932a_0a_b_49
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

marmijo:
@jlebon no, the plugin wasn't downloaded automatically. It looks like that functionality [was disabled](https://github.com/openshift/jenkins/commit/911d6cd3294402d89024a70d23fd6118ad87ea4b#diff-9d951448b8ccce670213102a68d25a4479892e10539863b7de3188b3ef2404c1R206) earlier this year.

PR #912, opened 2023-09-22
jobs: limit xz memory usage on multi-arch too

jlebon:
We recently hit `xz` on the ppc64le builder getting OOMkilled. Let's match x86_64 and also limit memory in the multi-arch paths too. Except we're using a more conservative value (though still plenty) since there's no resource-based scheduling on these builders.

We should probably add support for e.g. `cosa compress --xz-memlimit` instead to make this less awkward.

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -335,7 +335,8 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
         }
 
         stage('Archive') {
-            shwrap("cosa compress")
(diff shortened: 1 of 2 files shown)


dustymabe:
maybe we could/should use some value more than 1G? Our smallest builder has 32G of memory, which means many of these could fit even at 2G or 4G.

dustymabe:
I guess the worry is that limiting the memory so much would cause things to take much longer on our slowest builder (ppc64le), which in turn slows everything else down because we have to finish all arches before the next step can run.

jlebon:
2G WFM. Updated!

dustymabe (approved):
LGTM

PR #914, opened 2023-09-22
jobs: bump multi-arch xz limit to 4G

jlebon:
At 2G, we increased compression time from 2m to 15m. Let's bump it a little more to see if we can strike a better balance between performance and memory.

Also fix the gaping typo of having the `cosa compress` invocation being commented out in 11f2ca6.

Fixes: 11f2ca6 ("jobs: limit xz memory usage on multi-arch too")


dustymabe (approved):
LGTM

PR #918, opened 2023-10-10
jobs/build-fcos-buildroot: build for all arches

jlebon:
We initially started building it only for x86_64 because there was no need to build for other arches[[1]], but with the work underway to have multi-arch support in CoreOS CI, that's no longer the case.

[1]: https://github.com/coreos/fedora-coreos-config/issues/1950


dustymabe (approved):
LGTM

PR #920, opened 2023-10-16
jobs/kola-*: trim artifacts to 30

jlebon:
Keeping around the test console logs of the last 100 builds seems excessive. Let's trim it back to 30 to match bump-lockfile. This should help reduce our storage requirements (each kola job right now takes about 1.2G- 1.4G).


aaradhak (approved):
LGTM

marmijo (approved):
LGTM

PR #923, opened 2023-11-08
config: start building hyperv image for aarch64

n1hility:
Follow-up to coreos/fedora-coreos-tracker#1411

I am not familiar with your process, so let me know what I should be following. If you think there should be a new issue for this, for example.  So far after this I am planning to follow up with additional PRs for the web content. 

I tested and verified this works locally on Windows ARM using cost build and cosa buildextend-hyperv.

Thanks!

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -53,6 +53,7 @@ default_artifacts:
     - aws
     - azure
     - gcp
+    - hyperv
   ppc64le:
     - powervs
   s390x:


n1hility:
PTAL @dustymabe any guidance/redirection on process appreciated!

baude:
LGTM, never even crossed my mind

jlebon (approved):
LGTM. Will let @dustymabe review as well in case he has more context.

n1hility:
@dustymabe cool just to confirm you need a second person to verify? (I did test a build and binary on a windows arm device, if you need just one). The challenge is these are only available as physical metal devices atm (mostly consumer), no cloud hosting providers support it (e.g. ec2 bare metal instances do not allow you to provision windows, and azure ampere instances do not support virtualization). That said, I do know someone else on another team who has win arm hardware I could call in a favor for.

dustymabe:
@n1hility just one is enough. I assumed you hadn't done a local build+test but now I see you did:

> I tested and verified this works locally on Windows ARM using cost build and cosa buildextend-hyperv.



> The challenge is these are only available as physical metal devices atm (mostly consumer), no cloud hosting providers support it (e.g. ec2 bare metal instances do not allow you to provision windows, and azure ampere instances do not support virtualization). That said, I do know someone else on another team who has win arm hardware I could call in a favor for.

I wonder if this is evidence that would argue against us doing this. Seems kind of niche? Do we know a group of users we are trying to target here? Is this podman-machine related?

n1hility:
awesome thanks!

dustymabe:
@n1hility can you give the latest `testing-devel` hyperV aarch64 image a spin? download from: https://builds.coreos.fedoraproject.org/browser?stream=testing-devel&arch=aarch64

n1hility:
Sweet. will check it out and get back to you

PR #925, opened 2023-11-09
jobs/release: update logic that makes AMIs public

marmijo:
Remove the not symbol here to ensure it creates the correct behavior where this code doesn't run when `public: false` is set like in the rhcos devel pipeline.

### DIFF
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -369,7 +369,7 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
                 def bucket_prefix = "${pipecfg.s3.bucket}/${pipecfg.s3.builds_key}"
 
                 // make AMIs public if not already the case
-                if (!pipecfg.clouds?.aws?.public) {
+                if (pipecfg.clouds?.aws?.public) {
                     def rc = shwrapRc("""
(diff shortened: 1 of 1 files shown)


marmijo:
This section of code has been running in the rhcos-devel and fcos pipelines, but not in rhcos-art, which I believe this is the opposite behavior from what was intended.

dustymabe:
IIUC: If `public: true` is set then the AMIs are created as public facing [from the start when first created in the build pipeline](https://github.com/coreos/fedora-coreos-pipeline/blob/788477bde9a8ee3745eca366e1e63e9028760bb8/libcloud.groovy#L154-L156). I think this code in the release job is just there to say: if the AMIs weren't created as `public: true` to begin with then we need to set it now.

The release job is always meant to make things public (you are releasing it to the world):

https://github.com/coreos/fedora-coreos-pipeline/blob/788477bde9a8ee3745eca366e1e63e9028760bb8/docs/config.yaml#L168-L170

Now, if you want functionality to be able to run the release job and not make things public then we'd need to make new code/logic for that.

marmijo:
Alright, thanks for the info! I couldnt find where else `public` was being used in the pipeline and I thought this was the opposite from what we want. I'll close out this PR then.

PR #929, opened 2023-11-14
jenkins/plugins: update plugins to latest versions

marmijo:
Update all Jenkins plugins to their latest versions.


dustymabe:
wow - some of those versions changed scheme completely.

dustymabe (approved):

marmijo:
Yes, they were pretty out-of-date :smile:

jlebon:
Did this work when you tested it? Looks like we needed https://github.com/coreos/fedora-coreos-pipeline/pull/930.

marmijo:
Ah, thanks for catching that! I wasn't able to test it at the time as this PR was in response to Jenkins failing to come up with outdated plugins.

PR #930, opened 2023-11-17
jenkins/plugins: make plugin versions compatible with Jenkins

jlebon:
The recent plugin version bumps breaks when building with the current Jenkins version frozen in the cluster. Downgrade some of them to make them compatible.

Fixes 436c1c8 ("jenkins/plugins: update plugins to latest versions").


dustymabe (approved):
LGTM

PR #933, opened 2023-11-27
multi-arch-builders/tofu: Add PowerVs configuration

ravanelli:
* Add Tofu configuration for provisioning our ppc64le instance on PowerVs
 * Include supplementary documentation for our Tofu and PowerVs procedures

### DIFF
--- a/multi-arch-builders/coreos-ppc64le-builder.bu
+++ b/multi-arch-builders/coreos-ppc64le-builder.bu
@@ -6,10 +6,6 @@
 #
 variant: fcos
 version: 1.4.0
-ignition:
-  config:
-    merge:
-      - local: builder-common.ign
 passwd:
   users:
     - name: builder
@@ -23,3 +19,19 @@ storage:
       overwrite: true
       contents:
         inline: coreos-ppc64le-builder
+     # It is a workaround due the IP/Route issue in PowerVs
+     # See more in the ppc64le README
+     - path: /etc/NetworkManager/system-connections/env2.nmconnection
+      mode: 0600
+      contents:
+        inline: |
+          [connection]
+          id=en
+          type=ethernet
+          interface-name=env2
+          [ipv4]
+          address1=10.130.1.149/25,10.130.1.129
+          dns=127.0.0.53;
+          dns-search=
+          may-fail=false
+          method=manual
--- a/multi-arch-builders/provisioning/ppc64le/README.md
+++ b/multi-arch-builders/provisioning/ppc64le/README.md
@@ -1,0 +1,58 @@
+# OpenTofu
+
+ OpenTofu, a Terraform fork, is an open-source infrastructure as code (IaC) tool
+ lets you define both cloud and on-prem resources in human-readable configuration files
+ that you can version, reuse, and share.
+
+ To proceed with the next steps, ensure that 'tofu' is installed on your system.
+ See: https://github.com/opentofu/opentofu/releases
+
+## Before starting
+
+### PowerVS credentials
+
+ - Ensure that you have access to our account.
+ - Verify that the Fedora CoreOS image has been uploaded to the designated bucket.
+   - TODO: Add bucket creation and image upload to tofu
+   - See documetation in how to upload the image manually:
+       https://cloud.ibm.com/docs/power-iaas?topic=power-iaas-deploy-custom-image
+### PowerVs Issues
+
+ - PowerVS seems to encounter a problem in creating the default local IP with the default route,
+resulting in issues to ssh to the server post-boot.
+To mitigate this, we've incorporated networking configurations into the Ignition file. However,
+we still with one issue during the Splunk Butane configuration, where the CA certification couldn't be
+downloaded during provisioning. If you encounter this issue, comment out the Red Hat CA download step
+and perform it manually on the machine after provisioning. 
+
+ - Additionally, it's important to note that PowerVS lacks the user data field in the web interface for providing
+the Ignition config.
+
+### TF vars via environment variables
+
+If you'd like to override the target distro (defaults to `fcos`) you
+can:
+
+```
+export TF_VAR_distro=rhcos
+```
+
+If you are deploying RHCOS you'll need to define variables for splunk configuration:
+
+```
+export TF_VAR_splunk_hostname=...
+export TF_VAR_splunk_sidecar_repo=...
+export TF_VAR_itpaas_splunk_repo=...
+```
+
+## Running tofu
+```bash
+   # To begin using it, run 'init' within this directory.
+   tofu init
+   # If you don't intend to make any changes to the code, simply run it:
+   tofu apply
+   # If you plan to make changes to the code as modules/plugins, go ahead and run it:
+   tofu init -upgrade
+   # To destroy it run:
+   tofu destroy -target aws_instance.coreos-aarch64-builder
+```
--- a/multi-arch-builders/provisioning/ppc64le/provider.tf
+++ b/multi-arch-builders/provisioning/ppc64le/provider.tf
@@ -1,0 +1,18 @@
+terraform {
+  required_providers {
+    ct = {
+      source  = "poseidon/ct"
+      version = "0.13.0"
+    }
+    ibm = {
+      source = "IBM-Cloud/ibm"
+      version = ">= 1.12.0"
+    }
+  }
+}
+
+provider "ibm" {
+    ibmcloud_api_key = var.ibmcloud_api_key
+    region    = "us-south"
+    zone      = var.ibmcloud_zone
+}
--- a/multi-arch-builders/provisioning/ppc64le/variables.tf
+++ b/multi-arch-builders/provisioning/ppc64le/variables.tf
@@ -1,0 +1,87 @@
+
+variable "ibmcloud_api_key" {
+    description = "Denotes the IBM Cloud API key to use"
+    default = ""
+}
+
+variable "ibmcloud_region" {
+    description = "Denotes which IBM Cloud region to connect to"
+    default     = "us-south"
+}
+
+#INSERTED FOR MULTI-ZONE REGION SUCH AS FRANKFURT
+
+variable "ibmcloud_zone" {
+    description = "Denotes which IBM Cloud zone to connect to - .i.e: eu-de-1 eu-de-2  us-south etc."
+    default     = "us-south"
+}
+
+# Got the ID from `ibmcloud resource service-instances --long field` command, refer GUID for the instance
+variable "power_instance_id" {
+    description = "Power Virtual Server instance ID associated with your IBM Cloud account (note that this is NOT the API key)"
+    default = "556eb201-32bf-4ae2-8ab5-dfd7bbe97789"
+}
+
+
+# The PowerVs cost are high, check the price before adding
+# more processors and memory. This number may change
+# due the PowerVs availability.
+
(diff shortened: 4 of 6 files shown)


multi-arch-builders/coreos-ppc64le-builder.bu
```diff
@@ -23,3 +19,23 @@ storage:
...
+          [connection]
+          id=en
+          type=ethernet
+          interface-name=env2
+          [ipv4]
+          address1=10.130.1.149/25,10.130.1.129
+          dns=127.0.0.53;
+          dns-search=
+          may-fail=false
+          method=manual
```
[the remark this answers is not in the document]
dustymabe: ok I see a comment in the readme about it - maybe mention briefly "a problem" here and mention more details are in the readme.

multi-arch-builders/coreos-ppc64le-builder.bu
```diff
@@ -6,10 +6,6 @@
 #
```
dustymabe, line 6: I guess we need to delete the "Merge in the builder-common.ign Ignition file" line above (and should probably also remove it from the `multi-arch-builders/coreos-aarch64-builder.bu`) file too, which should have been done when that file was changed originally.

multi-arch-builders/provisioning/ppc64le/provider.tf
```diff
@@ -0,0 +1,18 @@
...
+      source = "IBM-Cloud/ibm"
+      version = ">= 1.12.0"
+    }
+  }
+}
+
+provider "ibm" {
+    ibmcloud_api_key = var.ibmcloud_api_key
+    region    = "us-south"
+    zone             = var.ibmcloud_zone
```
dustymabe, line 17: minor: spacing looks inconsistent here

multi-arch-builders/provisioning/ppc64le/variables.tf
```diff
@@ -0,0 +1,78 @@
...
+    default     = "us-south"
+}
+
+#INSERTED FOR MULTI-ZONE REGION SUCH AS FRANKFURT
+
+variable "ibmcloud_zone" {
+    description = "Denotes which IBM Cloud zone to connect to - .i.e: eu-de-1 eu-de-2  us-south etc."
+    default     = "us-south"
+}
+
```
dustymabe, line 18: interesting.. so some regions have only one zone and the name of the zone matches the region? I feel like later if they ever add a new zone to those regions it won't be good, but I guess that's their problem :)

multi-arch-builders/provisioning/ppc64le/variables.tf
```diff
@@ -0,0 +1,78 @@
...
+# Got the ID from `ibmcloud resource service-instances --long field` command, refer GUID for the instance
+variable "power_instance_id" {
+    description = "Power Virtual Server instance ID associated with your IBM Cloud account (note that this is NOT the API key)"
+    default = "556eb201-32bf-4ae2-8ab5-dfd7bbe97789"
+}
+
+variable "memory" {
+    description = "Amount of memory (GB) to be allocated to the VM"
+    default     = "50"
+}
```
dustymabe, line 28: Is this just any number we want? For powerVS do they not have fixed instance sizes?

multi-arch-builders/provisioning/ppc64le/variables.tf
```diff
@@ -0,0 +1,78 @@
...
+
+variable "processors" {
+    description = "Number of virtual processors to allocate to the VM"
+    default     = "15"
+}
```
dustymabe, line 33: 15 seems like a random number - would have expected a power of 2 (like 8 or 16) here.

multi-arch-builders/provisioning/ppc64le/variables.tf
```diff
@@ -0,0 +1,78 @@
...
+
+variable "ssh_key_name" {
+    description = "SSH key name in IBM Cloud to be used for SSH logins"
+    default     = ""
+}
+
+variable "shareable" {
+    description = "Should the data volume be shared or not - true/false"
+    default     = "true"
```
dustymabe, line 47: does this enable the data volume to be shared with other instances? do we need that? what are the drawbacks of true versus false?

multi-arch-builders/provisioning/ppc64le/variables.tf
```diff
@@ -0,0 +1,78 @@
...
+}
+
+variable "network" {
+    description = "List of networks that should be attached to the VM - Create this network before running terraform"
```
dustymabe, line 51: Any instructions on how to create that network?

multi-arch-builders/provisioning/ppc64le/variables.tf
```diff
@@ -0,0 +1,78 @@
...
+    default     = "redhat-internal-rhcos" 
+}
+
+variable "system_type" {
+    description = "Type of system on which the VM should be created - s922/e880"
+    default     = "e880"
+}
+
+variable "image_name" {
+    description = "Name of the image from which the VM should be deployed - IBM i image name"
```
dustymabe, line 61: typo?

```suggestion
    description = "Name of the image from which the VM should be deployed - IBM image name"
```

dustymabe (commented):
Thanks for working on this! A few comments

multi-arch-builders/provisioning/ppc64le/variables.tf
ravanelli: Each server model has a maximum allowable number, which may vary depending on utilization (interesting)

On the e880, the maximum core availability is now 12.04, reduced from the previous 15 I added here.
We now have on the e980, with  31.5. I will add a note in how to check it before ,so we may add the max we can.

multi-arch-builders/provisioning/ppc64le/variables.tf
ravanelli: Price I would say. We can change it to false as well

multi-arch-builders/provisioning/ppc64le/variables.tf
ravanelli: There is the internal docs, since it is for VPC. Not sure it is worth adding a general one here, maybe we won't need to create it without using VCP.

multi-arch-builders/provisioning/ppc64le/README.md
```diff
@@ -0,0 +1,57 @@
...
+ See: https://github.com/opentofu/opentofu/releases
+
+## Before starting
+
+### PowerVS credentials
+
+ - Ensure that you have access to our account.
+ - Verify that the Fedora CoreOS image has been uploaded to the designated bucket.
+   - See documetation in how to upload the image:
+       https://cloud.ibm.com/docs/power-iaas?topic=power-iaas-deploy-custom-image
```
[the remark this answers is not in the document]
ravanelli: This case `yes, but we don't want to spend any more time on this for now`

multi-arch-builders/provisioning/ppc64le/README.md
```diff
@@ -0,0 +1,57 @@
...
+### PowerVs Issues
+
+ - PowerVS seems to encounter a problem in creating the default local IP with the default route,
+resulting in issues to ssh to the server post-boot.
+To mitigate this, we've incorporated networking configurations into the Ignition file. However,
+we still with one issue during the Splunk Butane configuration, where the CA certification couldn't be
+downloaded during provisioning. If you encounter this issue, comment out the Red Hat CA download step
+and perform it manually on the machine after provisioning. 
+
+ - Additionally, it's important to note that PowerVS lacks the user data field in the interface for providing
```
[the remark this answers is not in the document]
ravanelli: For the interface, is just to keep track of this issue to.

multi-arch-builders/coreos-ppc64le-builder.bu
ravanelli: Yes, `multi-arch-builders/coreos-aarch64-builder.bu` is already updated, just making the same here

multi-arch-builders/provisioning/ppc64le/variables.tf
ravanelli: For case for below

multi-arch-builders/provisioning/ppc64le/README.md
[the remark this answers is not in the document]
dustymabe: cool. Let's add a comment here (and link if there is a good one).

multi-arch-builders/provisioning/ppc64le/README.md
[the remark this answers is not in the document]
dustymabe: but I don't understand. I'm specifically talking about

> Additionally, it's important to note that PowerVS lacks the user data field in the interface for providing

but if I look at the main.tf we are using:

```
    pi_user_data          = base64encode(data.ct_config.butane.rendered)
```

so that seems like we are passing user data, but the comment says PowerVS doesn't support user data?  Maybe I'm misinterpreting the comment?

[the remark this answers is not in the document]
ravanelli: A, I probably should say in web interface, instead of `user data field in the interface `

multi-arch-builders/provisioning/ppc64le/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+# Got the ID from `ibmcloud resource service-instances --long field` command, refer GUID for the instance
+variable "power_instance_id" {
+    description = "Power Virtual Server instance ID associated with your IBM Cloud account (note that this is NOT the API key)"
+    default = "556eb201-32bf-4ae2-8ab5-dfd7bbe97789"
+}
+
+
+# The PowerVs cost are high, check the price before adding
+# more processors and memory. This number may change
+# due the PowerVs availability.
```
dustymabe, line 28: FTR I think our fcos ppc64le server only has 32G of memory so we could probably go down here if that makes sense if the costs are really high.

multi-arch-builders/provisioning/ppc64le/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+
+# The s922 model is the cheapest model
+variable "system_type" {
+    description = "Type of system on which the VM should be created - s922/e880/e980"
+    default     = "s922"
+}
+
+variable "proc_type" {
+    description = "Processor type for the LPAR - shared/dedicated"
+    default     = "capped"
```
dustymabe, line 48: The description mentions `shared/dedicated` as options but we have set it to `capped`?

multi-arch-builders/provisioning/ppc64le/variables.tf
```diff
@@ -0,0 +1,87 @@
...
+    description = "Amount of memory (GB) to be allocated to the VM"
+    default     = "50"
+}
+
+variable "processors" {
+    description = "Number of virtual processors to allocate to the VM"
+    default     = "15"
+}
+
+# The s922 model is the cheapest model
```
dustymabe, line 40: optional: for systemd_type and proc_type it might be useful to add a hyperlink to a place where there is a description of the options.

multi-arch-builders/coreos-ppc64le-builder.bu
dustymabe: and also line 3 from `multi-arch-builders/coreos-aarch64-builder.bu` if you don't mind

PR #936, opened 2023-11-29
Revert "config: skip applehv on stable for next set of releases"

dustymabe:
This reverts commit 41121e394c82bc6af843eed3b72472ef8f1adb84. The new Ignition version is being promoted to stable now sow we can allow applehv to be built for stable.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -1,9 +1,6 @@
 streams:
   stable:
     type: production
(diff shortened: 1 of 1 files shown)


dustymabe:
After the next `stable` release we can revert this change. Marking as draft until we are ready.

PR #937, opened 2023-12-01
manifests: drop "aliased" `jenkins` and `jenkins-agent-base` imagestream tags

jlebon:
In the case where we don't need custom certs, we had the
`openshift/jenkins:2` imagestream feed into the `jenkins:latest`
imagestream, which then was the input to the S2I build.

And similarly, we had `openshift/jenkins-agent-base:latest` feed  into
`jenkins-agent-base:latest`.

In eb70bf7 ("manifests/jenkins-s2i: set `scheduled` on our
imagestreams"), we added `scheduled: true` for those tags thinking they would
keep those imagestreams up to date with the `openshift/` ones. But in
fact, `scheduled: true` only works when an imagestream tracks an image
in an external registry.

There is no way as far as I can tell to have an imagestream tag track
another imagestream tag. The closest is aliases, but those only work
for tags within the same imagestream, not different imagestreams. This is
by design (see https://github.com/openshift/origin/pull/13595).

So we'd have to add more code to periodically `oc tag` them to keep them
up to date, and that doesn't seem worth it.

Instead, cut it out entirely.

So now, in the non-cert case, we really just have our buildconfig output
to an imagestream and that's it.

The cert flow remains unchanged.

To deal with the agent image being *either* the
`openshift/jenkins-agent-base` one or our custom one with a cert, we
have to parameterize the `defaultImage` property of the Kubernetes
plugin in `jenkins.yaml`.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -548,6 +548,13 @@ using a development cluster, it normally isn't, and you can drop it. For
 the Fedora prod cluster, use `ocs-storagecluster-ceph-rbd` as shown
 above.
 
+If using an additional root CA certificate, then you will also need to
+specify the `AGENT_NAMESPACE` parameter to yours, e.g.:
+
+```
+  --param=AGENT_NAMESPACE=fedora-coreos-pipeline \
+```
+
 Now, create the Jenkins configmap:
 
 ```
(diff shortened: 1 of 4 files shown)


jlebon:
This is closer to what's currently deployed live in the cluster. I had to do this as part of the last Jenkins update to actually get newer images. We'll want to make sure we get this in before doing a nuke and redeploy.

dustymabe (approved):
LGTM

PR #938, opened 2023-12-01
build-arch: Add env var to enable the new powervs builder

ravanelli:
TODO: We must revert it once we get the new ppc64le builder
back on line
 This is only enabling the new builder to run without KVM,
since there is no support for it.

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -75,6 +75,7 @@ cosa_img = cosa_img ?: pipeutils.get_cosa_img(pipecfg, params.STREAM)
 def stream_info = pipecfg.streams[params.STREAM]
 
 def cosa_controller_img = stream_info.cosa_controller_img_hack ?: cosa_img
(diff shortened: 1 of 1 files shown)


ravanelli:
I'm still thinking if this hack should be added around here https://github.com/coreos/coreos-assembler/blob/main/cmd/remote-session.go#L136

jlebon:
We could name the pipecfg hack knob something even more generic, e.g. `ppc64le_no_kvm`, and use that to also drive the other hacks that'll be needed, like running less tests.

dustymabe:
To alleviate some of the pressure here we can just do this as a short term thing directly on the builder that doesn't support KVM:

```
cat > $HOME/.config/containers/containers.conf.d/cosa_no_kvm.conf <<'EOF'
[container]
env=["COSA_NO_KVM=1"]
EOF

[core@cosa-devsh ~]$ podman run -it --rm quay.io/fedora/fedora-minimal:latest
bash-5.2# env | grep COSA
COSA_NO_KVM=1
```

ravanelli:
Closing it, since we did a hack in powervs and than we enabled  the env var feature via coreos-assembler

PR #940, opened 2023-12-04
config: add temporary `hacks.ppc64le_kola_minimal` knob

jlebon:
This knob will be used in the RHCOS pipeline to avoid running heavy tests without KVM support on ppc64le.

Related: https://issues.redhat.com/browse/COS-2554

PR #941, opened 2023-12-04
jobs/build-arch: make `ppc64le_kola_minimal` specific to ppc64le

marmijo:
Originally, the logic for the `ppc64le_kola_minimal` hack trigged on all arches if the hack was enabled. Update the condition to only trigger when building for the ppc64le architecture.

See: https://github.com/coreos/fedora-coreos-pipeline/pull/940


jlebon (approved):
Thanks!

PR #942, opened 2023-12-05
jobs/build-arch: don't even run `iso-live-login` test with ppc64le_kola_minimal

jlebon:
Even that test takes too long and times out. We could increase the test
timeout but at this point, let's try to sprint to something that works.


aaradhak (approved):
LGTM

PR #943, opened 2023-12-05
jobs/build-arch: only run one ppc64le build at a time with ppc64le_kola_minimal

jlebon:
Our builder only has 32G of RAM. Combined with the fact that ppc64le VMs require a minimum of 4G of RAM, that doesn't leave much capacity for parallelism. We don't run tests, but multiple artifacts launch VMs.

Let's bring it down to 1 build at a time to make it more reliable. If the throughput is too slow, we'll bump to a beefier instance and increase parallelism here.


dustymabe:
FTR I think our FCOS builder also only has 32G of RAM

dustymabe (approved):

jlebon:
> FTR I think our FCOS builder also only has 32G of RAM

One big difference though is that in FCOS we don't build the extensions container or the legacy container (both done in VMs).

Actually, since we're full privs I think we could make those non-virtualized but meh... I don't think it's worth getting that working if this suffices.

PR #944, opened 2023-12-06
jobs/build-arch: increase timeout with ppc64le_kola_minimal

jlebon:
We're tightly fitting in the 4h timeout currently on the RHCOS ppc64le builder and sometimes timing out. Bump it up to 5h.

PR #946, opened 2023-12-06
jobs/build: increase timeout with ppc64le_kola_minimal

marmijo:
The build-arch timeout was increased to 5h with the ppc64le_kola_minimal hack. Increase the build job's timeout to 5h as well so it doesnt timeout when waiting for the multi-arch job to complete.

See: https://github.com/coreos/fedora-coreos-pipeline/pull/944


dustymabe (approved):
LGTM

jlebon (approved):

PR #948, opened 2023-12-19
jobs/build-arch: increase timeout (again)

dustymabe:
The build-arch job is for some reason timing out again at 4h. Let's bump to get through this round of releases and then try to dig in to see if we can speed things up again.


dustymabe:
Merging so I can start the `next` and `testing` runs.

PR #950, opened 2024-01-08
jobs/check-aws-region: run daily

mike-nguyen:
Instead of checking for changes in the SCM, lets trigger this job daily like the bump-lockfiles job.


dustymabe (approved):
LGTM

PR #951, opened 2024-01-17
jobs/release: emit a single `stream.release` fedmsg

jlebon:
That message supports specifying multiple arches now. So let's do that since it's just cleaner that way.


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/3698

dustymabe (approved):
LGTM

PR #952, opened 2024-01-22
support setting pipeline env vars from the config

dustymabe:

Relies on https://github.com/coreos/coreos-ci-lib/pull/157

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -141,9 +141,11 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
         // We set the session to time out after 5h. This essentially
         // performs garbage collection on the remote if we fail to clean up.
         pipeutils.withPodmanRemoteArchBuilder(arch: basearch) {
-        def session = shwrapCapture("""
-        cosa remote-session create --image ${cosa_img} --expiration ${timeout_mins}m --workdir ${env.WORKSPACE}
-        """)
+        def session = pipeutils.createCosaRemoteSession(
+            expiration: "${timeout_mins}m",
+            image: cosa_img,
+            workdir: WORKSPACE,
+        )
         withEnv(["COREOS_ASSEMBLER_REMOTE_SESSION=${session}"]) {
 
         // add any additional root CA cert before we do anything that fetches
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -142,7 +142,7 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
         // performs garbage collection on the remote if we fail to clean up.
         pipeutils.withPodmanRemoteArchBuilder(arch: basearch) {
         def session = shwrapCapture("""
-        cosa remote-session create --image ${cosa_img} --expiration 5h --workdir ${env.WORKSPACE}
+        cosa remote-session create --image ${cosa_img} --expiration ${timeout_mins}m --workdir ${env.WORKSPACE}
         """)
         withEnv(["COREOS_ASSEMBLER_REMOTE_SESSION=${session}"]) {
 
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -76,8 +76,8 @@ def stream_info = pipecfg.streams[params.STREAM]
 
 def cosa_controller_img = stream_info.cosa_controller_img_hack ?: cosa_img
 
-// Determine if we should use osbuild for image building
-def use_osbuild = pipeutils.get_use_osbuild_for_stream(pipecfg, params.STREAM)
+// Grab any environment variables we should set
+def container_env = pipeutils.get_env_vars_for_stream(pipecfg, params.STREAM)
 
 // If we are a mechanical stream then we can pin packages but we
 // don't maintain complete lockfiles so we can't build in strict mode.
@@ -113,6 +113,7 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
     cosaPod(cpu: "${ncpus}",
             memory: "${cosa_memory_request_mb}Mi",
             image: cosa_controller_img,
+            env: container_env,
             serviceAccount: "jenkins") {
     timeout(time: timeout_mins, unit: 'MINUTES') {
     try {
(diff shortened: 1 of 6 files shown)


dustymabe:
marking as draft.. i'd like to merge this myself so I can be around afterwards to clean up any fallout. I did do some minimalish testing in stage.

dustymabe:
hmm. not sure what that CI failure is trying to get me to do? rename to `makeCosaRemoteSession()`?

jlebon:
> hmm. not sure what that CI failure is trying to get me to do? rename to `makeCosaRemoteSession()`?

Yeah, that's annoying.

I found https://docs.codacy.com/faq/code-analysis/how-to-skip-an-analysis/ which seems like a way to work around this.

dustymabe:
> I found https://docs.codacy.com/faq/code-analysis/how-to-skip-an-analysis/ which seems like a way to work around this.

Thanks for the link. I wonder if that only skip it for this run and then new PRs will hit the problem? I'll dig in to see if I can figure it out.

dustymabe:
OK. I just decided to comply rather than waste time trying to figure this out. See the last commit.

ravanelli (approved):
LGTM, thanks for working in it!

jobs/build-arch.Jenkinsfile
```diff
@@ -141,9 +141,11 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
         // We set the session to time out after 5h. This essentially
         // performs garbage collection on the remote if we fail to clean up.
         pipeutils.withPodmanRemoteArchBuilder(arch: basearch) {
-        def session = shwrapCapture("""
-        cosa remote-session create --image ${cosa_img} --expiration ${timeout_mins}m --workdir ${env.WORKSPACE}
-        """)
+        session = pipeutils.createCosaRemoteSession(
```
jlebon: def

jobs/debug-pod.Jenkinsfile
```diff
@@ -72,9 +72,11 @@ cosaPod(cpu: "${ncpus}",
         // We set the session to time out after 4h. This essentially
         // performs garbage collection on the remote if we fail to clean up.
         pipeutils.withPodmanRemoteArchBuilder(arch: params.ARCH) {
-        def session = shwrapCapture("""
-        cosa remote-session create --image ${cosa_img} --expiration ${params.TIMEOUT}h --workdir ${env.WORKSPACE}
-        """)
+        session = pipeutils.createCosaRemoteSession(
```
jlebon: def

utils.groovy
```diff
@@ -217,6 +217,25 @@ def bump_builds_json(stream, buildid, arch, s3_stream_dir, acl) {
...
 
+// Create a COSA remote session, which is usually used to
+// build on a different architecture.
+//
+// Available parameters:
+//    expiration: string that represents a golang duration for how
+//                long the container should last (i.e. 4h, 30m)
+//    image: string that represents the container image to pull
+//    workdir: string that represents the in container working directory
+def createCosaRemoteSession(params = [:]) {
```
jlebon: Aside: this (and a bunch of other functions in this file) would be a good candidate to put in coreos-ci-lib.

utils.groovy
```diff
@@ -123,8 +123,22 @@ def get_source_config_ref_for_stream(pipecfg, stream) {
     }
 }
 
-def get_use_osbuild_for_stream(pipecfg, stream) {
-    return pipecfg.streams[stream].use_osbuild ? '1' : ''
+def get_env_vars_for_stream(pipecfg, stream) {
+    env = [:]
```
jlebon, line 127: def

utils.groovy
```diff
@@ -123,8 +123,22 @@ def get_source_config_ref_for_stream(pipecfg, stream) {
...
+    // Grab the globally configured env vars first
+    if (pipecfg.env) {
+        for (x in pipecfg.env) {
+            env[x.key] = x.value
+        }
```
jlebon, line 132: Isn't that just

```groovy
env.putAll(pipecfg.env)
```

?

Similarly below.

utils.groovy
```diff
@@ -221,17 +235,23 @@ def bump_builds_json(stream, buildid, arch, s3_stream_dir, acl) {
...
 //    workdir: string that represents the in container working directory
 def createCosaRemoteSession(params = [:]) {
+    def env = params['env']
     def expiration = params['expiration']
     def image = params['image']
     def workdir = params['workdir']
+    def envArgs = ""
+    for (x in env) {
+        envArgs += "--env ${x.key}=${x.value} "
+    }
```
jlebon, line 251: What we do elsewhere is e.g.

```
envArgs = env.collect { k, v -> "--env ${k}=${v}" }
```

and then `${envArgs.join(" ")}` below.

utils.groovy
```diff
@@ -221,17 +235,23 @@ def bump_builds_json(stream, buildid, arch, s3_stream_dir, acl) {
...
 //                long the container should last (i.e. 4h, 30m)
 //    image: string that represents the container image to pull
 //    workdir: string that represents the in container working directory
 def createCosaRemoteSession(params = [:]) {
+    def env = params['env']
     def expiration = params['expiration']
     def image = params['image']
     def workdir = params['workdir']
+    def envArgs = ""
+    for (x in env) {
```
jlebon, line 249: This assumes `env` is non-null, no?

I think we need something like `def env = params['env'] ?: [:]` above, like we do elsewhere.

utils.groovy
dustymabe: maybe :) - I use groovy about once every 3-6 months so I always end up dropping back to languange non-specific primitives like a for loop.

I'll have to test to make sure `putAll` gives us what we want here.

utils.groovy
dustymabe: will do

utils.groovy
dustymabe: yep - for another day

utils.groovy
```diff
@@ -217,6 +227,28 @@ def bump_builds_json(stream, buildid, arch, s3_stream_dir, acl) {
...
+// build on a different architecture.
+//
+// Available parameters:
+//    env: (optional) map of key/val pairs of environment variables to set
+//    expiration: string that represents a golang duration for how
+//                long the container should last (i.e. 4h, 30m)
+//    image: string that represents the container image to pull
+//    workdir: string that represents the in container working directory
+def makeCosaRemoteSession(params = [:]) {
+    def env = params['env'] ?: [:]
```
jlebon, line 240: Though note this is shadowing the builtin `env` var. I don't think it matters in this case, but would be cleaner to not shadow.

jlebon (approved):

utils.groovy
dustymabe: fixed

jlebon (approved):

utils.groovy
dustymabe: reworked things to not shadow the `env` var

PR #953, opened 2024-01-23
multi-arch-builders: Disable Zincati

ravanelli:
- PPC64LE has issues in Petitboot with auto updates, due it disable it.
 See: https://github.com/coreos/fedora-coreos-tracker/issues/1633

### DIFF
--- a/multi-arch-builders/coreos-ppc64le-builder.bu
+++ b/multi-arch-builders/coreos-ppc64le-builder.bu
@@ -23,3 +23,7 @@ storage:
       overwrite: true
       contents:
         inline: coreos-ppc64le-builder
+systemd:
+  units:
+    - name: zincati.service
+      enabled: false
--- a/multi-arch-builders/coreos-ppc64le-rhcos-builder.bu
+++ b/multi-arch-builders/coreos-ppc64le-rhcos-builder.bu
@@ -10,3 +10,7 @@ ignition:
     merge:
       - local: coreos-ppc64le-builder.ign
       - local: builder-splunk.ign
+systemd:
+  units:
+    - name: zincati.service
+      enabled: false


marmijo (approved):
LGTM

ravanelli:
Maybe we should also add it for the fcos butane.

multi-arch-builders/coreos-ppc64le-builder.bu
```diff
@@ -23,3 +23,7 @@ storage:
       overwrite: true
       contents:
         inline: coreos-ppc64le-builder
+systemd:
+  units:
+    - name: zincati.service
```
jlebon, line 28: Can we add a comment linking to https://github.com/coreos/fedora-coreos-tracker/issues/1633 here? And similarly below.

jlebon (approved):

jlebon:
Is this an issue on both the RHCOS ppc64le builder and the FCOS one?

dustymabe:
> Is this an issue on both the RHCOS ppc64le builder and the FCOS one?

The FCOS builder has been updating fine IIUC:

```
core@buildvm-ppc64le-fcos01:~$ rpm-ostree status 
State: idle
AutomaticUpdatesDriver: Zincati
  DriverState: active; periodically polling for updates (last checked Tue 2024-01-23 15:59:43 UTC)
Deployments:
● fedora:fedora/ppc64le/coreos/testing
                  Version: 39.20240112.2.0 (2024-01-15T19:38:07Z)
                   Commit: 133b828e8a8b3dd24d51b4a17b479b23cbebe6ade033e6e791cdfd40cfae817f
             GPGSignature: Valid signature by E8F23996F23218640CB44CBE75CF5AC418B8E74C

  fedora:fedora/ppc64le/coreos/testing
                  Version: 39.20240104.2.0 (2024-01-06T02:39:29Z)
                   Commit: 6643c1a85f4a3a5787487efd665d2ac0177efca5fac288850faaef0eeb927e99
             GPGSignature: Valid signature by E8F23996F23218640CB44CBE75CF5AC418B8E74C
core@buildvm-ppc64le-fcos01:~$ uptime
 16:00:27 up 1 day,  9:00,  1 user,  load average: 0.07, 0.26, 0.67
```

ravanelli:
> > Is this an issue on both the RHCOS ppc64le builder and the FCOS one?
> 
> The FCOS builder has been updating fine IIUC:
> 
> ```
> core@buildvm-ppc64le-fcos01:~$ rpm-ostree status 
> State: idle
> AutomaticUpdatesDriver: Zincati
>   DriverState: active; periodically polling for updates (last checked Tue 2024-01-23 15:59:43 UTC)
> Deployments:
> ● fedora:fedora/ppc64le/coreos/testing
>                   Version: 39.20240112.2.0 (2024-01-15T19:38:07Z)
>                    Commit: 133b828e8a8b3dd24d51b4a17b479b23cbebe6ade033e6e791cdfd40cfae817f
>              GPGSignature: Valid signature by E8F23996F23218640CB44CBE75CF5AC418B8E74C
> 
>   fedora:fedora/ppc64le/coreos/testing
>                   Version: 39.20240104.2.0 (2024-01-06T02:39:29Z)
>                    Commit: 6643c1a85f4a3a5787487efd665d2ac0177efca5fac288850faaef0eeb927e99
>              GPGSignature: Valid signature by E8F23996F23218640CB44CBE75CF5AC418B8E74C
> core@buildvm-ppc64le-fcos01:~$ uptime
>  16:00:27 up 1 day,  9:00,  1 user,  load average: 0.07, 0.26, 0.67
> ```

Do you know the server model/firmware version for fcos?

ravanelli:
> I'm not super comfortable with making this change. It seems almost like we have given up trying to solve the problem. @ravanelli has valiantly been fighting this fight and I appreciate that.
> 
> Maybe someone else needs to pick it up and push forward for a solution. Either there is a bug in our software stack (which would affect customers) or the piece of hardware/firmware we are using are faulty, in which case we need to push on RH/IBM to provide us something that works.
> 
> Until we can figure out the problem what we will have is drift on our build servers where some build servers are on newer versions than others, which isn't good. I'd say until this problem is solved we need to periodically schedule manual reprovisioning of the server to keep it in line with the latest updates.

I think it is a matter of the firmware, as in a similar issue we had in the past. Our rhcos build firmware is updated but still with this issue.

dustymabe:
> It's a VM

This could be a strategy for us too.. Just install RHEL on the host that we own and then run an FCOS VM that will be our builder. I don't love it, because then we have to periodically update RHEL too (which may have the same "can't reboot" problem), but at least would allow us to update the FCOS VM without interruption.

jlebon:
> I'd say until this problem is solved we need to periodically schedule manual reprovisioning of the server to keep it in line with the latest updates.

I definitely agree, but even if we did periodic reprovisioning, we'd still want this PR. The VM idea sounds interesting.

ravanelli:
> > It's a VM
> 
> This could be a strategy for us too.. Just install RHEL on the host that we own and then run an FCOS VM that will be our builder. I don't love it, because then we have to periodically update RHEL too (which may have the same "can't reboot" problem), but at least would allow us to update the FCOS VM without interruption.

RHEL8 has no official support for qemu/kvm in ppc64le neither, as it dropped kvm support at all in RHEL9. We would need to use Fedora as the host.

ravanelli:
Closing  it, since we agreed we will continue to do it manually for our rhcos builder. I created some internal docs to remember us to do it.

PR #954, opened 2024-01-23
plugins: add `matrix-communications`

jlebon:
I'd like to have `bodhi-trigger` in CoreOS CI send messages to Matrix when failures occur. If that works well, we could also in the future move pipeline notifications to Matrix as well.


dustymabe (approved):
LGTM

PR #959, opened 2024-02-13
Add a new Jenkins job to build containers for kola

jbtrystram:
Add a new Jenkins job to build containers for kola
    
    Some kola tests require external images. This job will build
    and publish the images present under the test/containers path
    of coreos-assembler.
    
    Inital PR for coreos-assembler: https://github.com/coreos/coreos-assembler/pull/3727
    
    x-ref https://github.com/coreos/fedora-coreos-tracker/issues/1639

Once https://github.com/coreos/coreos-assembler/pull/3727 is merged i'll need to create the repos in quay and update the default values, before merging this one.

### DIFF
--- a/jobs/build-kola-containers.Jenkinsfile
+++ b/jobs/build-kola-containers.Jenkinsfile
@@ -1,0 +1,225 @@
+def gitref, commit, shortcommit, contexts, changeset
+node {
+    checkout scm
+    // these are script global vars
+    pipeutils = load("utils.groovy")
+}
+
+properties([
+    pipelineTriggers([
+        [$class: 'GenericTrigger',
+         genericVariables: [
+          [
+           key: 'COREOS_ASSEMBLER_GIT_REF',
+           value: '$.ref',
+           expressionType: 'JSONPath',
+           regexpFilter: 'refs/heads/', //Optional, defaults to empty string
+           defaultValue: ''  //Optional, defaults to empty string
+          ]
+         ],
+         causeString: 'Triggered on $ref',
+         token: 'build-kola-test-containers',
+         tokenCredentialId: '',
+         printContributedVariables: true,
+         printPostContent: true,
+         silentResponse: false,
+         regexpFilterText: '$COREOS_ASSEMBLER_GIT_REF',
+         regexpFilterExpression: 'main'
+        ]
+    ]),
+    parameters([
+      string(name: 'ARCHES',
+             description: 'Space-separated list of target architectures',
+             defaultValue: "x86_64" + " " + pipeutils.get_supported_additional_arches().join(" "),
+             trim: true),
+      string(name: 'COREOS_ASSEMBLER_GIT_URL',
+             description: 'Override the coreos-assembler git repo to use',
+             defaultValue: "https://github.com/coreos/coreos-assembler.git",
+             trim: true),
+      string(name: 'COREOS_ASSEMBLER_GIT_REF',
+             description: 'Override the coreos-assembler git ref to use',
+             defaultValue: "main",
+             trim: true),
+      string(name: 'CONTAINER_REGISTRY_ORG',
+             description: 'Override the registry org to push the containers to',
+             defaultValue: "quay.io/coreos-assembler",
+             trim: true),
+      string(name: 'CONTAINER_REGISTRY_STAGING_REPO',
+             description: 'Override the staging registry where intermediate images go',
+             defaultValue: "quay.io/coreos-assembler/staging",
+             trim: true),
+      string(name: 'PATH_TO_CONTEXTS',
+             description: """Override the path to the contexts directories to use as build contexts.
+             Each directory should contain a Containerfile.
+             The image will be named after the directory name.""",
+             defaultValue: "tests/containers/",
+             trim: true),
+      string(name: 'COREOS_ASSEMBLER_IMAGE',
+             description: 'Override the coreos-assembler image to use',
+             defaultValue: "quay.io/coreos-assembler/coreos-assembler:main",
+             trim: true),
+      booleanParam(name: 'FORCE',
+             defaultValue: false,
+             description: 'Whether to force a rebuild'),
+    ]),
+    buildDiscarder(logRotator(
+        numToKeepStr: '100',
+        artifactNumToKeepStr: '100'
+    )),
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+])
+
+node {
+    change = checkout(
+        changelog: true,
+        poll: false,
+        scm: [
+            $class: 'GitSCM',
+            branches: [[name: "origin/${params.COREOS_ASSEMBLER_GIT_REF}"]],
+            userRemoteConfigs: [[url: params.COREOS_ASSEMBLER_GIT_URL]],
+            extensions: [[$class: 'CloneOption',
+                          noTags: true,
+                          reference: '',
+                          shallow: false]]
+        ]
+    )
+
+    gitref = params.COREOS_ASSEMBLER_GIT_REF
+    def output = shwrapCapture("git rev-parse HEAD")
+    commit = output.substring(0,40)
+    shortcommit = commit.substring(0,7)
+
+    // gather the context folders list
+    def path = params.PATH_TO_CONTEXTS
+    contexts = shwrapCapture("""
+        cd ${path}
+        find . -maxdepth 1 -mindepth 1 -type d -exec basename {} \\;
+    """).trim().split("\n")
+
+    if ( !params.FORCE ) {
+        // get the git commit ref for the last built container
+        def previous_ref = shWrapCapture("""
+            skopeo inspect --no-creds docker://${params.CONTAINER_REGISTRY_ORG}/${contexts[0]}:latest \
+            | jq -r '.Labels.org.opencontainers.image.revision'
+        """)
+
+        // Check for changes in tests/containers/* since the last build
+        // If none, no need to run this
+        changeset = shwrapRc("git diff --quiet ${previous_ref} -- ${path}")
+    } else {
+        changeset = 1
+  }
+}
+
+if ( changeset == 0 ) {
+    currentBuild.result = 'SUCCESS'
+    currentBuild.description = "[${gitref}@${shortcommit}] 💤 (no change)."
+    return
+} else if ( changeset != 1 ) {
+    currentBuild.result = 'FAILURE'
+    currentBuild.description = "[${gitref}@${shortcommit}] ❌ Cannot determine changes. git diff return code ${changeset}."
+    message = "build-kola-test-containers #${env.BUILD_NUMBER} <${env.BUILD_URL}|:jenkins:> <${env.RUN_DISPLAY_URL}|:ocean:> [${gitref}@${shortcommit}]"
+    pipeutils.trySlackSend(message: message)
+    return
+}
+
+currentBuild.description = "[${gitref}@${shortcommit}] Waiting"
+
+// Get the list of requested architectures to build for
+def basearches = params.ARCHES.split() as Set
+// and  the list of images to build
+def imageNames = contexts as Set
+
+lock(resource: "build-kola-containers") {
+    cosaPod(image: params.COREOS_ASSEMBLER_IMAGE,
+            memory: "512Mi", kvm: false,
+            serviceAccount: "jenkins") {
+    timeout(time: 60, unit: 'MINUTES') {
+    try {
+
+        currentBuild.description = "[${gitref}@${shortcommit}] Running"
+
+        // By default we will allow re-using cache layers for one day.
+        // This is mostly so we can prevent re-downloading the RPMS
+        // and repo metadata and over again in a given day for successive
+        // builds.
+        def cacheTTL = "24h"
+        def force = ""
+        if (params.FORCE) {
+            force = '--force'
+            // Also set cacheTTL to 0.1s to allow users an escape hatch
+            // to force no cache layer usage.
+            cacheTTL = "0.1s"
+        }
+
+        withCredentials([file(credentialsId: 'cosa-push-registry-secret', variable: 'REGISTRY_SECRET')]) {
+            stage('Build Containers') {
+                parallel basearches.collectEntries{arch -> [arch, {
+                    for (imageName in imageNames) {
+                        def dirname = "${params.PATH_TO_CONTEXTS}/${imageName}"
+                        pipeutils.withPodmanRemoteArchBuilder(arch: arch) {
+                            // Forcing the tag to include the image name
+                            // prevents racing between arches and overwrite the images
+                            shwrap("""
+                                cosa remote-build-container \
+                                    --git-sub-dir ${dirname}\
+                                    --arch ${arch} --cache-ttl ${cacheTTL} \
+                                    --git-ref ${commit} ${force} \
+                                    --git-url ${params.COREOS_ASSEMBLER_GIT_URL} \
+                                    --tag ${imageName}-${arch}-${shortcommit} \
+                                    --repo ${params.CONTAINER_REGISTRY_STAGING_REPO} \
+                                    --push-to-registry --auth=\$REGISTRY_SECRET
+                            """)
+                        }
+                    }
+                }]}
+            }
+
+            stage('Push Manifests') {
+                for (imageName in imageNames) {
+                    def images = ""
+                    for (arch in basearches) {
+                        images += " --image=docker://${params.CONTAINER_REGISTRY_STAGING_REPO}:${imageName}-${arch}-${shortcommit}"
+                    }
+
+                    shwrap("""
+                        export STORAGE_DRIVER=vfs # https://github.com/coreos/fedora-coreos-pipeline/issues/723#issuecomment-1297668507
+                        cosa push-container-manifest --v2s2 \
+                            --auth=\$REGISTRY_SECRET --tag latest \
+                            --repo ${params.CONTAINER_REGISTRY_ORG}/${imageName} ${images}
+                    """)
+                }
+            }
+
+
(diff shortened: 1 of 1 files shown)


jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,224 @@
...
+            // Also set cacheTTL to 0.1s to allow users an escape hatch
+            // to force no cache layer usage.
+            cacheTTL = "0.1s"
+        }
+
+        withCredentials([file(credentialsId: 'jbtrystram-testorg-robot-push-secret', variable: 'REGISTRY_SECRET')]) {
+            stage('Build Containers') {
+                parallel basearches.collectEntries{arch -> [arch, {
+                    for (directory in contexts) {
+                        def dir = params.PATH_TO_CONTEXTS + directory
```
jlebon, line 147: Let's use e.g. `dirname` here; `dir` is also the name of a step.

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,224 @@
...
+                                    --git-url ${params.COREOS_ASSEMBLER_GIT_URL} \
+                                    --repo ${params.CONTAINER_REGISTRY_ORG}/$directory \
+                                    --push-to-registry --auth=\$REGISTRY_SECRET
+                            """)
+                        }
+                }
+                }]}
+            }
+
+            stage('Push manifests') {
```
jlebon, line 163: Super minor, but for consistency

```suggestion
            stage('Push Manifests') {
```

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,224 @@
...
+                for (architecture in basearches) {
+                    def arch = architecture
+                    for (directory in contexts) {
+                        def dir = directory
```
jlebon, line 167: Minor: that seems unnecessary for a two line block. I guess you're trying to be consistent in all the for loops. Maybe let's choose a different name instead for the iteration variable that works well as is.

Edit: OK see other comment about renaming this to `imageName`.

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,224 @@
...
+        find . -maxdepth 1 -mindepth 1 -type d -not -path '*/\\.*' -exec basename {} \\;
+    """).trim().split("\n")
+}
+
+currentBuild.description = "[${gitref}@${shortcommit}] Waiting"
+
+// Get the list of requested architectures to build for
+def basearches = params.ARCHES.split() as Set
+// and  the list of contexts to build
+def contexts = contexts as Set
```
jlebon, line 119: This is being defined before it's set.

jlebon: Also, I think this would probably be clearer named as e.g. `imageNames`? And then the for loops below would be `for (imageName in imageNames)`? We use it more in the context of image naming than the context of directory naming overall.

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,224 @@
...
+          ]
+         ],
+         causeString: 'Triggered on $ref',
+         token: 'build-kola-test-containers',
+         tokenCredentialId: '',
+         printContributedVariables: true,
+         printPostContent: true,
+         silentResponse: false,
+         regexpFilterText: '$COREOS_ASSEMBLER_GIT_REF',
+         regexpFilterExpression: 'main|rhcos-.*'
```
jlebon, line 27: This will trigger for older RHCOS branches too which don't have the iSCSI tests and don't need to have them built.

Honestly, kinda thinking for now we just build for `main` and even when we branch for e.g. 4.16, those tests can keep using the `:main` tag of the container image. If we feel we need to branch the images too, we can do that then. But it's worth mentioning it in their READMEs in cosa.

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,224 @@
...
+                    for (directory in contexts) {
+                        def dir = params.PATH_TO_CONTEXTS + directory
+                        pipeutils.withPodmanRemoteArchBuilder(arch: arch) {
+                            shwrap("""
+                                cosa remote-build-container \
+                                    --git-sub-dir $dir\
+                                    --arch $arch --cache-ttl ${cacheTTL} \
+                                    --git-ref $commit ${force} \
+                                    --git-url ${params.COREOS_ASSEMBLER_GIT_URL} \
+                                    --repo ${params.CONTAINER_REGISTRY_ORG}/$directory \
```
jlebon, line 155: How come we don't use the staging repo?

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,224 @@
...
+    commit = output.substring(0,40)
+    shortcommit = commit.substring(0,7)
+}
+
+node {
+    def path=params.PATH_TO_CONTEXTS
+    // gather the context folders list
+    contexts = sh(returnStdout: true, script: """
+        cd ${path}
+        find . -maxdepth 1 -mindepth 1 -type d -not -path '*/\\.*' -exec basename {} \\;
```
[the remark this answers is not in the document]
jbtrystram: avoid directories leading with `.`

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,224 @@
...
+                          reference: '',
+                          shallow: true]]
+        ]
+    )
+
+    gitref = params.COREOS_ASSEMBLER_GIT_REF
+    def output = shwrapCapture("git rev-parse HEAD")
+    commit = output.substring(0,40)
+    shortcommit = commit.substring(0,7)
+}
```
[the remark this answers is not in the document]
jbtrystram: yep, looks like a copy pasta ! Thanks for catching it

jobs/build-kola-containers.Jenkinsfile
jbtrystram: these image will probably have very little change, other than the occasionnal base image bump, so I agree sticking to one main image is fine, until we find out we need to change that

jobs/build-kola-containers.Jenkinsfile
jbtrystram: the staging repo seems to be for the coreos-assembler image , from what i understand. 
That would not be a good fit there, would it ? Do you want to publish the images to staging first ? If so, we'd need another job that test them before promoting them. That sounds like a complicated process for images that are very simple and would be updated very rarely

jobs/build-kola-containers.Jenkinsfile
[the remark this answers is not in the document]
jlebon: Hmm, I don't see any:

```
$ cd tests/containers
$ find . -maxdepth 1 -mindepth 1 -type d -exec basename {} \;
tang
targetcli
```

jobs/build-kola-containers.Jenkinsfile
jlebon: It's used as an intermediate place to assemble manifests (multi-arch images). It doesn't need to be restricted to cosa (the `fcos-buildroot` image also uses it). No need for another job. You can follow the same pattern in the original cosa and fcos-buildroot jobs where we first push to the intermediate repo, then push the manifest to the final repo, and then delete the tags from the intermediate repo.

jobs/build-kola-containers.Jenkinsfile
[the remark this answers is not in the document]
jbtrystram: it was more a safety measure but you are right, it's a bit superfluous

jobs/build-kola-containers.Jenkinsfile
jbtrystram: I was going to ask why can't we assemble the manifest locally and push it once ?
But we are using real-world different  hosts for each arch, each push to the intermediate repo, then we assemble. Gotcha. Will update !

jbtrystram:
I updated this PR with a few changes following @jlebon review : 
- Only publish "latest" tagged manifest ( as we decided to only run on `main`) 
- Abort early (but with success) if no changes were made in $PATH_TO_CONTEXTS (see L89-108)
- Add a FORCE flag to force a rebuild
-  Publishing intermediate images to the staging repo 
- Addressing the other review comments

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,209 @@
...
+                          shallow: true]]
+        ]
+    )
+
+    gitref = params.COREOS_ASSEMBLER_GIT_REF
+    def output = shwrapCapture("git rev-parse HEAD")
+    commit = output.substring(0,40)
+    shortcommit = commit.substring(0,7)
+
+    def path=params.PATH_TO_CONTEXTS
```
jlebon, line 88:
```suggestion
    def path = params.PATH_TO_CONTEXTS
```

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,209 @@
...
+    // Check for changes in tests/containers/*
+    // If not, no need to run this
+    // not using shWrapCapture here because i don't want -euxo pipefail to be set, as grep returns 1 on non-match
+    def changeset = sh(returnStatus: true, script:"git diff-tree -r --name-only HEAD | grep ${path}")
```
jlebon, line 92: This is only checking the changed files in the latest commit, but multiple commits may have been merged, right?

I like the idea, but I think the way to do this is probably to `skopeo inspect` the latest built cosa image and get the git commit from the `org.opencontainers.image.revision` label and compare against that.

jlebon: A simpler invocation is `git diff --quiet ${base}^ -- ${path}`.

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    // gather the context folders list
+    contexts = shwrapCapture("""
+        cd ${path}
+        find . -maxdepth 1 -mindepth 1 -type d -exec basename {} \\;
+    """).trim().split("\n")
+}
+
+if ( !changes ) {
```
jlebon, line 104: I don't see anything setting `changes` to true.

But anyway, it doesn't seem like we need it? We could return when we detect the changeset doesn't have any relevant changes.

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,209 @@
...
+
+currentBuild.description = "[${gitref}@${shortcommit}] Waiting"
+
+// Get the list of requested architectures to build for
+def basearches = params.ARCHES.split() as Set
+// and  the list of images to build
+def imageNames = contexts as Set
+
+lock(resource: "build-containers") {
+    cosaPod(image: "quay.io/coreos-assembler/coreos-assembler",
```
jlebon, line 118: That's the default image.

Minor: I would just keep the COREOS_ASSEMBLER_IMAGE parameter for consistency with other jobs of this kind. It can come in handy in the future.

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,209 @@
...
+                        def dirname = params.PATH_TO_CONTEXTS + imageName
+                        pipeutils.withPodmanRemoteArchBuilder(arch: arch) {
+                            shwrap("""
+                                cosa remote-build-container \
+                                    --git-sub-dir $dirname\
+                                    --arch $arch --cache-ttl ${cacheTTL} \
+                                    --git-ref $commit ${force} \
+                                    --git-url ${params.COREOS_ASSEMBLER_GIT_URL} \
+                                    // Force the tag to inclue the image name to avoid racing between aarches
+                                    // and overwrite the images
```
jlebon, line 152:
```suggestion
                                    // Force the tag to include the image name to avoid racing between arches
                                    // and overwrite the images
```

That's neat. I guess those lines correctly get parsed as Groovy comments because it happens before the parsing of the multi-line string. My brain keeps reading it as invalid.

jobs/build-kola-containers.Jenkinsfile
jbtrystram: good point ! I'll look into that

jobs/build-kola-containers.Jenkinsfile
jbtrystram: > I don't see anything setting changes to true.
/me facepalms. 
Also 
```if foo {
	bar=false
}
``` 
 I just realized I did not caught that. It hurts a little bit :sob:

jobs/build-kola-containers.Jenkinsfile
jbtrystram: No you are right, it does not work :) Fixed

jobs/build-kola-containers.Jenkinsfile
jbtrystram: @jlebon this git command works fine indeed ! Why do you point to the parent of $base though ? Just curious :)

jlebon: Good catch, yes. That should just be `git diff --quiet ${base} -- ${path}`.

jbtrystram: ah ok :)

jbtrystram:
@jlebon I just squashed all the commit and pushed an update. 
One change i made last is that if FORCE is set, then the whole git diff thing is not even tried. That will allow the initial run (where skopeo would have failed because no inital to look for the commit reference)

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,221 @@
...
+        def changeset = sh(returnStatus: true, script:"git diff --quiet ${previous_ref} -- ${path}")
+        changes = ( changeset == 1 )
+    } else {
+    changes = true
+  }
+}
+
+if ( !changes ) {
+    currentBuild.result = 'SUCCESS'
+    currentBuild.description = "[${gitref}@${shortcommit}] No changes in ${path}. Skiped the build."
```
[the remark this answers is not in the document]
jlebon: If we want to be consistent with the `build` job:

```suggestion
    currentBuild.description = "[${gitref}@${shortcommit}] 💤 (no change)"
```

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,221 @@
...
+}
+
+currentBuild.description = "[${gitref}@${shortcommit}] Waiting"
+
+// Get the list of requested architectures to build for
+def basearches = params.ARCHES.split() as Set
+// and  the list of images to build
+def imageNames = contexts as Set
+
+lock(resource: "build-containers") {
```
jlebon, line 129: Maybe `build-kola-containers` instead?

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,221 @@
...
+            // Also set cacheTTL to 0.1s to allow users an escape hatch
+            // to force no cache layer usage.
+            cacheTTL = "0.1s"
+        }
+
+        withCredentials([file(credentialsId: 'cosa-push-registry-secret', variable: 'REGISTRY_SECRET')]) {
+            stage('Build Containers') {
+                parallel basearches.collectEntries{arch -> [arch, {
+                    for (imageName in imageNames) {
+                        def dirname = params.PATH_TO_CONTEXTS + imageName
```
jlebon, line 155:
```suggestion
                        def dirname = "${params.PATH_TO_CONTEXTS}/${imageName}"
```

So that we don't rely on the parameter always having a trailing slash.

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,221 @@
...
+        // get the git commit ref for the last built container
+        def previous_ref = shWrapCapture("""
+            skopeo inspect --no-creds docker://${params.CONTAINER_REGISTRY_ORG}/${contexts[0]}:latest \
+            | jq -r '.Labels.org.opencontainers.image.revision'
+        """)
+
+        // Check for changes in tests/containers/* since the last build
+        // If none, no need to run this
+        // not using shWrapCapture here because we want the exit code 
+        def changeset = sh(returnStatus: true, script:"git diff --quiet ${previous_ref} -- ${path}")
```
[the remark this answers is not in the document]
jbtrystram: could I just clone with `shallow: false ` ?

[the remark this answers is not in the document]
jlebon: That should do it, yeah.

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,221 @@
...
+        changes = ( changeset == 1 )
```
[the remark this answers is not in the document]
jbtrystram: I added an if..else statement to handle this and return the exit code if it's neither 1 or 0

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -0,0 +1,223 @@
...
+}
+
+if ( changeset == 0 ) {
+    currentBuild.result = 'SUCCESS'
+    currentBuild.description = "[${gitref}@${shortcommit}] 💤 (no change)."
+    return
+} else if ( changeset != 1 ) {
+    currentBuild.result = 'FAILURE'
+    currentBuild.description = "[${gitref}@${shortcommit}] ❌ Cannot determine changes. git diff return code ${changeset}."
+    return
```
jlebon, line 121: Since this will skip the Slack messaging, we should probably manually call `pipeutils.trySlackSend()` here.

jlebon (approved):
One final comment, otherwise LGTM!

jlebon (approved):

PR #961, opened 2024-02-16
config: promote `versionary` key, rework it

jlebon:
The versionary FCOS script was one of the few things that differed between FCOS and RHCOS and was awkwardly attached as a hack knob to work.

Now, we promote it to a proper key under the `misc` section, with support for stream-level overrides. This is prep for also using a versionary in other streams.

The versionary is now part of the source config and cosa knows how to execute it if the `--versionary` argument is passed.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -37,9 +37,6 @@ hacks:
   # OPTIONAL: skip UEFI on older RHCOS
   skip_uefi_tests_on_older_rhcos: true
 
-# OPTIONAL/TEMPORARY: whether to use `versionary` to derive version numbers
-versionary_hack: true
-
 # OPTIONAL: coreos-assembler image to build with. Supports ${STREAM} variable.
 # If unset, defaults to `quay.io/coreos-assembler/coreos-assembler:main`.
 cosa_img: quay.io/jlebon/coreos-assembler:pr3139-${STREAM}
@@ -60,6 +57,8 @@ streams:
       source_config_ref: main
       # OPTIONAL: override OS variant to use for this stream
       variant: rhcos-9.0
+      # OPTIONAL: override whether to use a versionary for this stream
+      versionary: false
     stable:
       type: production
       # OPTIONAL: override cosa image to use for this stream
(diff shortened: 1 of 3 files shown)


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/3735
Requires: https://github.com/coreos/fedora-coreos-config/pull/2859

docs/config.yaml
```diff
@@ -270,6 +269,8 @@ clouds:
 
 # OPTIONAL: miscellaneous options
 misc:
+  # OPTIONAL: whether to use a versionary to derive version numbers
```
dustymabe, line 272: Is a `versionary` enough of a noun? 

```suggestion
  # OPTIONAL: whether to use a versionary script to derive version numbers
```

jobs/build.Jenkinsfile
```diff
@@ -274,10 +267,22 @@ lock(resource: "build-${params.STREAM}") {
...
-            def version = new_version ? "--version ${new_version}" : ""
+            def version_arg = ""
+            if (params.VERSION) {
+                version_arg = "--version ${params.VERSION}"
+            } else {
+                def use_versionary = pipecfg.misc?.versionary
+                if (stream_info.containsKey('versionary')) {
+                    // stream override always wins
+                    use_versionary = stream_info.versionary
+                }
```
dustymabe, line 278: i'm guessing if we used this in one other place we'd put a helper function in `utils.groovy`?

docs/config.yaml
jlebon: I'm trying to make it become one. :) But fine to make it less cool.

jobs/build.Jenkinsfile
jlebon: Yeah, agree. I don't anticipate that need for now.

jlebon:
This needs to wait until https://github.com/coreos/fedora-coreos-config/pull/2859 has made it to stable.

dustymabe:
> This needs to wait until [coreos/fedora-coreos-config#2859](https://github.com/coreos/fedora-coreos-config/pull/2859) has made it to stable.

or just PR the `testing` branch and add it so it will be there for the next cycle.

jlebon:
> > This needs to wait until [coreos/fedora-coreos-config#2859](https://github.com/coreos/fedora-coreos-config/pull/2859) has made it to stable.
> 
> or just PR the `testing` branch and add it so it will be there for the next cycle.

WFM. https://github.com/coreos/fedora-coreos-config/pull/2863

jlebon:
The `stable` branch now has the versionary script!

PR #966, opened 2024-02-26
config: use OSBuild for the branched stream too

dustymabe:
We're phasing this in and it will be enabled for `next` soon enough so let's enable it for `branched` now.


c4rt0 (approved):
/LGTM

PR #967, opened 2024-02-27
config: enable next-devel; use OSBuild

dustymabe:


dustymabe:
opening this PR early so we are reminded to switch to using OSBuild when we do enable `next-devel`.

jlebon (approved):

marmijo (approved):

PR #968, opened 2024-03-07
Add podman-os build job

ravanelli:
- Integrate essential Podman build jobs into our CI
infrastructure to facilitate unblocking tasks for
the Podman team. The objective is to ensure
flexibility by avoiding strong dependencies on our
internal components such as cosa and libraries,
enabling seamless transition for the Podman team in the future.

### DIFF
--- a/jobs/build-podman-os.Jenkinsfile
+++ b/jobs/build-podman-os.Jenkinsfile
@@ -1,0 +1,349 @@
+def gitref, commit, shortcommit
+
+def artifacts = [
+    [platform: "qemu", suffix: "qcow2"],
+    [platform: "applehv", suffix: "raw"],
+    [platform: "hyperv", suffix: "vhdx"]
+]
+
+node {
+    checkout scm
+    // these are script global vars
+    pipeutils = load("utils.groovy")
+}
+
+// Keep it here for now instead of utils
+// since it is podman job only
+
+def generate_disk_images(artifacts, shortcommit, arch, staging_repo, repo) {
+    generate_diskvar_json(shortcommit, arch, artifacts, staging_repo, repo)
+    // First create the qcow2 and store it in cache
+    supermin_run("--cache", arch, "qcow2", shortcommit)
+    def artifacts_runvm = artifacts.drop(1) // drop qcow2
+    parallel artifacts_runvm.collectEntries{artifact -> [artifact["platform"], {
+        supermin_run("--snapshot", arch, "${artifact["suffix"]}", shortcommit)
+    }]}
+}
+
+
+def generate_diskvar_json(shortcommit, arch, artifacts, staging_repo, repo) {
+    artifacts.each { artifact ->
+        def jsonContent = """
+        {
+            "osname": "fedora-coreos",
+            "container-repo": "${staging_repo}",
+            "container-tag": "${arch}-${shortcommit}",
+            "image-type": "${artifact["platform"]}",
+            "container-imgref": "ostree-remote-registry:fedora:${repo}:5.1",
+            "metal-image-size": "3072",
+            "cloud-image-size": "10240"
+        }
+        """
+        def file_path="./diskvars-${arch}-${artifact["suffix"]}.json"
+        writeFile file: file_path, text: jsonContent
+    }
+    if (arch != "x86_64") {
+        shwrap("""cosa remote-session sync ./diskvars-${arch}-*.json  {:,}""")
+    }
+}
+
+
+def supermin_run(cache, arch, artifact, shortcommit) {
+    shwrap("""cosa supermin-run ${cache} \
+       /usr/lib/coreos-assembler/runvm-osbuild \
+       --config ./diskvars-${arch}-${artifact}.json \
+       --filepath podman-$arch-${shortcommit}.${artifact} \
+       --mpp /usr/lib/coreos-assembler/osbuild-manifests/coreos.osbuild.${arch}.mpp.yaml
+    """)
+}
+
+
+properties([
+    pipelineTriggers([
+        [$class: 'GenericTrigger',
+         genericVariables: [
+          [
+           key: 'PODMAN_MACHINE_GIT_REF',
+           value: '$.ref',
+           expressionType: 'JSONPath',
+           regexpFilter: 'refs/heads/', //Optional, defaults to empty string
+           defaultValue: ''  //Optional, defaults to empty string
+          ]
+         ],
+         causeString: 'Triggered on $ref',
+         token: 'build-podman-os',
+         tokenCredentialId: '',
+         printContributedVariables: true,
+         printPostContent: true,
+         silentResponse: false,
+         regexpFilterText: '$PODMAN_MACHINE_GIT_REF',
+         regexpFilterExpression: 'main'
+        ]
+    ]),
+    parameters([
+      string(name: 'ARCHES',
+             description: 'Space-separated list of target architectures',
+             defaultValue: "x86_64 aarch64",
+             trim: true),
+      string(name: 'PODMAN_MACHINE_GIT_URL',
+             description: 'Override the coreos-assembler git repo to use',
+             defaultValue: "https://github.com/containers/podman-machine-os",
+             trim: true),
+      string(name: 'PODMAN_MACHINE_GIT_REF',
+             description: 'Override the coreos-assembler git ref to use',
+             defaultValue: "main",
+             trim: true),
+      string(name: 'CONTAINER_REGISTRY_REPO',
+             description: 'Override the registry to push the container to',
+             defaultValue: "quay.io/podman/machine-os",
+             trim: true),
+      string(name: 'CONTAINER_REGISTRY_STAGING_REPO',
+             description: 'Override the staging registry where intermediate images go',
+             defaultValue: "quay.io/podman/machine-os-staging",
+             trim: true),
+      string(name: 'COREOS_ASSEMBLER_IMAGE',
+             description: 'Override the coreos-assembler image to use',
+             defaultValue: "quay.io/coreos-assembler/coreos-assembler:main",
+             trim: true),
+      booleanParam(name: 'MANIFEST_RELEASE',
+                   defaultValue: false,
+                   description: 'Publish the Manifest in the official repo'),
+      booleanParam(name: 'FORCE',
+                   defaultValue: false,
+                   description: 'Whether to force a rebuild'),
+    ]),
+    buildDiscarder(logRotator(
+        numToKeepStr: '100',
+        artifactNumToKeepStr: '100'
+    )),
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+])
+
+// Note the supermin VM just uses 2G. The really hungry part is xz, which
+// without lots of memory takes lots of time. For now we just hardcode these
+// here; we can look into making them configurable through the template if
+// developers really need to tweak them.
+// XXX bump an extra 2G (to 10.5) because of an error we are seeing in
+// testiso: https://github.com/coreos/fedora-coreos-tracker/issues/1339
+def cosa_memory_request_mb = 4.5 * 1024 as Integer
+
+// Now that we've established the memory constraint based on xz above, derive
+// kola parallelism from that. We leave 512M for overhead and VMs are at most
+// 1.5G each (in general; root reprovisioning tests require 4G which is partly
+// why we run them separately).
+// XXX: https://github.com/coreos/coreos-assembler/issues/3118 will make this
+// cleaner
+def ncpus = ((cosa_memory_request_mb - 512) / 1536) as Integer
+
+
+node {
+    change = checkout(
+        changelog: true,
+        poll: false,
+        scm: [
+            $class: 'GitSCM',
+            branches: [[name: "origin/${params.PODMAN_MACHINE_GIT_REF}"]],
+            userRemoteConfigs: [[url: params.PODMAN_MACHINE_GIT_URL]],
+            extensions: [[$class: 'CloneOption',
+                          noTags: true,
+                          reference: '',
+                          shallow: true]]
+        ]
+    )
+
+    gitref = params.PODMAN_MACHINE_GIT_REF
+    def output = shwrapCapture("git rev-parse HEAD")
+    commit = output.substring(0,40)
+    shortcommit = commit.substring(0,7)
+}
+
+lock(resource: "build-podman-os") {
+    cosaPod(cpu: "${ncpus}",
+            memory: "${cosa_memory_request_mb}Mi",
+            image: params.COREOS_ASSEMBLER_IMAGE,
+            serviceAccount: "jenkins") {
+    timeout(time: 60, unit: 'MINUTES') {
+    try {
+
+        currentBuild.description = "[${gitref}@${shortcommit}] Running"
+        // Get the list of requested architectures to build for
+        def arches = params.ARCHES.split() as Set
+        def archinfo = arches.collectEntries{[it, [:]]}
+        for (arch in archinfo.keySet()) {
+            // initialize some data
+            archinfo[arch]['session'] = ""
+        }
+
+        // By default we will allow re-using cache layers for one day.
+        // This is mostly so we can prevent re-downloading the RPMS
+        // and repo metadata and over again in a given day for successive
+        // builds.
+        def cacheTTL = "24h"
+        def force = ""
+        if (params.FORCE) {
+            force = '--force'
+            // Also set cacheTTL to 0.1s to allow users an escape hatch
+            // to force no cache layer usage.
+            cacheTTL = "0.1s"
+        }
+
+        withCredentials([file(credentialsId: 'podman-push-registry-secret', variable: 'REGISTRY_SECRET')]) {
+            stage('Build Layered Image') {
+                 parallel arches.collectEntries{arch -> [arch, {
+                     pipeutils.withPodmanRemoteArchBuilder(arch: arch) {
+                         shwrap("""
+                             cosa remote-build-container \
+                                 --arch $arch --cache-ttl ${cacheTTL} \
+                                 --git-ref $commit ${force} \
+                                 --git-url ${params.PODMAN_MACHINE_GIT_URL} \
+                                 --git-sub-dir podman-image-daily \
+                                 --repo ${params.CONTAINER_REGISTRY_STAGING_REPO} \
+                                 --push-to-registry --auth=\$REGISTRY_SECRET
+                         """)
+                     }
+                 }]}
+            }
+        }
+        // Initialize the sessions on the remote builders
+        stage("Initialize Remotes") {
+            parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                if (arch != "x86_64") {
+                    pipeutils.withPodmanRemoteArchBuilder(arch: arch) {
+                        archinfo[arch]['session'] = pipeutils.makeCosaRemoteSession(
+                            expiration: "60m",
+                            image: params.COREOS_ASSEMBLER_IMAGE,
+                            workdir: WORKSPACE,
+                        )
+                    }
+                }
+            }]}
+        }
+        stage("Init") {
+            parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                pipeutils.withOptionalExistingCosaRemoteSession(
+                arch: arch, session: archinfo[arch]['session']) {
+                    shwrap("""cosa init --force https://github.com/coreos/fedora-coreos-config.git""")
+                }
+            }]}
+        }
+        stage("Build Disk Images") {
+            parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                pipeutils.withOptionalExistingCosaRemoteSession(
+                arch: arch, session: archinfo[arch]['session']) {
+                    generate_disk_images(artifacts, shortcommit, arch, params.CONTAINER_REGISTRY_STAGING_REPO,
+                    params.CONTAINER_REGISTRY_REPO)
+                }
+            }]}
+        }
+        stage("Kola") {
+            parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                pipeutils.withOptionalExistingCosaRemoteSession(
+                    arch: arch, session: archinfo[arch]['session']) {
+                    shwrap("""cosa kola run  --basic-qemu-scenarios --qemu-image=./podman-${arch}-${shortcommit}.qcow2""")
+                }
+            }]}
+        }
+        stage("Compress Artifacts") {
+            parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                pipeutils.withOptionalExistingCosaRemoteSession(
+                arch: arch, session: archinfo[arch]['session']) {
+                    parallel artifacts.collectEntries{artifact -> [artifact["platform"] , {
+                        def file = "podman-$arch-${shortcommit}.${artifact["suffix"]}"
+                        shwrap("""cosa shell -- /usr/bin/zstd -14 $file""")
+                    }]}
+                }
+            }]}
+        }
+        stage("Sync Artifacts") {
+            parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                pipeutils.withOptionalExistingCosaRemoteSession(
+                arch: arch, session: archinfo[arch]['session']) {
+                    if (arch != "x86_64") {
+                        parallel artifacts.collectEntries{artifact -> [artifact["platform"] , {
+                            shwrap("""cosa remote-session sync {:,}podman-$arch-${shortcommit}.${artifact["suffix"]}.zst """)
+                        }]}
+                    }
+                }
+            }]}
+        }
+        withCredentials([file(credentialsId: 'ravanelli-push-registry-secret', variable: 'REGISTRY_SECRET')]) {
+            stage("Push Manifest") {
+                def manifest = "${params.CONTAINER_REGISTRY_STAGING_REPO}:5.1-${shortcommit}"
+                def cmds = "/usr/bin/buildah manifest create ${manifest}"
+                parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                    def image = "docker://${params.CONTAINER_REGISTRY_STAGING_REPO}:${arch}-${shortcommit}"
+                    cmds += " && /usr/bin/buildah manifest add ${manifest} ${image}"
+                    artifacts.each { artifact ->
+                        def file = "podman-$arch-${shortcommit}.${artifact["suffix"]}.zst"
+                        def platform = "${artifact["platform"]}"
+                        cmds += " && /usr/bin/buildah manifest add --artifact --artifact-type='' \
+                                --os=linux --arch=${arch} --annotation 'disktype=${platform}' ${manifest} ${file}"
+                    }
+                }]}
+                //  `cosa supermin-run` only mounts the current directory, so we
+                // need to temporarily put the secret here. We'll delete it right after.
+                shwrap("""cp ${REGISTRY_SECRET} push-secret""")
+
+                // We need to run all buildah commands in the same superminvm to not lose
+                // the manifest creation/additions
+                shwrap("""cosa supermin-run bash -c \
+                    "${cmds} && \
+                    /usr/bin/buildah manifest push --authfile ./push-secret --all ${manifest} \
+                    docker://${manifest}"
+                """)
+                shwrap("""rm -f  push-secret""")
+            }
+        }
+        if (params.MANIFEST_RELEASE) {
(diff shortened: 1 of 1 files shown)


ravanelli:
I still need the podam repo names so I can replace from mine

jlebon (commented):
First, thank you for working on this!

I haven't done a detailed review yet, but one thing that jumped out is the use of two separate jobs. Notably, we're building the x86_64 artifacts first, and then the aarch64 artifacts in a separate job. I think ideally it would all be in a single job instead. This keeps it self-contained, but also e.g. things like manifest list building become easier if it's all in a single job.

See e.g. how the `bump-lockfile` or [`test-override`](https://github.com/coreos/coreos-ci/blob/main/jobs/test-override.Jenkinsfile) jobs are structured.

OK, here's a rough outline of what I mean:

```groovy
cosaPod(...) {
	// build images for every arch, push to some staging repo
	stage("Build Layered Image") {
		parallel arches.collectEntries{arch -> [arch, {
			cosa remote-build-container ...
		}]}
	}
	// now we build disk images
	stage("Initialize Remotes") {
		parallel arches.collectEntries{arch -> [arch, {
			if (arch != x86_64) {
				cosa remote-session create
				// store remote ID in e.g. archinfo map
			}
		}]}
	}
	stage("Build Disk Images") {
		parallel arches.collectEntries{arch -> [arch, {
			withOptionalExistingCosaRemoteSession(arch: arch, session: archinfo[arch]['session'] {
				cosa supermin-run ...
			}
		}]}
	}
	// convert disk images to container images
	// push all the container images to some staging repo
	// build manifest lists
	// push out manifest lists to final repo
	// clean up staging repo
}
```

This also avoids copying any container image between builders.

WDYT?

ravanelli:
In fact, this was my first approach, two arches in one  job, but thinking about the case of an arch failing, this would limit the rebuild part, so I decided to split it and follow the way we do with our multiarch ones

jlebon:
> In fact, this was my first approach, two arches in one job, but thinking more about the case of an arch failing, this would limit the rebuild part

This is a good point. I think we should be able to hit the cache at least for the building of the layered images stage. You're right we'd probably lose the work on the disk images though. That said, since we're not e.g. running the full testsuite, I think in practice this job will be quite stable (just like e.g. `build-cosa` is pretty stable today and mostly flakes on infra issues).

> and both jobs can run almost in parallel

In the suggested approach, we'd be parallelizing across arches right from the start, so it *should* be faster too. :) (Though I think the speed should be quite reasonable either approach overall.)

> What do you mean by: This also avoids copying any container image between builders?

IIUC, we're doing `cosa remote-build-container ... --write-to-file ...` which will transfer the OCI archive from the builder to Jenkins, and then later on `cosa remote-session sync` to sync it back to the remote, right? I think ideally we should push directly to the staging repo from the remote builders (both for the OS container and the disk images) to avoid incurring that cost.

>  I'm fine with this way too!

:+1: 

I think you have all the key pieces here, so it'd just be about re-arranging them a bit. :) If you want to defer this and try to get this approach in first, I'm OK with that too. If you want to get together to try to hammer this out, we can do that too.

lsm5:
> I still need the podam repo names so I can replace from mine

@ravanelli thanks, you can use `quay.io/podman/staging` and `quay.io/podman/machine-images` for now. Also, the secret can be replaced with `podman-push-registry-secret`

ravanelli:
We still need to define the manifest final merge, if it is not possible I will need the artifacts back to the host to then create the full manifest with all arches on it.

Here is how the staging looks like:
https://quay.io/repository/ravanelli/coreos-assembler-staging

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,379 @@
...
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+])
+
+// Note the supermin VM just uses 2G. The really hungry part is xz, which
+// without lots of memory takes lots of time. For now we just hardcode these
+// here; we can look into making them configurable through the template if
+// developers really need to tweak them.
+// XXX bump an extra 2G (to 10.5) because of an error we are seeing in
+// testiso: https://github.com/coreos/fedora-coreos-tracker/issues/1339
+def cosa_memory_request_mb = 10.5 * 1024 as Integer
```
jlebon, line 121: I think we can probably bring this down by a lot. Especially since we're using zstd which I don't think is as memory hungry.

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,379 @@
...
+
+// Now that we've established the memory constraint based on xz above, derive
+// kola parallelism from that. We leave 512M for overhead and VMs are at most
+// 1.5G each (in general; root reprovisioning tests require 4G which is partly
+// why we run them separately).
+// XXX: https://github.com/coreos/coreos-assembler/issues/3118 will make this
+// cleaner
+def ncpus = ((cosa_memory_request_mb - 512) / 1536) as Integer
```
jlebon, line 129: This too.

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,379 @@
+def gitref, commit, shortcommit
+def containername = 'podman-fcos'
```
jlebon, line 2: This is only used in the `lock()`; seems OK to drop this variable and hardcode "build-podman-fcos" in the `lock()`?

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,379 @@
...
+
+        currentBuild.description = "[${gitref}@${shortcommit}] Running"
+        // Get the list of requested architectures to build for
+        def arches = params.ARCHES.split() as Set
+        def archinfo = arches.collectEntries{[it, [:]]}
+        for (arch in archinfo.keySet()) {
+            // initialize some data
+            archinfo[arch]['session'] = ""
+        }
+        def artifacts = [
```
jlebon, line 171: This is important; probably worth hoisting to the top definitions for more visibility?

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,379 @@
...
+                                image: params.COREOS_ASSEMBLER_IMAGE,
+                                workdir: WORKSPACE,
+                            )
+                            withEnv(["COREOS_ASSEMBLER_REMOTE_SESSION=${archinfo[arch]['session']}"]) {
+                                 shwrap("""cosa init --force https://github.com/coreos/fedora-coreos-config.git""")
+                            }
+                        }
+                    } else {
+                         shwrap("""cosa init --force https://github.com/coreos/fedora-coreos-config.git""")
+                    }
```
jlebon, line 226: This part could be a separate step from remote session initialization.

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,379 @@
...
+                }]}
+            }
+
+            stage("Build Disk Images") {
+                  parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                        if (arch != "x86_64") {
```
jlebon, line 232: We shouldn't need this conditional. `withOptionalExistingCosaRemoteSession()` already knows to directly execute the passed body directly if it's x86_64 (well really, if no `session` is defined). So this would just be:

```
parallel ... {
  pipeutils.withOptional...() {
    generate_disk_images(...)
  }
}
```

And similarly for the other stages.

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,379 @@
...
+                        }
+                   }]}
+	    }
+
+           stage("Kola") {
+               parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                   if (arch != "x86_64") {
+                       pipeutils.withOptionalExistingCosaRemoteSession(
+                           arch: arch, session: archinfo[arch]['session']) {
+                           shwrap("""cosa kola run  --basic-qemu-scenarios --qemu-image=./podman-${shortcommit}.qcow2""")
```
jlebon, line 248: Aside: keep an eye on https://github.com/coreos/coreos-assembler/pull/3652; we'll have to update this once that PR merges.

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,379 @@
...
+
+           stage("Compress Artifacts") {
+               parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                    if (arch != "x86_64") {
+                        def session =  archinfo[arch]['session']
+                        pipeutils.withOptionalExistingCosaRemoteSession(
+                        arch: arch, session: session) {
+                            parallel artifacts.collectEntries{artifact -> [artifact["platform"] , {
+                                def file = "podman-$arch-${shortcommit}.${artifact["suffix"]}"
+                                shwrap("""podman --remote exec -i $session /usr/bin/zstd -14 $file""")
```
jlebon, line 264: The canonical way to run arbitrary commands is `cosa shell --`. That'll work both in the remote and non-remote case (related to the previous comment about `withOptionalExistingCosaRemoteSession()`).

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,379 @@
...
+        def cacheTTL = "5h"
+        def force = ""
+        if (params.FORCE) {
+            force = '--force'
+            // Also set cacheTTL to 0.1s to allow users an escape hatch
+            // to force no cache layer usage.
+            cacheTTL = "0.1s"
+        }
+
+        withCredentials([file(credentialsId: 'podman-push-registry-secret', variable: 'REGISTRY_SECRET')]) {
```
jlebon, line 190: Ideally we would limit the credential scope to the stages that need it.

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,379 @@
...
+        currentBuild.result = 'FAILURE'
+        throw e
+    } finally {
+        if (currentBuild.result == 'SUCCESS') {
+            currentBuild.description = "[${gitref}@${shortcommit}] ⚡"
+        } else {
+            currentBuild.description = "[${gitref}@${shortcommit}] ❌"
+        }
+        if (currentBuild.result != 'SUCCESS') {
+            message = "build-cosa #${env.BUILD_NUMBER} <${env.BUILD_URL}|:jenkins:> <${env.RUN_DISPLAY_URL}|:ocean:> [${gitref}@${shortcommit}]"
```
jlebon, line 374: This needs updating.

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,379 @@
...
+                // Although this solution is not ideal, let's proceed with it until we
+                // can find a more effective approach.
+                // This doc includes a 'use' case that didn't function as expected.
+                // https://www.jenkins.io/doc/pipeline/steps/credentials-binding/
+                shwrap("""cp ${REGISTRY_SECRET} push-secret""")
+
+                parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                    def manifest = "podman-v5"
+                    def cmds = ""
+                    if (arch != "x86_64") {
```
jlebon, line 288: Similarly, I think we should be able to get rid of this.

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,379 @@
...
+           }
+
+           stage("Push Manifest") {
+
+                //  Supermin is unable to access the secret mounted in another directory.
+                // Although this solution is not ideal, let's proceed with it until we
+                // can find a more effective approach.
+                // This doc includes a 'use' case that didn't function as expected.
+                // https://www.jenkins.io/doc/pipeline/steps/credentials-binding/
+                shwrap("""cp ${REGISTRY_SECRET} push-secret""")
```
jlebon, line 283: So actually, we do have a nice wrapper for this: [`utils.syncCredentialsIfInRemoteSession()`](https://github.com/coreos/coreos-ci-lib/blob/cf8e6a14b2c8b6d83940aa3f20ab6847345fd379/vars/utils.groovy#L65). You can see sample usage in other code in this repo.

jlebon (commented):
This is looking really great! :+1:

lsm5:
> We still need to define the manifest final merge, if it is not possible I will need to sync the artifacts back to the host to then create the full manifest with all arches on it.
> 
> Here is how the staging looks like: https://quay.io/repository/ravanelli/coreos-assembler-staging?tab=tags
> 
> @lsm5 @baude, also please validate the release manifest part, is that what you need? We can keep pushing to staging until we decided to move it to the official repo.

@baude PTAL

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,339 @@
...
+// since it is podman job only
+
+def generate_disk_images(artifacts, shortcommit, arch, repo) {
+    generate_diskvar_json(shortcommit, arch, artifacts, repo)
+    // First create the qcow2 and store it in cache
+    supermin_run("--cache", arch, "qcow2", shortcommit)
+    def artifacts_runvm = artifacts.drop(1) // drop qcow2
+    parallel artifacts_runvm.collectEntries{artifact -> [artifact["platform"], {
+        supermin_run("--snapshot", arch, "${artifact["suffix"]}", shortcommit)
+    }]}
```
dustymabe, line 25: ouch.. this isn't the most ergonomic because of the way runvm-osbuild essentially takes one platform/filename at a time. I'll maybe see if there is way to make it better so we can just create all the platforms with one single run.

jobs/build-podman-os.Jenkinsfile
```diff
@@ -0,0 +1,347 @@
...
+                // https://www.jenkins.io/doc/pipeline/steps/credentials-binding/
+                shwrap("""cp ${REGISTRY_SECRET} push-secret""")
+
+                // We need to run all buildah commands in the same superminvm to not lose
+                // the manifest creation/additions
+                shwrap("""cosa supermin-run bash -c \
+                    "/usr/bin/buildah manifest create ${manifest} ${cmds} && \
+                    /usr/bin/buildah manifest push --authfile ./push-secret --all ${manifest} \
+                    docker://${params.CONTAINER_REGISTRY_STAGING_REPO}:5.0-${shortcommit}"
+                """)
```
dustymabe, line 297: in the future we should look into trying to do this outside of supermin.

right now in FCOS we don't use supermin when combining into a manifest list and pushing to the registry:

https://github.com/coreos/fedora-coreos-pipeline/blob/6eebb70044dc67a349e826006ebcbf922e8b5b24/jobs/release.Jenkinsfile#L267-L271

which just uses https://github.com/coreos/coreos-assembler/blob/main/src/cmd-push-container-manifest


we also don't use it when building cosa:

https://github.com/coreos/fedora-coreos-pipeline/blob/6eebb70044dc67a349e826006ebcbf922e8b5b24/jobs/build-cosa.Jenkinsfile#L131-L154


Maybe the `export STORAGE_DRIVER=vfs` is the key.

jobs/build-podman-os.Jenkinsfile
```diff
@@ -0,0 +1,347 @@
...
+                        docker://${params.CONTAINER_REGISTRY_REPO}:5.0
+                    """)
+                    parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                        // Release the manifest and container images
+                        shwrap("""
+                            skopeo copy --all --authfile \$REGISTRY_SECRET   \
+                            docker://${params.CONTAINER_REGISTRY_STAGING_REPO}:${arch}-${shortcommit} \
+                            docker://${params.CONTAINER_REGISTRY_REPO}:5.0-${arch}
+                        """)
+                    }]}
```
dustymabe, line 316: can you help me understand what is going on here behind the scenes?

What is the difference between these two?:
- staging to main repo `5.0-${shortcommit} -> 5.0`
- for each arch
    - staging to main repo `${arch}-${shortcommit} -> 5.0-${arch}`

dustymabe:
also, why did you drop the `stage('Delete Intermediate Tags')`? was it not working? or do we ant to keep them around for some reason?

jobs/build-podman-os.Jenkinsfile
ravanelli: The `build-cosa` doesn't use kvm. Since we need supermin to run osbuild we need to enable it. Seems it is causing issues with podman. For `release` the  `push-container-manifest` uses skopeo

ravanelli:
> also, why did you drop the `stage('Delete Intermediate Tags')`? was it not working? or do we ant to keep them around for some reason?
Kind of, if we clean up staging we lose the ability in avoiding to build the container image if it is already there. As we can also add a check  further for the artifacts builds as well. I was thinking in maybe adding an expiration tag for it.

dustymabe:
> Kind of, if we clean up staging we lose the ability in avoiding to build the container image if it is already there. As we can also add a check further for the artifacts builds as well. I was thinking in maybe adding an expiration tag for it.

hmm. if we get to the point where we delete the intermediate tags then we got all the way to success right? so we wouldn't ever need them again?

We do this today in the build-cosa job.. if the job fails part way through then it won't clean up the intermediate tags and if some arches were successful then it will just re-use those. If it gets all the way to success then it cleans them up.

ravanelli:
>> hmm. if we get to the point where we delete the intermediate tags then we got all the way to success right? so we wouldn't ever need them again?
Yes!

>> We do this today in the build-cosa job.. if the job fails part way through then it won't clean up the intermediate tags and if some arches were successful then it will just re-use those. If it gets all the way to success then it cleans them up.
Yeah, that's make sense. I'm fine adding it back (will do)

jobs/build-podman-os.Jenkinsfile
dustymabe: yes. I think the question I'm asking is: why?

why do we need that? are the users going to use them directly? is it some artifact of the way we are pushing that they need to be there?

jobs/build-podman-os.Jenkinsfile
dustymabe: > Since we need supermin to run osbuild we need to enable it.

right. we need supermin for when we run `osbuild`, but we're not running `osbuild` here, we're running `buildah`. 



> For `release` the `push-container-manifest` uses skopeo

right. `podman`, `buildah`, `skopeo` all use the same libraries underneath. I suspect we can figure out how to run `buildah manifest` commands here without `supermin` too, but we don't need to block on that.

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,339 @@
...
+        for (arch in archinfo.keySet()) {
+            // initialize some data
+            archinfo[arch]['session'] = ""
+        }
+
+        // By default we will allow re-using cache layers for one day.
+        // This is mostly so we can prevent re-downloading the RPMS
+        // and repo metadata and over again in a given day for successive
+        // builds.
+        def cacheTTL = "5h"
```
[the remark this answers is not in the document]
dustymabe: this is now 24h

jobs/build-podman.Jenkinsfile
```diff
@@ -0,0 +1,339 @@
...
+        // Initialize the sessions on the remote builders
+        stage("Initialize Remotes") {
+            parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                if (arch != "x86_64") {
+                    pipeutils.withPodmanRemoteArchBuilder(arch: arch) {
+                        archinfo[arch]['session'] = pipeutils.makeCosaRemoteSession(
+                            // The cache size for aarch64 in runvm is insufficient,
+                            // even when considering that all artifacts are around
+                            // the same size as those for x86_64
+                            env: ["OSBUILD_EXPORT_FORCE_NO_PRESERVE_OWNER":"true", "RUNVM_CACHE_SIZE": "25G"],
```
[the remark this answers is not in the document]
dustymabe: looks like this was updated

jobs/build-podman-os.Jenkinsfile
```diff
@@ -0,0 +1,360 @@
...
+
+def generate_diskvar_json(shortcommit, arch, artifacts, repo) {
+    artifacts.each { artifact ->
+        def jsonContent = """
+        {
+            "osname": "fedora-coreos",
+            "container-repo": "$repo",
+            "container-tag": "${arch}-${shortcommit}",
+            "image-type": "${artifact["platform"]}",
+            "container-imgref": "ostree-remote-registry:fedora:quay.io/containers/podman-machine-os:5.0",
```
dustymabe, line 36: I feel like this value needs to be updated?

jobs/build-podman-os.Jenkinsfile
ravanelli: I think it is a question for @baude

baude: @dustymabe i dont get what you are asking me, ill reach out

dustymabe: ok talked with brent,

basically we need the `5.0` tag to include the oci artifacts too (i.e. the containers and not just the disk images) so we'll need to add a few more steps to the `buildah manifest` command run in the supermin VM. 

we don't actually need the `5.0-aarch64` and `5.0-x86_64` in the registry so we can just not push those (i.e. we can delete the second `skopeo copy`).

jobs/build-podman-os.Jenkinsfile
```diff
@@ -0,0 +1,360 @@
...
+             description: 'Override the coreos-assembler git repo to use',
+             defaultValue: "https://github.com/baude/podman-machine-images.git",
+             trim: true),
+      string(name: 'PODMAN_MACHINE_GIT_REF',
+             description: 'Override the coreos-assembler git ref to use',
+             defaultValue: "main",
+             trim: true),
+      string(name: 'CONTAINER_REGISTRY_REPO',
+             description: 'Override the registry to push the container to',
+             defaultValue: "quay.io/podman/machine-images",
```
dustymabe, line 96: From talking with brent I think this is where he is pushing too:

```suggestion
             defaultValue: "quay.io/podman/machine-os",
```

jobs/build-podman-os.Jenkinsfile
```diff
@@ -0,0 +1,360 @@
...
+
+def generate_diskvar_json(shortcommit, arch, artifacts, repo) {
+    artifacts.each { artifact ->
+        def jsonContent = """
+        {
+            "osname": "fedora-coreos",
+            "container-repo": "$repo",
+            "container-tag": "${arch}-${shortcommit}",
+            "image-type": "${artifact["platform"]}",
+            "container-imgref": "ostree-remote-registry:fedora:quay.io/podman/machine-images:5.0",
```
dustymabe, line 36: if we do https://github.com/coreos/fedora-coreos-pipeline/pull/968/files#r1531138529 then this will need updating too

lsm5 (commented):
lgtm as far as i can tell. The repo paths look ok too, but maybe `quay.io/podman/staging` should be replaced with something more specific like: `quay.io/podman/machine-os-staging` ? @baude if you agree, I can set that up and update here.

jobs/build-podman-os.Jenkinsfile
```diff
@@ -0,0 +1,355 @@
...
+                // This doc includes a 'use' case that didn't function as expected.
+                // https://www.jenkins.io/doc/pipeline/steps/credentials-binding/
+                shwrap("""cp ${REGISTRY_SECRET} push-secret""")
+
+                // We need to run all buildah commands in the same superminvm to not lose
+                // the manifest creation/additions
+                shwrap("""cosa supermin-run bash -c \
+                    "/usr/bin/buildah manifest create ${manifest} ${cmds} && \
+                    /usr/bin/buildah manifest push --authfile ./push-secret --all ${manifest} \
+                    docker://${manifest}"
```
dustymabe, line 298: is the `docker://` required here? 
```suggestion
                    ${manifest}"
```

Looking at the COSA code that pushes I don't think we have `docker://` on there.

dustymabe (approved):
LGTM

jobs/build-podman-os.Jenkinsfile
ravanelli: If I recall buildah complains about the transport if we don't add it.

dustymabe: ok

dustymabe (approved):

jobs/build-podman-os.Jenkinsfile
```diff
@@ -0,0 +1,355 @@
...
+    )
+
+    gitref = params.PODMAN_MACHINE_GIT_REF
+    def output = shwrapCapture("git rev-parse HEAD")
+    commit = output.substring(0,40)
+    shortcommit = commit.substring(0,7)
+}
+
+
+// Get the list of requested architectures to build for
```
jlebon, line 159: Minor/optional: comment looks out of date -- delete?

jobs/build-podman-os.Jenkinsfile
```diff
@@ -0,0 +1,355 @@
...
+                }
+            }]}
+        }
+        stage("Init") {
+            parallel archinfo.keySet().collectEntries{arch -> [arch, {
+                pipeutils.withOptionalExistingCosaRemoteSession(
+                arch: arch, session: archinfo[arch]['session']) {
+                    shwrap("""cosa init --force https://github.com/coreos/fedora-coreos-config.git""")
+                }
+                shwrap("""cosa init --force https://github.com/coreos/fedora-coreos-config.git""")
```
jlebon, line 229: Wait, why are we doing another `cosa init` here? The one above should work fine across all arches.

jobs/build-podman-os.Jenkinsfile
```diff
@@ -0,0 +1,355 @@
...
+                        def platform = "${artifact["platform"]}"
+                        cmds += " && /usr/bin/buildah manifest add --artifact --artifact-type='' \
+                                --os=linux --arch=${arch} --annotation 'disktype=${platform}' ${manifest} ${file}"
+                    }
+                }]}
+                //  Supermin is unable to access the secret mounted in another directory.
+                // Although this solution is not ideal, let's proceed with it until we
+                // can find a more effective approach.
+                // This doc includes a 'use' case that didn't function as expected.
+                // https://www.jenkins.io/doc/pipeline/steps/credentials-binding/
```
jlebon, line 290: Minor/optional:

```suggestion
                //  `cosa supermin-run` only mounts the current directory, so we
                // need to temporarily put the secret here. We'll delete it right after.
```

jlebon (commented):
@ravanelli Nice work! This looks much nicer now.

Only two comments I think we should address. The rest is a string of minor comments that you're free to ignore/address in a follow up.

jobs/build-podman-os.Jenkinsfile
```diff
@@ -0,0 +1,355 @@
...
+    shortcommit = commit.substring(0,7)
+}
+
+
+// Get the list of requested architectures to build for
+lock(resource: "build-podman-os") {
+    cosaPod(cpu: "${ncpus}",
+            memory: "${cosa_memory_request_mb}Mi",
+            image: params.COREOS_ASSEMBLER_IMAGE,
+            env: ["OSBUILD_EXPORT_FORCE_NO_PRESERVE_OWNER":"true"],
```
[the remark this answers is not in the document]
dustymabe: TBH I'm pretty sure we're not actually using this (or rather needing this) here becuase we are using runvm-osbuild which sets `outdir=cache/osbuild/out` so it's the same FS and thus no copying over virtiofs.

jobs/build-podman-os.Jenkinsfile
```diff
@@ -0,0 +1,355 @@
...
+                           parallel artifacts.collectEntries{artifact -> [artifact["platform"] , {
+                               shwrap("""cosa remote-session sync {:,}podman-$arch-${shortcommit}.${artifact["suffix"]}.zst """)
+                           }]}
+                       }
+                   }
+            }]}
+        }
+        withCredentials([file(credentialsId: 'ravanelli-push-registry-secret', variable: 'REGISTRY_SECRET')]) {
+            stage("Push Manifest") {
+                def manifest = "${params.CONTAINER_REGISTRY_STAGING_REPO}:5.1-${shortcommit}"
```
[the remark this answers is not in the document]
ravanelli: Yes, we are still defining from where it will come from, that's why I keept  it hard code for now

dustymabe:
I think my `OSBUILD_EXPORT_FORCE_NO_PRESERVE_OWNER` comment is still outstanding: https://github.com/coreos/fedora-coreos-pipeline/pull/968#discussion_r1537680287

We don't need to remove it, but it is confusing to have it when we don't need it.

dustymabe (approved):

jobs/build-podman-os.Jenkinsfile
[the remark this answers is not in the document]
ravanelli: I missed it before, fixed

dustymabe (approved):
LGTM

lsm5:
@ravanelli @jlebon @dustymabe thanks a lot for this!!

PR #969, opened 2024-03-19
`next-devel` is now on Fedora 40 so disable `branched` stream

aaradhak:
Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1674


dustymabe (approved):
LGTM

PR #970, opened 2024-03-20
drop skip_govcloud_hack

dustymabe:
We are no longer building any release that requires this hack. Let's drop it.


jlebon (approved):
:tada:

PR #972, opened 2024-03-26
build-podman-os: Fix secret mount

ravanelli:


ravanelli:
Seems the last fix missing.

dustymabe:
let's talk about this a minute.

in the case where we aren't releasing we're pushing to the staging repo and then immediately deleting from the staging repo. I don't think this is what we want.

I think what we should do here is only push `${params.CONTAINER_REGISTRY_STAGING_REPO}:5.1-${shortcommit}` to the staging repo if we aren't releasing. If we are releasing then push to `${params.CONTAINER_REGISTRY_REPO}:5.1`.

In no case should we delete the `5.1` tag as intermediate. The only tags we delete as intermediate are the `${arch}-${shortcommit}` ones.

dustymabe:
I can open a followup PR to implement this and have you review if that would be easier.

jlebon (approved):

ravanelli:
> in the case where we aren't releasing we're pushing to the staging repo

The real push is done in the manifest [see](https://github.com/coreos/fedora-coreos-pipeline/blob/main/jobs/build-podman-os.Jenkinsfile#L291), since it is running using runvm we need to push it there to some place.  
I had remove this delete before, and as you mentioned if we are here it should be ok to delete everything.

ravanelli:
Ok, I think I see what changed here between the interactions, before I added the `'Delete Intermediate Tags` only if we were releasing the manifest, it is now out of this block

dustymabe:
I think there is still some restructuring we can do here, but I think maybe we should do it as part of a larger piece of work. i.e. if someone today tested a payload pushed to the staging repo and then wanted to release it, this job wouldn't just copy the existing payload over. It would build again (fresh) and release that.

dustymabe (approved):
Let's merge this and then we can talk about later improvements.

PR #974, opened 2024-04-01
drop support for hacks associated with old RHCOS releases

dustymabe:
- cosa_controller_img_hack
- oscontainer_old
- legacy_oscontainer_old


jlebon (approved):
Nice!

We should also delete the secret from the cluster and then signal that we no longer need it and remove it from BitWarden also.

PR #975, opened 2024-04-08
 build-podman-os: Add extra-kargs

ravanelli:
- Reflect changes from https://github.com/coreos/coreos-assembler/pull/3770, adding the extra-kargs configuration from image-base.yaml

### DIFF
--- a/jobs/build-podman-os.Jenkinsfile
+++ b/jobs/build-podman-os.Jenkinsfile
@@ -33,6 +33,9 @@ def generate_diskvar_json(shortcommit, arch, artifacts, staging_repo, repo) {
             "osname": "fedora-coreos",
             "container-repo": "${staging_repo}",
             "container-tag": "${arch}-${shortcommit}",
+            // Do not pass anything to kernel args for podman
+            // See: https://github.com/coreos/fedora-coreos-pipeline/pull/975
+            "extra-kargs-string": "",
             "image-type": "${artifact["platform"]}",
             "container-imgref": "ostree-remote-registry:fedora:${repo}:5.1",
             "metal-image-size": "3072",


jobs/build-podman-os.Jenkinsfile
```diff
@@ -27,12 +27,17 @@ def generate_disk_images(artifacts, shortcommit, arch, staging_repo, repo) {
 
 
 def generate_diskvar_json(shortcommit, arch, artifacts, staging_repo, repo) {
+    extra_kargs = shwrapCapture("""
+           python3 -c 'import sys, yaml; data = yaml.safe_load(sys.stdin)["extra-kargs"][0]; \
+           print(data)' < src/config/image-base.yaml
+    """)
```
dustymabe, line 33: honestly I'm not sure what the best approach is right here.

eventually I think we want this whole process decoupled from Fedora CoreOS and this is probably the opposite direction.

I think we should either hardcode the value here (`mitigations=auto,nosmt`) or possibly we decide that `mitigations=auto,nosmt` doesn't make sense for podman machine (i.e. it's running on someone's laptop, not shared hardware) and we omit adding that kernel argument at all here.

Would be interested in input from @jlebon and @baude here.

jlebon: Chatted with Dusty about this OOB.

In the general case, I think ideally running osbuild does pick up the values from the embedded `image.json` (or maybe eventually the same config files as bootc).

Specifically for this case, yeah I think probably we don't need to pick up this karg here (and in the future, this would be e.g. overriding it during the container build phase instead).

cgwalters: >  ideally running osbuild does pick up the values from the embedded image.json (or maybe eventually the same config files as bootc).

For the latter this will happen as part of https://github.com/osbuild/bootc-image-builder/issues/18 which will unblock using https://containers.github.io/bootc/man-md/bootc-install-config.html which can be embedded in the container image.

So once FCOS derives from a shared base image we can just add a little bootc install drop-in which will configure that install-time karg in a container-native way.

dustymabe (approved):
LGTM

PR #977, opened 2024-04-10
build-podman-os: Change trigger to run daily jobs

ravanelli:
- Ajust the job trigger to run daily instead of repo changes


jlebon (approved):

jlebon:
(One can guess, though more context in the commit message about the why would be helpful.)

jlebon (approved):

PR #979, opened 2024-04-16
jobs/build-kola-containers: build weekly

dustymabe:
Currently the last time this job ran was over a month ago. Let's build periodically to keep the content fresh and to also know sooner if breaking changes occur.

### DIFF
--- a/jobs/build-kola-containers.Jenkinsfile
+++ b/jobs/build-kola-containers.Jenkinsfile
@@ -7,6 +7,8 @@ node {
 
 properties([
     pipelineTriggers([
+        // run weekly to ensure regular updates
+        cron("@weekly"),
         [$class: 'GenericTrigger',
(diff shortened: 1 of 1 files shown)


jobs/build-kola-containers.Jenkinsfile
```diff
@@ -7,6 +7,8 @@ node {
 
 properties([
     pipelineTriggers([
+        // run weekly to ensure regular updates
+        cron("H H */7 * *"),
```
jlebon, line 11: Hmm, shouldn't it be `H H * * H`?

But actually, we can use `@weekly` instead: https://www.jenkins.io/doc/book/pipeline/syntax/#cron-syntax

dustymabe: updated!

PR #981, opened 2024-04-17
jobs/build-podman-os: adapt basic kola tests

dustymabe:
This changed form with https://github.com/coreos/coreos-assembler/pull/3652


jlebon (approved):
I guess we could tag them with `basic` too so you can do `--tag basic`. Feels slightly cleaner but meh...

PR #983, opened 2024-04-23
config: enable OSBuild for testing/testing-devel

dustymabe:
They are now on F40 and we can roll this out!

xref: https://github.com/coreos/fedora-coreos-tracker/issues/1653


jlebon (approved):
Soon, we'll be able to flip the default in cosa!

travier (approved):

PR #984, opened 2024-04-24
config.yaml: disable next-devel

aaradhak:
`testing-devel` is moved back over to F40 and we don't have any differences in `testing-devel` and `next-devel`, so we'll disable it.

Ref : https://github.com/coreos/fedora-coreos-tracker/issues/1674


dustymabe (approved):
LGTM

PR #985, opened 2024-04-29
Bring back `cosa_controller_img` knob, drop hack status

jlebon:
The podman remote feature assumes that the version on the client and the one on the host match versions, within some margin. So the `cosa_controller_img` knob we had before was useful.

This reverts commit dbad52bc5f11b73fad6459e54a1ed797fc0fb3a8. But:
1. we drop the `_hack` suffix and update the documentation
2. the default is to always use the `:main` cosa image, which most of the time will have a podman remote matching the FCOS version on the builder (except for that small period of time during a major rebase when FCOS is ahead of cosa)

Motivated by our pipeline hitting
https://github.com/containers/podman/discussions/22531
due to most FCOS streams being on f40 and cosa still being on f39.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -75,6 +75,9 @@ streams:
           - vmware
       # OPTIONAL: override default coreos-assembler image to build with
       cosa_img: quay.io/coreos-assembler/coreos-assembler:stable
+      # OPTIONAL: Override default coreos-assembler image to use for
+      # the controller pod in the build-arch job (default is :main)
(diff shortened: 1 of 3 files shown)


dustymabe (approved):
LGTM

dustymabe:
I still like the mental model of using the same image everywhere (i.e. if you specified cosa:4.16 then it runs everywhere), but this appears more practical as our infra needs change over time.

PR #994, opened 2024-05-20
jobs/bump-jenkins-job: Change in emoji of slack notification

aaradhak:
A PR would be created after the build run of bump-jenkins-job. To notify about the open PR, it would be better if the emoji is set to ':pr'.


dustymabe (approved):
LGTM

PR #995, opened 2024-05-21
jobs/release: tag containers images with version

jbtrystram:
This tag the manifest with the full version number. First step for https://github.com/coreos/fedora-coreos-tracker/issues/1367

Is it that simple ? 
Obviously I need to add the GC code as well

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -37,7 +37,7 @@ s3:
 registry_repos:
   oscontainer:
     repo: quay.io/fedora/fedora-coreos
-    tags: ["${STREAM}"]
+    tags: ["${STREAM}", "${VERSION}"]
   kubevirt:
     repo: quay.io/fedora/fedora-coreos-kubevirt
     tags: ["${STREAM}"]


dustymabe:
haven't actually looked deeply at the code change, but one point to bring up:

I think we need to make this behavior conditional? i.e. we don't want it to run in all of our pipelines do we?

dustymabe:
> Obviously I need to add the GC code as well

maybe we can work with @gursewak1997 on this over in the code he is working on https://github.com/coreos/coreos-assembler/pull/3798 - we don't need to update the same PR, but maybe let's let that land and then we'll figure out the right way to slot in GC for quay?

jlebon:
This should be configured via the pipecfg instead. See [docs/config.yaml](https://github.com/coreos/fedora-coreos-pipeline/blob/503fd4fdce235c4bfa72d07678fc00cc3f415b5b/docs/config.yaml#L133-L135).

jbtrystram:
@jlebon thanks for the review ! Updated

travier (commented):
LGTM but I'm not super familiar with some bits here so let's wait for at least Jonathan

jbtrystram:
> If we merge this but the GC side takes time to land, we'll have a tail of tags that will hang around and pollute our repo. I'm OK not blocking this on adding the GC side of it, but at least can we do this for production streams only for now?

I looked a bit at it but this config file seems to apply to all the streams, so I am not sure to implement that. 
I'll start working on the GC side of it now anyway :)

jbtrystram:
follow-up i am thinking about while working on the GC code for this:  we should include the container images published to quay in the `meta.json` file pointing to the tags created

jlebon:
> follow-up i am thinking about while working on the GC code for this: we should include the container images published to quay in the `meta.json` file pointing to the tags created

That should happen automatically. See https://github.com/coreos/coreos-assembler/blob/38235218542c39a29255933d3cea64e91754e85f/src/cmd-push-container-manifest#L128.

jbtrystram:
@jlebon i don't see knobs to adjust that per-stream in pipecfg. 
Or we merge as it is, as the GC code is actively worked on right now

jlebon:
Re. the GC side of this, just cross-posting: https://github.com/coreos/coreos-assembler/pull/3798#issuecomment-2192463923

I'm thinking the GC script should live in either cosa or https://github.com/coreos/fedora-coreos-releng-automation/... kinda leaning towards the latter to make it easier to share with other projects in the future? (Similarly to fedora-ostree-pruner there).

jlebon:
> > I'm OK not blocking this on adding the GC side of it, but at least can we do this for production streams only for now?
> 
> @jlebon i don't see knobs to adjust that per-stream in pipecfg. Or we merge as it is, as the GC code is actively worked on right now

It's not documented (you're welcome to add it to `docs/config.yaml` as part of this PR :)), but we do support it at the stream-level: https://github.com/coreos/fedora-coreos-pipeline/blob/dc58259bdd07573a6dc935c31d65213502ce8b25/utils.groovy#L460.

It's a bit awkward because I think you'd need to redefine the whole `oscontainer` object instead of having a way to just specify the additional tags, but it should be fine for now. Once we have the GC side in place, we can remove it and make it global.

>  Can we merge this ? It will make working on the GC code easier

Definitely for working on the GC side, I would suggest using your own registry repo.

jlebon (approved):
GC pieces are in cosa. We will wire it up shortly. Let's do this.

c4rt0:
Ouch, got here too late :)

PR #998, opened 2024-06-04
jobs/build-podman-os: set `deploy-via-container: true`

jlebon:
We implicitly before relied on the osbuild manifest templates to fall into the `ostree container image deploy` path for our use case just by having the `container_repo` key set. But a recent cosa change tweaked this so that now we *must* also explicitly specify `deploy-via-container`.

See also https://github.com/coreos/coreos-assembler/commit/48f11cc8cd928dbd2b2b06927c372b0a10dd1543.


c4rt0 (approved):
/lgtm

PR #1001, opened 2024-06-05
utils.groovy: escape MATRIX_WEBHOOK_URL and TOKEN

jlebon:
These are environment variables, so we need to escape it to let bash interpret it instead of Groovy.

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -715,7 +715,7 @@ def matrixSend(message) {
 
     shwrap("""
            curl -X POST -H "Content-Type: application/json" \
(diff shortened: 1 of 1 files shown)


jbtrystram:
Good catch ! But why did it worked on my tests ? Different Jenkins version ?  
Also, in the same file : https://github.com/coreos/fedora-coreos-pipeline/blob/main/utils.groovy#L335-L340

I am confused :)

jbtrystram (approved):

jbtrystram:
ah i didn't know the difference between ''' and """ ! thanks

Issue #566, opened 2022-07-18
Strengthen `build-arch` rebuild capabilities


jlebon:
Right now, if `build-arch` fails, one can *mostly* rerun it with the same parameters. There is one gotcha: if the failure happens at any point after the `Archive` step, the artifacts for the given build ID and arch will already exist.

We should strengthen the job so that it handles this window better.

jlebon:
One simple solution @dustymabe and I we came up with is to check upfront if the artifacts already exist, and if so just skip straight to the `Sign Images` step.

jlebon:
A temporary hack on this is to have a `salvage-build` job that we can run manually that just fetches the artifacts of a build, signs them, emits the compose repo import, and schedules the follo-wup kola jobs.

jschintag:
One option is to have the steps that notifies the ostree repo to be a warning instead of a failure, saving us from having to run cloud and update tests manually for the build.

PR #1003, opened 2024-06-17
jobs/fix-build: new job to fix archived builds

jlebon:
Very often, our pipeline flakes after the final archiving stage, e.g. if
the fedora-messaging queue is down, or Robosignatory is backed up.

Ideally, as discussed in #566, we could just rerun the job and have it
know to salvage the builds, but there's a lot of details there that we
need to get right. (And... it's probably not worth tackling if we're
eventually moving away from Jenkins anyway.)

For now at least, it's not hard to just have a separate job that reruns
those last final steps from the `build` and `build-arch` jobs so we can
salvage those builds.

You need to know what you're doing when using this job. It should only
be used when strictly necessary. It's often better to just do a new
build instead.

Fixes: #566


marmijo (approved):
LGTM

PR #1004, opened 2024-06-17
jobs/fix-build: support verifying signuates

jlebon:
Normally, `cosa sign` verifies that the signatures from Robosignatory are valid and then makes them public. We need to retain this check when using `fix-build` for safety (and also because we need the sigs public).

Just make this so that if `SKIP_SIGNING` is on (i.e. we know the builds were eventually signed), then pass `--verify-only` to `cosa sign` which will skip re-requesting signing.


jlebon:
Requires: https://github.com/coreos/coreos-assembler/pull/3821

jbtrystram (approved):
LGTM

PR #1005, opened 2024-06-17
jobs/fix-builds: minor fixes

jlebon:
- include version in job description
- fix passing `--arch` args to `cosa buildfetch`
- use `:hammer:` emoji instead of `:key:` for Slack message


gursewak1997 (approved):
lgtm

Issue #1002, opened 2024-06-14
Add support for excluding some tests per-stream


jlebon:
With the new `rhel-9.4` and `c9s` streams added in https://github.com/openshift/os/pull/1445, we need to be able to run kola tests but skipping those that are specific to OpenShift. Those ones have already been tagged with `openshift` ([example](https://github.com/coreos/coreos-assembler/pull/3741)). So basically, we need a stream-level knob that in practice boils down to `--tag !openshift` when we run kola.

E.g. something like:

```yaml
streams:
  rhel-9.4:
    skip_kola_tags: [openshift]
```

?

(So at least it's generic over the tag name.)

marmijo:
I opened a few PRs to solve this issue:

- https://github.com/coreos/coreos-ci-lib/pull/159
- https://github.com/coreos/fedora-coreos-pipeline/issues/1002
- Internal rhcos-art-pipcfg to skip openshift tag on c9s and rhel-9.2 streams

marmijo:
Closing as completed.

PR #1008, opened 2024-06-21
jobs/build: add support for stream level `skip_kola_tags` knob

marmijo:
Kola can now support skipping kola tags per stream \[1]. Add support for that here by modifying each Jenkins job to pass the list of kola tags to skip to the kola function.

\[1]: https://github.com/coreos/coreos-ci-lib/pull/159

See: https://github.com/coreos/fedora-coreos-pipeline/issues/1002

### DIFF
--- a/jobs/kola-azure.Jenkinsfile
+++ b/jobs/kola-azure.Jenkinsfile
@@ -135,6 +135,7 @@ cosaPod(memory: "${cosa_memory_request_mb}Mi", kvm: false,
                      build: params.VERSION, arch: params.ARCH,
                      extraArgs: params.KOLA_TESTS,
                      skipUpgrade: true,
+                     skipKolaTags: stream_info.skip_kola_tags,
                      platformArgs: """-p=azure                               \
                          --azure-credentials \${AZURE_KOLA_TESTS_CONFIG}     \
                          --azure-location $region                            \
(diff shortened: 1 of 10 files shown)


jobs/kola-azure.Jenkinsfile
```diff
@@ -138,7 +138,8 @@ cosaPod(memory: "${cosa_memory_request_mb}Mi", kvm: false,
                      platformArgs: """-p=azure                               \
                          --azure-credentials \${AZURE_KOLA_TESTS_CONFIG}     \
                          --azure-location $region                            \
-                         --azure-disk-uri /subscriptions/${azure_subscription}/resourceGroups/${azure_testing_resource_group}/providers/Microsoft.Compute/images/${azure_image_name}""")
+                         --azure-disk-uri /subscriptions/${azure_subscription}/resourceGroups/${azure_testing_resource_group}/providers/Microsoft.Compute/images/${azure_image_name}""",
+                         skipKolaTags: stream_info.skip_kola_tags)
```
jlebon, line 142: Indentation here is off.

jlebon (approved):
Small nit, but LGTM overall!

One thing I'll mention is that another way we could have done this is to extend test metadata to specify the variants against which it can run (similarly to what we have today for `distros`, or maybe even as an extension to that, e.g. `distros: [fcos, rhcos:my-variant]`).

OTOH, this is straightforward and it's not clear whether we'll need this functionality again once we've fully moved over to the new variants. At which point, we could also revert this PR and the one in coreos-ci-lib.

jobs/kola-azure.Jenkinsfile
marmijo: Fixed!

jlebon (approved):

PR #1011, opened 2024-07-03
plugins: fast-track lockable-resources to v1255

jlebon:
We're seeing
https://github.com/jenkinsci/lockable-resources-plugin/issues/550 in the RHCOS pipeline which prevents Jenkins from starting up. Let's try updating to a version which claims to fix this.

### DIFF
--- a/jenkins/controller/plugins.txt
+++ b/jenkins/controller/plugins.txt
@@ -36,3 +36,5 @@ jms-messaging:1.1.27
 # The below list are plugins that are also in base-plugins.txt but for
 # some reason or other we need to temporarily freeze or fast-track.
 #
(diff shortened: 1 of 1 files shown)


jlebon:
Let me sanity-check this depsolves cleanly.

marmijo (approved):
LGTM

jlebon:
Trying to update this plugin was going to require updating a lot of other ones. I ended up digging into the PVC and deleting the state file that was causing issues.

jlebon:
For the record, I deleted org.jenkins.plugins.lockableresources.LockableResourcesManager.xml, which is the file holding lockable resources. Since all our locks are dynamic, and we don't use build persistence, it's fine to wipe this.

PR #1012, opened 2024-07-04
jobs/sync-stream-metadata: send Slack message on failure

jlebon:
This unassuming job is actually critical to the release pipeline flow. We have checkbox items in the release checklist to make sure that it ran correctly, but for other things (e.g. reverting a rollout, or adding release notes), there's zero visibility into whether the sync actually happened unless one goes to check the job.

Let's just always send a Slack notification if the job fails.

Prompted by noticing that it failed recently because of https://github.com/coreos/fedora-coreos-streams/pull/924 only by pure chance (I was looking at the job for another reason).

Patch best viewed with whitespace ignored.

PR #1014, opened 2024-07-18
jobs/sync-stream-metadata: use jenkins service account to start pod

marmijo:
Fetching the Slace API Token seems to be failing in this job because the default fedora-coreos-pipeline service account can't access the openshift secrets directly. This is fallout from https://github.com/coreos/fedora-coreos-pipeline/pull/1012 where sending a Slack message on failure was added to this job.


aaradhak (approved):
LGTM

jlebon:
Thanks!

PR #1016, opened 2024-07-22
jobs/garbage-collection: Add Garbage Collection job to the pipeline

gursewak1997:
Added GC job which will be triggered from the release job for a given stream and will perform pruning operations based on `gc-policy.yaml` file

### DIFF
--- a/jobs/garbage-collection.Jenkinsfile
+++ b/jobs/garbage-collection.Jenkinsfile
@@ -1,0 +1,100 @@
+node {
+    checkout scm
+    // these are script global vars
+    pipeutils = load("utils.groovy")
+    pipecfg = pipeutils.load_pipecfg()
+    gc_policy_data = pipeutils.load_gc()
+    repo = "coreos/fedora-coreos-config"
+}
+
+// This job runs the Garbage collection on the selected stream in accordance with the
+// gc-policy.yaml available in https://github.com/coreos/fedora-coreos-pipeline. If
+// the pruning step succeeds, it uploads the updated builds/builds.json to s3.
+properties([
+    pipelineTriggers([]),
+    parameters([
+        choice(name: 'STREAM',
+               choices: pipeutils.get_streams_choices(pipecfg),
+               description: 'CoreOS stream to run GC'),
+        booleanParam(name: 'DRY_RUN',
+                     defaultValue: true,
+                     description: 'Only print what would be deleted')
+    ])
+])
+
+def build_description = "[${params.STREAM}]"
+def cosa_img = 'quay.io/coreos-assembler/coreos-assembler:main'
+def container_env = pipeutils.get_env_vars_for_stream(pipecfg, params.STREAM)
+def s3_stream_dir = pipeutils.get_s3_streams_dir(pipecfg, params.STREAM)
+def dry_run = params.DRY_RUN ? "--dry-run" : ""
+
+lock(resource: "gc-${params.STREAM}") {
+    cosaPod(image: cosa_img, env: container_env,
+            memory: "1024Mi",
+            serviceAccount: "jenkins") {
+        try {
+            currentBuild.description = "${build_description} Running"
+
+            stage('Init') {
+                def branch = params.STREAM
+                sh "cosa init --branch \"${branch}\" https://github.com/${repo}"
+            }
+
+            // Write YAML data to a new file in cosaPod
+            def new_gc_policy_path = 'tmp/gc-policy.yaml'
+            writeYaml file: new_gc_policy_path, data: gc_policy_data
+
+            stage('BuildFetch') {
+                pipeutils.shwrapWithAWSBuildUploadCredentials("""
+                cosa buildfetch --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG} \
+                    --stream ${params.STREAM} --url=s3://${s3_stream_dir}/builds
+                """)
+            }
+            def originalBuildsJson = readJSON file: 'builds/builds.json'
+            def originalTimestamp = originalBuildsJson.timestamp
+
+            withCredentials([file(variable: 'GCP_KOLA_TESTS_CONFIG', credentialsId: 'gcp-image-upload-config')]) {
+                def acl = pipecfg.s3.acl ?: 'public-read'
+
+                stage('Garbage Collection') {
+                    pipeutils.shwrapWithAWSBuildUploadCredentials("""
+                    cosa cloud-prune --policy ${new_gc_policy_path} \
+                    --stream ${params.STREAM} ${dry_run} \
+                    --gcp-json-key=\${GCP_KOLA_TESTS_CONFIG} \
+                    --acl=${acl} \
+                    --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG}
+                    """)
+                }
+            }
+
+            def currentBuildsJson = readJSON file: 'builds/builds.json'
+            def currentTimestamp = currentBuildsJson.timestamp
+
+            // If the timestamp on builds.json after the 'Garbage Collection' step
+            // is the same as before, that means, there were no resources to be pruned
+            // and hence, no need to update the builds.json.
+            if (originalTimestamp != currentTimestamp) {
+                // Nested lock for the Upload Builds JSON step
+                lock(resource: "builds-json-${params.STREAM}") {
+                    stage('Upload Builds JSON') {
+                        pipeutils.shwrapWithAWSBuildUploadCredentials("""
+                        cosa cloud-prune --policy ${new_gc_policy_path} \
+                        --stream ${params.STREAM} \
+                        --upload-builds-json ${dry_run} \
+                        --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG}
+                        """)
+                    }
+                }
+            }
+            currentBuild.result = 'SUCCESS'
+
+        } catch (e) {
+            currentBuild.result = 'FAILURE'
+            throw e
+        } finally {
+            if (currentBuild.result != 'SUCCESS') {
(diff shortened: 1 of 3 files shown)


jobs/garbage-collection.Jenkinsfile
```diff
@@ -0,0 +1,57 @@
...
+    credentials = [file(variable: "GCP_IMAGE_UPLOAD_CONFIG",
+                    credentialsId: "gcp-image-upload-config")]
+    // pick up the GCP project to use from the config
+    def gcp_project = shwrapCapture("jq -r .project_id \${GCP_KOLA_TESTS_CONFIG}")
+    def acl = pipecfg.s3.acl ?: 'public-read'
+
+    if (utils.credentialsExist(credentials)){
+        stage('Garbage Collection') {
+            pipeutils.shwrapWithAWSBuildUploadCredentials("""
+            cosa cloud-prune --policy ../gc-policy.yaml \
```
gursewak1997, line 31: Definitely sure this won't pickup the yaml file. Not sure how exactly we can utilise the file from the repo.

jobs/garbage-collection.Jenkinsfile
```diff
@@ -0,0 +1,57 @@
...
+            --stream ${params.STREAM} \
+            --dry-run \
```
gursewak1997, line 33: ADding `dry-run` for now till we get the initial run and we can make everything looks good. Can also add it as a parameter to the run but we would need to pass that through from the release job itself. Wasn't sure about that.

jobs/garbage-collection.Jenkinsfile
jbtrystram: Not sure if you are still blocked on this but looking at the pipeutils code the file should be in the working direcory because `piputils.load_pipecfg` already cloned the pipeline repo : 
https://github.com/coreos/fedora-coreos-pipeline/blob/main/utils.groovy#L25-L46

gursewak1997: This should be resolved now:)

jobs/garbage-collection.Jenkinsfile
gursewak1997: Yep, actually did end up adding the parameter for that reason.

jbtrystram: yeah cool I just noticed that doing a review :)

jobs/garbage-collection.Jenkinsfile
```diff
@@ -0,0 +1,91 @@
...
+                        }
+                        stage('Garbage Collection') {
+                            pipeutils.shwrapWithAWSBuildUploadCredentials("""
+                            cosa cloud-prune --policy ${new_gc_policy_path} \
+                            --stream ${params.STREAM} ${dry_run} \
+                            --gcp-project=${gcp_project} \
+                            --gcp-json-key=\${GCP_KOLA_TESTS_CONFIG} \
+                            --acl=${acl} \
+                            --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG}
+                            """)
```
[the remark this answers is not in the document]
gursewak1997: We have handled this situation pretty graciously in the `cloud-prune` script i.e. If the artifacts are already/recently deleted the deletion process would return SUCCESS.
So ideally, if we delete some resources, and don't end up uploading the new `builds.json`, it will take care of it in the next rerun.

jobs/garbage-collection.Jenkinsfile
aaradhak, line 1: Just a suggestion: It might be nice to get a description at the top of the file with an overall summary of the goal of the job and description of how the goal is achieved

jobs/garbage-collection.Jenkinsfile
```diff
@@ -0,0 +1,100 @@
...
+                        
+                        // Upload the updated builds.json back to s3 bucket
+                        stage('Upload Builds JSON') {
+                            pipeutils.shwrapWithAWSBuildUploadCredentials("""
+                            cosa cloud-prune --policy ${new_gc_policy_path} \
+                            --stream ${params.STREAM} \
+                            --upload-builds-json ${dry_run} \
+                            --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG}
+                            """)
+                        }
```
jlebon, line 85: This needs to be done under the same lock taken by `bump_builds_json()`. (Let's rename that lock there to just e.g. `builds-json-${stream}`.)

jobs/release.Jenkinsfile
```diff
@@ -273,7 +273,12 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
                 }]}
             }
         }
-
+        stage('Fork Garbage Collection') {
```
jlebon, line 276: Feels a bit odd to have this stage in the middle. I think let's put it last within this block.

jobs/garbage-collection.Jenkinsfile
```diff
@@ -0,0 +1,100 @@
...
+                                --stream ${params.STREAM} --url=s3://${s3_stream_dir}/builds
+                            """)
+                        }
+                        
+                        // Run cloud-prune to delete the resources specified in the gc-policy.yaml
+                        stage('Garbage Collection') {
+                            pipeutils.shwrapWithAWSBuildUploadCredentials("""
+                            cosa cloud-prune --policy ${new_gc_policy_path} \
+                            --stream ${params.STREAM} ${dry_run} \
+                            --gcp-project=${gcp_project} \
```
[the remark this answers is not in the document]
gursewak1997: I am not sure I understood this comment:)
Addressed all other comments though.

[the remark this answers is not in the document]
jlebon: Isn't the GCP project in the `meta.json`? So we don't need to extract it from the secret. Nor even have a switch for it.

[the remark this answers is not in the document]
gursewak1997: I see. Yes you are right. I'll update the cosa code to get the GCP project name from meta.json.

jobs/garbage-collection.Jenkinsfile
```diff
@@ -0,0 +1,89 @@
...
+// gc-policy.yaml available in https://github.com/coreos/fedora-coreos-pipeline. If
+// the pruning step succeeds, it uploads the updated builds/builds.json to s3.
+properties([
+    pipelineTriggers([]),
+    parameters([
+        choice(name: 'STREAM',
+               choices: pipeutils.get_streams_choices(pipecfg),
+               description: 'CoreOS stream to run GC'),
+        booleanParam(name: 'DRY_RUN',
+                     defaultValue: false,
```
jlebon, line 20: To start let's default to `true`?

jobs/release.Jenkinsfile
```diff
@@ -333,6 +332,12 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
                 }
             }
         }
+        stage('Fork Garbage Collection') {
+            build job: 'garbage-collection', wait: false, parameters: [
+                string(name: 'STREAM', value: params.STREAM),
+                booleanParam(name: 'DRY_RUN', value: false)
+            ]
+        }
```
jlebon, line 340: Let's hook this up in a separate PR so we can manually run the job to sanity-check it before automating it.

jobs/garbage-collection.Jenkinsfile
```diff
@@ -0,0 +1,89 @@
...
+                     description: 'Dry run Garbage Collection')
```
jlebon, line 21:
```suggestion
                     description: 'Only print what would be deleted')
```

utils.groovy
```diff
@@ -92,6 +92,12 @@ def validate_pipecfg(pipecfg, is_hotfix) {
     }
 }
 
+def load_gc() {
+    // Read YAML file
```
jlebon, line 96: This needs to follow a similar logic as `load_pipecfg()`, except that in the non-"in-tree" case, let's not clone again, but verify that the `pipecfg` directory is already there and if so, load it from there. Also shouldn't care about hotfix-related bits.

jobs/garbage-collection.Jenkinsfile
```diff
@@ -0,0 +1,89 @@
...
+        try {
+            currentBuild.description = "${build_description} Running"
+
+            stage('Init') {
+                def branch = params.STREAM
+                sh "cosa init --branch \"${branch}\" https://github.com/${repo}"
+            }
+
+            // Write YAML data to a new file in cosaPod
+            def new_gc_policy_path = './gc-policy.yaml'
```
jlebon, line 44: Let's put it under `tmp/` instead of polluting the toplevel cosa workdir.

jobs/garbage-collection.Jenkinsfile
```diff
@@ -0,0 +1,89 @@
...
+                    --stream ${params.STREAM} ${dry_run} \
+                    --gcp-json-key=\${GCP_KOLA_TESTS_CONFIG} \
+                    --acl=${acl} \
+                    --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG}
+                    """)
+                }
+            }
+
+            // Nested lock for the Upload Builds JSON step
+            lock(resource: "builds-json-${params.STREAM}") {
```
jlebon, line 69: We should just make this whole block conditional on whether it's a no-op. That way e.g. we don't lock, or show a misleading stage in the BlueOcean UI.

You can read the timestamp from builds.json before calling `cosa cloud-prune` and then compare the timestamp after.

utils.groovy
```diff
@@ -92,6 +92,24 @@ def validate_pipecfg(pipecfg, is_hotfix) {
...
+
+    if (url == 'in-tree') {
+        gc_policy_data = readYaml(file: "gc-policy.yaml")
+    } else {
+        if (fileExists('pipecfg')) {
+            gc_policy_data = readYaml(file: "pipecfg/gc-policy.yaml")
+        }
+        else {
+            error("Missing gc-policy.yaml file required for Garbage Collection")
+        }
```
jlebon, line 108:
```suggestion
        // assume the user called `load_pipecfg()` in this workdir; if not, let error out
        gc_policy_data = readYaml(file: "pipecfg/gc-policy.yaml")
```

jlebon (approved):
One minor comment, but LGTM overall!

PR #1017, opened 2024-07-26
jobs/cloud-replicate: make aliyun idemopotent

mike-nguyen:
Previously if the cloud-replicate job fails replicating to a region on Aliyun re-running the job will fail with a duplicate image error. We need to upload the meta.json on failure so the replication progress can be preserved for subsequent runs and avoid the duplicate image error.

Requires https://github.com/coreos/coreos-assembler/pull/3839 to work properly


marmijo (approved):
LGTM

mike-nguyen:
Closed in favor of https://github.com/coreos/coreos-assembler/pull/3872

PR #1019, opened 2024-08-15
release: Run garbage collection from release job

gursewak1997:
Trigger the garbage collection job during the `release` job for the respective `stream` which will run the pruning job as specified in `gc-policy.yaml`


gursewak1997:
Let's hold the merging till we run a few manual GC runs on some streams.

dustymabe (approved):
LGTM

PR #1022, opened 2024-08-20
gc-policy.yaml: Update duration for pruning for mechanical streams

gursewak1997:
Update the duration of pruning for the mechanical streams from three years to one year since we didn't start uploading AMIs till `38.20220826.91.0` and can hence use an aggressive approach for these streams. Also add `bodhi-updates` stream for the pruning

### DIFF
--- a/gc-policy.yaml
+++ b/gc-policy.yaml
@@ -1,4 +1,6 @@
 branched:
-    cloud-uploads: 3y
+    cloud-uploads: 1y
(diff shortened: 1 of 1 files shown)


jlebon (approved):
The `bodhi-updates` stream is effectively dead. Once we've fully pruned it (i.e. including the S3 bits), we can remove it from this policy file.

PR #1025, opened 2024-09-03
multi-arch-builders: Set the permission of vfio device

madhu-pillai:
To run the cex based luks encryption test, the builder user requires access to `/dev/vfio/0` device. This enable the builder user to use the cex device for luks encryption test.

### DIFF
--- a/multi-arch-builders/coreos-s390x-rhcos-builder.bu
+++ b/multi-arch-builders/coreos-s390x-rhcos-builder.bu
@@ -4,6 +4,7 @@
 # - Merge in the builder-splunk.ign Ignition file
 # - Enable the Secure Execution Host
 # - Create and initialize the secex-data volume
+# - Enable the Cex hardware based luks encryption
 #
 variant: fcos
 version: 1.4.0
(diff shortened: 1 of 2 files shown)


madhu-pillai:
Hi @jlebon ,
As we discussed earlier on the permission issue. I've added the permission in buane for the device node `/dev/vfio/0` . Kindly review.

multi-arch-builders/coreos-s390x-rhcos-builder.bu
```diff
@@ -110,4 +112,11 @@ storage:
...
-        local: create-cex-device.sh
\ No newline at end of file
+        local: create-cex-device.sh
+
+    - path: /dev/vfio/0
+      mode: 0666
+      user:
+        name: root
+      group:
+        name: root
```
jlebon, line 122: This will not do the right thing. We can't change a device node's permissions from Ignition directly.

Let's move it to `create-cex-device.sh` instead?

madhu-pillai: Will do that.

jlebon (approved):

PR #1026, opened 2024-09-03
jobs/kola-upgrade: account for overhead in x86_64 case

dustymabe:
When requesting memory for a COSA pod for x86_64 in the past (274a97a) we basically requested a pod with exactly the amount of memory we needed for the 3 virtual machines (bios,uefi,uefi-secure). This lead to a VM getting killed occasionally at first and then more consistently over time [1]. Most often this would be the uefi-secure VM most likely because that process was using more memory than the other two.

[1] https://github.com/coreos/fedora-coreos-tracker/issues/1756

### DIFF
--- a/jobs/kola-upgrade.Jenkinsfile
+++ b/jobs/kola-upgrade.Jenkinsfile
@@ -72,7 +72,8 @@ currentBuild.description = "[${params.STREAM}][${params.ARCH}] - ${start_version
 def cosa_memory_request_mb = 1024
 if (params.ARCH == 'x86_64') {
     // local (qemu+x86_64) testing will require more memory
-    cosa_memory_request_mb = 3584
+    // bios=1024, uefi=1024, uefi-secure=1536, overhead=512
+    cosa_memory_request_mb = 1024 + 1024 + 1536 + 512
 }
 
 lock(resource: "kola-upgrade-${params.ARCH}") {


jobs/kola-upgrade.Jenkinsfile
```diff
@@ -72,7 +72,8 @@ currentBuild.description = "[${params.STREAM}][${params.ARCH}] - ${start_version
 def cosa_memory_request_mb = 1024
 if (params.ARCH == 'x86_64') {
     // local (qemu+x86_64) testing will require more memory
-    cosa_memory_request_mb = 3584
+    // bios=1024, uefi=1024, uefi-secure=1024, overhead=512
```
marmijo, line 75:
```suggestion
    // bios=1024, uefi=1024, uefi-secure=1536, overhead=512
```
Just to make the math work. Did we have a 512 overhead before, or does uefi-secure require a little bit more memory?

marmijo (approved):
LGTM, with one comment

jobs/kola-upgrade.Jenkinsfile
dustymabe: basically before we didn't allow for any overhead. Each VM could use up to their provided memory, which when added up was all the pod itself was given. the kola processes and qemu processes and helpers use a little memory, not sure how much exactly, but it's some, so we'd end up using too much and something from our cgroup would need to get killed.

So we probably don't actually need 512, but it's a nice round number.

marmijo: Thanks for the clarification. The suggested change ensures the comment matches the assignment statement. The comment adds up to the original `3584`, but `cosa_memory_request_mb` is assigned `4096`.

dustymabe: ahh - I missed the comment was out of alignment

dustymabe: fixed

marmijo (approved):
LGTM

PR #1028, opened 2024-09-05
manifests/jenkins: bump capacity to 25Gi

jlebon:
This was done live in cluster, so let's make sure to update our definition so it remains the source of truth.


dustymabe (approved):
LGTM

PR #1030, opened 2024-09-09
config.yaml: Removed the CEX UUID from config.yaml

madhu-pillai:
IBM cloud does not have CEX hardware and CI test on FCOS is not possible. Hence removing the device UUID from env.


jlebon (approved):
For next time: s/removed/remove

See e.g. https://cbea.ms/git-commit/#imperative.

PR #1031, opened 2024-09-10
builder-common: Run additional prune job to free up space

c4rt0:
The Aarch64 builder consistently complains about a lack of space, particularly around 10am UTC / 12pm BST (London).
This additional prune job aims to mitigate the space issues.

See: https://github.com/openshift/os/issues/1554

### DIFF
--- a/multi-arch-builders/builder-common.bu
+++ b/multi-arch-builders/builder-common.bu
@@ -103,7 +103,8 @@ storage:
       contents:
         inline: |
             [Timer]
-            OnCalendar=*-*-* 05:00:00 UTC
+            OnCalendar=*-*-* 10:00:00 UTC
+            OnCalendar=*-*-* 22:00:00 UTC
             AccuracySec=30m
             Persistent=true
             [Install]


multi-arch-builders/builder-common.bu
```diff
@@ -104,6 +104,7 @@ storage:
         inline: |
             [Timer]
             OnCalendar=*-*-* 05:00:00 UTC
+            OnCalendar=*-*-* 10:00:00 UTC
```
jbtrystram, line 107: If we do twice a day, let's have regular intervals
```suggestion
            OnCalendar=*-*-* 17:00:00 UTC
```

jbtrystram (dismissed):

multi-arch-builders/builder-common.bu
c4rt0: Hmm... I think we hit the issue quite early in Europe, so maybe we should change the first 05:00:00 UTC to 22:00:00 instead?

jlebon:
Hmm, why not reprovision the builder with a larger disk size first? That seems like something we should do anyway, and would likely drop the need for this.

c4rt0:
> Hmm, why not reprovision the builder with a larger disk size first? That seems like something we should do anyway, and would likely drop the need for this.

I recall a conversation in which it was suggested that 200GB should suffice. I'm more than happy to drop this and open another PR if that is debunked

multi-arch-builders/builder-common.bu
dustymabe: works for me

dustymabe:
> I recall a conversation in which it was suggested that 200GB should suffice. I'm more than happy to drop this and open another PR if that is debunked

i think we can do both.

PR #1033, opened 2024-09-10
manifests: use `scheduled-upgrade` as source ImageStream tag

marmijo:
The `jenkins:2` tag hasn't been rebuilt in almost a year and the Jenkins version and plugin versions are getting out of date. The `scheduled-upgrade` and `scheduled-upgrade-redeploy` tags are rebuilt when there are changes in openshift/jenkins. By using these tags instead, we can keep our Jenkins insance more up-to-date. The `scheduled-upgrade-redeploy` tag will even automatically redeploy Jenkins when new changes are detected. This has been working in the production RHCOS pipeline for a while now.

https://docs.redhat.com/en/documentation/openshift_container_platform/4.16/html/jenkins/important-changes-to-openshift-jenkins-images#customizing-the-jenkins-image-stream-tag_important-changes-to-openshift-jenkins-images

### DIFF
--- a/manifests/jenkins-s2i.yaml
+++ b/manifests/jenkins-s2i.yaml
@@ -14,34 +14,34 @@ parameters:
     value: main
   - description: Source imagestream
     name: JENKINS_S2I_SRC_IMAGESTREAM_NAME
-    value: jenkins:2
+    value: jenkins:scheduled-upgrade-redeploy
   - description: Namespace of source imagestream
     name: JENKINS_S2I_SRC_IMAGESTREAM_NAMESPACE
     value: openshift
 
 # Here's what the flow looks like when no cert is required:
 #
-# ┌─────────────────────┐   ┌─────────────┐   ┌─────────────┐
-# │ imagestream         │   │ buildconfig │   │ imagestream │
-# │ openshift/jenkins:2 ├──►│ jenkins-s2i ├──►│ jenkins:2   │
-# └─────────────────────┘   └─────────────┘   └─────────────┘
+# ┌──────────────────────────────────────────────┐   ┌─────────────┐   ┌─────────────┐
+# │ imagestream                                  │   │ buildconfig │   │ imagestream │
+# │ openshift/jenkins:scheduled-upgrade-redeploy ├──►│ jenkins-s2i ├──►│ jenkins:2   │
+# └──────────────────────────────────────────────┘   └─────────────┘   └─────────────┘
 #
-# ┌─────────────────────────────────────┐
-# │ imagestream                         │
-# │ openshift/jenkins-agent-base:latest │
-# └─────────────────────────────────────┘
+# ┌────────────────────────────────────────────────┐
+# │ imagestream                                    │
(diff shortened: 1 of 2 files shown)


marmijo:
This will continue to work in both the RHCOS (with cert) and FCOS (no cert) cases. 

In the RHCOS case, the deploy script will overwrite the source ImageStream name to `jenkins:latest` from the rhcos namespace in `jenkins-s2i.yaml`  to use the image streams built by the other BuildConfigs.

https://github.com/coreos/fedora-coreos-pipeline/blob/b67b552663a3a8dfa63c4011add973bf2c66b65d/deploy#L58-L61

In the FCOS case, only `jenkins-s2i.yaml` is used.

manifests/jenkins-s2i.yaml
```diff
@@ -14,7 +14,7 @@ parameters:
     value: main
   - description: Source imagestream
     name: JENKINS_S2I_SRC_IMAGESTREAM_NAME
-    value: jenkins:2
+    value: jenkins:scheduled-upgrade-redeploy
```
dustymabe, line 17: does the below asciiart need to be updated?

manifests/jenkins-with-cert.yaml
```diff
@@ -63,7 +63,7 @@ objects:
         dockerStrategy:
           from:
             kind: ImageStreamTag
-            name: jenkins-agent-base:latest
+            name: jenkins-agent-base:scheduled-upgrade
```
dustymabe, line 66: I don't see `scheduled-upgrade` mentioned in the [docs](https://docs.redhat.com/en/documentation/openshift_container_platform/4.13/html/jenkins/important-changes-to-openshift-jenkins-images#customizing-the-jenkins-image-stream-tag_important-changes-to-openshift-jenkins-images) you linked but I do see `scheduled-upgrade-redeploy`. did you meant `scheduled-upgrade`?

marmijo: I did mean `scheduled-upgrade`. I'm a bit confused by it, but the `jenkins` imagestream uses [scheduled-upgrade-redeploy](https://github.com/openshift/jenkins/blob/401beedacfbffb3ff8b872c41deb87ce66c21b48/openshift/imagestreams/jenkins-rhel.yaml#L73) while the `jenkins-agent-base` imagestream uses the [scheduled-upgrade](https://github.com/openshift/jenkins/blob/83137040b1d89c1d0370c48ee7574e34d0014be4/openshift/imagestreams/jenkins-agent-base-rhel.yaml#L33) tag.

The wording in the PR might be a bit misleading. I chose to use `scheduled-upgrade` in the title to shorten it.

manifests/jenkins-s2i.yaml
marmijo: ah, thanks for catching that. Yes, I'll update that.

marmijo: Updated!

dustymabe (approved):
LGTM

PR #1035, opened 2024-09-13
config: enable next-devel 2024-09

marmijo:
Enabling next-devel for the Fedora 41 release process.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1695


dustymabe (approved):
LGTM

PR #1036, opened 2024-09-13

marmijo:
Now that `next-devel` has been enabled, we can disable the branched stream.

Wait for https://github.com/coreos/fedora-coreos-pipeline/pull/1035 before merging this.

See: https://github.com/coreos/fedora-coreos-tracker/issues/1695


dustymabe (approved):
LGTM

PR #1038, opened 2024-09-19
gc-policy.yaml: Add images and builds pruning

gursewak1997:
Update the `gc-policy.yaml` to prune the images and the whole builds itself. There's also a field for `images-kept` which we won't pruning when pruning the artifacts in meta.json's `images` key for each build.

### DIFF
--- a/gc-policy.yaml
+++ b/gc-policy.yaml
@@ -4,3 +4,6 @@ rawhide:
     cloud-uploads: 1y
 bodhi-updates:
     cloud-uploads: 1y
+    images: 58m
+    images-keep: [qemu, live-iso]
+    build: 62m
--- a/jobs/garbage-collection.Jenkinsfile
+++ b/jobs/garbage-collection.Jenkinsfile
@@ -14,7 +14,7 @@ properties([
     pipelineTriggers([]),
     parameters([
         choice(name: 'STREAM',
-               choices: pipeutils.get_streams_choices(pipecfg),
+               choices: pipeutils.get_streams_choices(pipecfg) + ['branched', 'bodhi-updates'],
                description: 'CoreOS stream to run GC'),
         booleanParam(name: 'DRY_RUN',
                      defaultValue: true,


gc-policy.yaml
```diff
@@ -1,6 +1,13 @@
...
+    images-keep: [qemu, live-iso]
+    # build: 3y
 rawhide:
     cloud-uploads: 1y
+    images: 2y
+    images-keep: [qemu, live-iso]
 bodhi-updates:
     cloud-uploads: 1y
+    images: 2y
+    images-keep: [qemu, live-iso]a
```
dustymabe, line 13:
```suggestion
    images-keep: [qemu, live-iso]
```

gursewak1997:
I am not sure about the timings for pruning the images/builds and can discuss more on that but this is the format we will have.

gursewak1997:
@jlebon @dustymabe  need your takes on this.
- Can we extend the GC to pruning the whole `builds`, since the code for it is merged. If so, what stream can we test it on and what would the time duration (How old builds do we prune) be?
- Any changes needed to `images` duration?

dustymabe:
> @jlebon @dustymabe need your takes on this.
> 
> * Can we extend the GC to pruning the whole `builds`, since the code for it is merged. If so, what stream can we test it on and what would the time duration (How old builds do we prune) be?

honestly I wouldn't test on anything but bodhi-updates initially (for any of them). While testing I would start with older times and then progressively get to the time we want (i.e. each time you can iterate on your code if you find problems). ultimately we don't care about bodhi-updates at all anymore so it's a good place to start.

> 
> * Any changes needed to `images` duration?

I'd say matching the cloud-pruning for now is good and we can look at shortening it later.

dustymabe:
IOW I would limit any changes in this PR to bodhi updates (and also consult the builds in that stream to make sure the times that we set here will actually exercise the code).

config.yaml
```diff
@@ -16,8 +16,8 @@ streams:
   #  type: mechanical
   #  env:
   #    COSA_USE_OSBUILD: true
-  # bodhi-updates:
-  #   type: mechanical
+  bodhi-updates:
+    type: mechanical
```
dustymabe, line 20: hmm. is this a requirement? I'm not actually sure new builds on this stream would succeed.

gursewak1997: afaik, we would need this to have it as a valid stream option when running the [GC](https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/blue/organizations/jenkins/garbage-collection/activity/).

gursewak1997: If everything runs okay, everything should be pruned off in the initial run for this stream and we can disable it again.

dustymabe: instead of doing this let's just [modify the job](https://github.com/coreos/fedora-coreos-pipeline/blob/b60103ffb8b66f45032ae9712df081b49507715c/jobs/garbage-collection.Jenkinsfile#L17) to hardcode `bodhi-updates` in the list of available streams to operate on for now.

I really don't want to set this because I don't want the pipeline daily trying to build bodhi updates.

gc-policy.yaml
```diff
@@ -4,3 +4,6 @@ rawhide:
     cloud-uploads: 1y
 bodhi-updates:
     cloud-uploads: 1y
+    images: 2y
+    images-keep: [qemu, live-iso]
+    build: 2y
```
dustymabe, line 9: I'd suggest we go with a different value of times here. i.e. build could be 2y and images could be 18 months? basically you want to see some builds where images was exercised but builds wasn't, right?

dustymabe (approved):
LGTM

PR #1040, opened 2024-09-20
jobs/build-podman-os.Jenkinsfile: get podman version from main

lsm5:
The current podman version is hardcoded to `5.1`. This commit will allow fetching podman from upstream main branch.

### DIFF
--- a/jobs/build-podman-os.Jenkinsfile
+++ b/jobs/build-podman-os.Jenkinsfile
@@ -38,7 +38,7 @@ def generate_diskvar_json(shortcommit, arch, artifacts, staging_repo, repo) {
             "container-tag": "${arch}-${shortcommit}",
             "extra-kargs-string": "",
             "image-type": "${artifact["platform"]}",
-            "container-imgref": "ostree-remote-registry:fedora:${repo}:5.1",
+            "container-imgref": "ostree-remote-registry:fedora:${repo}:${version}",
             "metal-image-size": "3072",
             "cloud-image-size": "10240",
             "deploy-via-container": "true",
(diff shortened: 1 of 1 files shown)


jobs/build-podman-os.Jenkinsfile
```diff
@@ -122,6 +130,24 @@ def cosa_memory_request_mb = 4.5 * 1024 as Integer
 def ncpus = ((cosa_memory_request_mb - 512) / 1536) as Integer
 
 
+node {
```
dustymabe, line 133: there is already a `node {` declaration higher up in this file.

lsm5: just me being a groovy noob :| . btw, there's another one below as well. Should I look at reusing either one or do you have a preference?

dustymabe: hmm weird. yeah, just re-use the one below

lsm5: Hmm, thinking about this further, I should grab podman version and upstream commit from the podman-next copr rpm that's fetched in the build process https://github.com/containers/podman-machine-os/blob/main/podman-image-daily/Containerfile#L31

jobs/build-podman-os.Jenkinsfile
```diff
@@ -38,7 +38,7 @@ def generate_diskvar_json(shortcommit, arch, artifacts, staging_repo, repo) {
             "container-tag": "${arch}-${shortcommit}",
             "extra-kargs-string": "",
             "image-type": "${artifact["platform"]}",
-            "container-imgref": "ostree-remote-registry:fedora:${repo}:5.1",
+            "container-imgref": "ostree-remote-registry:fedora:${repo}:${version}",
```
dustymabe, line 41: probably going to want to pass in version as an argument to this function?

dustymabe:
Closing as we no longer have this job

PR #1041, opened 2024-09-20
utils: update URLs in protected jenkins list

marmijo:
Add the new prod jenkins URL and update the namespace of the devel jenkins in the ITUP cluster. Also, remove the old devel jenkins URL as we are now fully ready to use the one in the ITUP cluster.

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -8,11 +8,11 @@ import org.yaml.snakeyaml.Yaml
 PROTECTED_JENKINSES = [
     'https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/':
         ['fcos-builds', 'prod/streams/${STREAM}', false],
-    'https://jenkins-rhcos-devel.apps.ocp-virt.prod.psi.redhat.com/':
-        ['rhcos-ci', 'prod/streams/${STREAM}', true],
-    'https://jenkins-rhcos--runtime-int.apps.int.preprod-stable-spoke1-dc-iad2.itup.redhat.com/':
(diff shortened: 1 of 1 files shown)


marmijo:
static code analysis was failing because single quotes were being used around strings containing GString expressions, so I added a commit to fix that. CI is failing now because the `STREAM` variable doesn't exist when validating the groovy syntax. Should we ignore that because `STREAM` would exist when running the pipeline?

dustymabe (approved):
LGTM

dustymabe:
> CI is failing now because the `STREAM` variable doesn't exist when validating the groovy syntax. Should we ignore that because `STREAM` would exist when running the pipeline?

I guess I don't fully understand. Is the CI that ran today on https://github.com/coreos/fedora-coreos-pipeline/pull/1038 not checking this same file? Shouldn't it have failed there too?

jlebon:
It must be single quotes. Double quotes will do string interpolation at the Groovy level, which is not what we want. The `STREAM` variable is substituted by `substituteStr`.

You can waive the failing check in the Codacy panel. It doesn't understand what's going on it sounds like.

marmijo:
> You can waive the failing check in the Codacy panel. It doesn't understand what's going on it sounds like.

Yeah, I figured there might be some sort of string substitution happening and that static code analysis thinks `${STREAM}` cant be evaluated properly.

> Is the CI that ran today on https://github.com/coreos/fedora-coreos-pipeline/pull/1038 not checking this same file?

Yes, it is! But, the file is changed in this PR causing the failure in CI.

PR #1043, opened 2024-09-27
garbage-collection: keep `container_env` as empty map

gursewak1997:
Update `container_env` as empty map to make sure we can run the garbage-collection on streams that we are disabled at the moment like `branched` and `bodhi-updates`


dustymabe:
maybe let's leave the original line in but commented out and add a comment about why we have this in temporarily

dustymabe (approved):
LGTM

Issue #1044, opened 2024-10-08
RHCOS multi-arch builder splunkforwarder container fails to build


marmijo:
The splunkfowarder container on the RHCOS multi-arch builders has been failing to build due to multiple reasons. Mainly because the internal gitlab repos being used are outdated and the version of the splunkforwarder package being installed is no longer available in the internal repos.

```
No match for argument: splunkforwarder-8.2.6
Error: Unable to find a match: splunkforwarder-8.2.6
```

We should stop building the base splunkforwarder image on the remote builders and instead just pull the image provided by IT, which has the newer splunk package already installed.

See: https://issues.redhat.com/browse/COS-2966

dustymabe:
+1

marmijo:
PR to update the butane config for the multi-arch builders to stop building the ITPAAS base splunkforwarder container and build only the splunk sidecar container: https://github.com/coreos/fedora-coreos-pipeline/pull/1045

PR #1045, opened 2024-10-08
muliti-arch-builders/splunk: only build the splunk sidecar container

marmijo:
The repo we are using to build the ITPAAS splunkforwarder container on the RHCOS multi-arch builders is out of date and references a splunkforwarder rpm that no longer exists in the internal repos. By omitting the build step for this container, we can pull in the already built image from ITPAAS that includes a more recent RPM. 
We'll still use `--arch=x86_64` for the sidecar container to ensure proper logging on non x86_64 builders.

See: https://github.com/coreos/fedora-coreos-pipeline/issues/1044

### DIFF
--- a/multi-arch-builders/builder-splunk.bu
+++ b/multi-arch-builders/builder-splunk.bu
@@ -102,14 +102,13 @@ storage:
           Type=oneshot
           RemainAfterExit=yes
           ExecStartPre=nm-online --timeout=30
-          # use `--arch=x86_64` because the RPM doesnt exist for aarch64.
-          # ppc64le and s390x RPMs are not compatible with journald input as mentioned
-          # in the splunk journald documentation for version 8.2.6
-          # https://docs.splunk.com/Documentation/Splunk/8.2.6/Data/CollecteventsfromJournalD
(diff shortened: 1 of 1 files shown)


dustymabe:
previously we weren't pulling any containers from $internal, we were accessing internal git repos but we weren't pulling any containers from internal registries. (IIUC)

are we sure we will be able to pull from those registries? one way to check is to log in to the various multi-arch builders and `podman pull` it.

marmijo:
> are we sure we will be able to pull from those registries? one way to check is to log in to the various multi-arch builders and podman pull it.

yes, we can pull from that registry. I've already restarted the splunk container on the RHCOS `aarch64` and `s390x` builders and it could `podman pull` the container just fine :+1:

dustymabe (approved):
LGTM

marmijo:
Of course I spoke too soon though. I just tried `ppc64le` and ran into 
```
fatal: unable to access 'https://gitlab.cee.redhat.com/it-cloud-openshift/splunk-sidecar.git/': SSL certificate problem: self-signed certificate in certificate chain
```

PR #1047, opened 2024-10-16
jobs: run `cosa push-container-manifest` privileged

jlebon:
In a recent 4.16 z-stream release, a cri-o backport changed the default seccomp policy to by default block `clone(CLONE_NEW*)` syscalls:

https://github.com/cri-o/cri-o/pull/8514

This affects us in the FCOS pipeline which runs in a cluster that was recently updated. The `podman manifest` commands all AIUI also flow through the default path where it wants to enter a namespace if running rootless even though we don't strictly need root; we're not running containers, just creating manifest lists. Ideally podman would be less eager there.

Anyway, work around this as necessary by running
`cosa push-container-manifest` privileged.

There are two general places where this command is used: in container image build jobs (e.g. `build-cosa`), and in the release job.

Use the multi-arch builders to do this. In the container image build jobs, this feels natural since we're already using the multi-arch builders there. In the release job, it requires cargo culting all the goop associated with remote sessions.

I initially used `cosa supermin-run` which seems like a nice fit here, but it's not in older cosa and it doesn't feel worth backporting and testing vs something we know works.

While we're here, fix the release job to respect container envs and make the Brew injection bit more DRY.

### DIFF
--- a/jobs/build-fcos-buildroot.Jenkinsfile
+++ b/jobs/build-fcos-buildroot.Jenkinsfile
@@ -146,12 +146,16 @@ lock(resource: "build-${containername}") {
                     def arch = architecture
                     images += " --image=docker://${params.CONTAINER_REGISTRY_STAGING_REPO}:${arch}-${shortcommit}"
                 }
-                shwrap("""
-                export STORAGE_DRIVER=vfs # https://github.com/coreos/fedora-coreos-pipeline/issues/723#issuecomment-1297668507
-                cosa push-container-manifest \
-                    --auth=\$REGISTRY_SECRET --tag ${gitref} \
-                    --repo ${params.CONTAINER_REGISTRY_REPO} ${images}
-                """)
+                // arbitrarily selecting the x86_64 builder; we don't run this
+                // locally because podman wants user namespacing (yes, even just
+                // to push a manifest...)
+                pipeutils.withPodmanRemoteArchBuilder(arch: "x86_64") {
+                    shwrap("""
+                        cosa push-container-manifest \
+                            --auth=\$REGISTRY_SECRET --tag ${gitref} \
+                            --repo ${params.CONTAINER_REGISTRY_REPO} ${images}
+                    """)
+                }
             }
 
             stage('Delete Intermediate Tags') {
--- a/jobs/build-kola-containers.Jenkinsfile
+++ b/jobs/build-kola-containers.Jenkinsfile
@@ -186,12 +186,16 @@ lock(resource: "build-kola-containers") {
                         images += " --image=docker://${params.CONTAINER_REGISTRY_STAGING_REPO}:${imageName}-${arch}-${shortcommit}"
                     }
 
-                    shwrap("""
-                        export STORAGE_DRIVER=vfs # https://github.com/coreos/fedora-coreos-pipeline/issues/723#issuecomment-1297668507
-                        cosa push-container-manifest --v2s2 \
-                            --auth=\$REGISTRY_SECRET --tag latest \
-                            --repo ${params.CONTAINER_REGISTRY_ORG}/${imageName} ${images}
-                    """)
+                    // arbitrarily selecting the x86_64 builder; we don't run this
+                    // locally because podman wants user namespacing (yes, even just
+                    // to push a manifest...)
+                    pipeutils.withPodmanRemoteArchBuilder(arch: "x86_64") {
+                        shwrap("""
+                            cosa push-container-manifest --v2s2 \
+                                --auth=\$REGISTRY_SECRET --tag latest \
+                                --repo ${params.CONTAINER_REGISTRY_ORG}/${imageName} ${images}
+                        """)
+                    }
                 }
             }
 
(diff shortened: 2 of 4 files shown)


jlebon:
For disclosure, I tested this in FCOS but not RHCOS. Esp. the Brew bits I did not test because I ran out of cycles. If someone can test that part and fix up whatever is necessary, that'd be greatly appreciated!

jobs/build-fcos-buildroot.Jenkinsfile
```diff
@@ -146,12 +146,16 @@ lock(resource: "build-${containername}") {
...
-                    --auth=\$REGISTRY_SECRET --tag ${gitref} \
-                    --repo ${params.CONTAINER_REGISTRY_REPO} ${images}
-                """)
+                // arbitrarily selecting the x86_64 builder; we don't run this
+                // locally because podman wants user namespacing (yes, even just
+                // to push a manifest...)
+                pipeutils.withPodmanRemoteArchBuilder(arch: "x86_64") {
+                    shwrap("""
+                        cosa push-container-manifest \
+                            --auth=\$REGISTRY_SECRET --tag ${gitref} \
```
dustymabe, line 155: I'm guessing `$REGISTRY_SECRET` here is the actual value and isn't a path to a file because I don't see where that is getting synced to the remote builder.

jobs/build-kola-containers.Jenkinsfile
```diff
@@ -186,12 +186,16 @@ lock(resource: "build-kola-containers") {
                         images += " --image=docker://${params.CONTAINER_REGISTRY_STAGING_REPO}:${imageName}-${arch}-${shortcommit}"
                     }
 
-                    shwrap("""
-                        export STORAGE_DRIVER=vfs # https://github.com/coreos/fedora-coreos-pipeline/issues/723#issuecomment-1297668507
```
dustymabe, line 190: if somehow something changed and we were able to go back to pushing in an openshift pod would we still need `STORAGE_DRIVER=vfs` ?

jobs/release.Jenkinsfile
```diff
@@ -106,6 +107,21 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
         def gcp_image = ""
         def ostree_prod_refs = [:]
 
+        // arbitrarily selecting the s390x builder; we don't run this
+        // locally because podman wants user namespacing (yes, even just
+        // to push a manifest...)
+        pipeutils.withPodmanRemoteArchBuilder(arch: 's390x') {
```
dustymabe, line 113: I know why we are using s390x here, but I do worry a little bit about there being some unintentional side effect of this.

jobs/release.Jenkinsfile
```diff
@@ -106,6 +107,21 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
+        def session = pipeutils.makeCosaRemoteSession(
+            env: container_env,
+            expiration: "30m",
```
dustymabe, line 116: any reason for choosing `30m` here? might be nice to make it at least `60m` to be able to investigate failures on the remote builder.

dustymabe:
> I initially used `cosa supermin-run` which seems like a nice fit here, but it's not in older cosa and it doesn't feel worth backporting and testing vs something we know works.


`cosa supermin-run` is what we use in `build-podman-os.Jenkinsfile` but you are right looks like it is in 4.16 and newer:

```
$ git branch --contains 4117515 | grep rhcos
  upstreamrhcos-4.16
  upstreamrhcos-4.17
```

I wonder if we could/should just use the latest COSA image in the release job anyway. I can look tomorrow to see if that would be a good idea or not.


though I do won

jobs/build-fcos-buildroot.Jenkinsfile
jlebon: Yeah it's weird. I was confused by that too but somehow it works.

dustymabe (approved):
LGTM

jobs/build-fcos-buildroot.Jenkinsfile
jlebon: OK so I think I understand this. This works in the `build-*` jobs because we're not using the remote session stuff there, only setting the `CONTAINER_HOST` bits and using the cosa from the pod. The cosa commands call podman, which takes care of proxying through `REGISTRY_AUTH_FILE`.

In the `release` job, I think it worked in testing because I initially tested with `cosa supermin-run` and it did push. And then when I tested with the remote builder approach, it no-oped on the actual push phase because nothing changed so didn't actually hit a credentials issue, but I think it would've failed. I pushed an update but will test it either later today or tomorrow morning.

jobs/build-kola-containers.Jenkinsfile
dustymabe: I know you asked about this over in https://github.com/cri-o/cri-o/pull/8514#issuecomment-2417534158 but I wonder if there is a more official place to ask and get an answer.

dustymabe:
> I pushed an update but will test it either later today or tomorrow morning.

I tested the current version of the code in https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/blue/organizations/jenkins/release/detail/release/1576/pipeline/244

Seems like it still hit an issue.

jlebon:
OK, reworked this to use `cosa supermin-run` because I don't want to figure out how to get rootless in rootless working well. Really the cleaner fix would probably be to have `podman manifest` not use any namespacing.

dustymabe:
LGTM

This succeeded in https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/blue/organizations/jenkins/release/detail/release/1577/pipeline

dustymabe (approved):

PR #1048, opened 2024-10-21
build-cosa: Add scheduled trigger to run weekly on Mondays/Wednesdays

ravanelli:
- Add a trigger for the Cosa build to execute at least twice a week (Mondays and Wednesdays);
 - This ensures regular builds even if no PRs, maintaining build continuity during periods of inactivity.

### DIFF
--- a/jobs/build-cosa.Jenkinsfile
+++ b/jobs/build-cosa.Jenkinsfile
@@ -26,7 +26,8 @@ properties([
          silentResponse: false,
          regexpFilterText: '$COREOS_ASSEMBLER_GIT_REF',
          regexpFilterExpression: 'main|rhcos-.*'
-        ]
+        ], 
+        cron('H H * * 1,3')
     ]),
     parameters([
       string(name: 'ARCHES',


ravanelli:
It needs testing, to see if both triggers can coexist

dustymabe:
> * Add a trigger for the Cosa build to execute at least twice a week (Mondays and Wednesdays);

once a week would probably be sufficient if we wanted to simplify it. I would just schedule it for Monday morning, but fine with it the way you have it too.

dustymabe (approved):
LGTM

dustymabe:
I guess one thing we should also make sure of is that older `rhcos-4.XX` branches don't get triggered this way.

ravanelli:
> I guess one thing we should also make sure of is that older `rhcos-4.XX` branches don't get triggered this way.
I will try to test it first before we merge it!

jlebon:
> I guess one thing we should also make sure of is that older `rhcos-4.XX` branches don't get triggered this way.

As written, it wouldn't. It would only trigger builds with defaults set for all the parameters (which are set to build from `main`).

ravanelli:
@jlebon @dustymabe are we ok to merge it? 
I did a test, and only got triggered.

dustymabe (approved):
LGTM

jobs/build-cosa.Jenkinsfile
```diff
@@ -26,7 +26,8 @@ properties([
          silentResponse: false,
          regexpFilterText: '$COREOS_ASSEMBLER_GIT_REF',
          regexpFilterExpression: 'main|rhcos-.*'
-        ]
+        ], 
+        cron('H H * * 1,3')
```
c4rt0, line 30: Sorry for a silly question, but isn't cron format ('M H * * 1,3') ?

c4rt0: Ok, sorry for the noise. I looked it up and in Jenkins it specifies a range.

/lgtm

c4rt0 (approved):
/lgtm

PR #1049, opened 2024-10-24
Add steps to create and configure a versioned S3 bucket

gursewak1997:
Add tutorial with detailed steps for creating and configuring a versioned S3 bucket, including lifecycle rules for managing versions and delete markers.

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -242,6 +242,8 @@ If you want to store builds persistently, now is a good time to allocate
 S3 storage.  See the [upstream coreos-assembler docs](https://github.com/coreos/coreos-assembler/blob/main/README-design.md)
 around build architecture.
 
+If you need to set up versioning and lifecycle management for the bucket, refer to [s3 bucket versioning doc](https://github.com/coreos/fedora-coreos-pipeline/blob/main/docs/s3-bucket-versioning.md) for detailed steps. For FCOS production builds using the `fcos-builds` bucket we apply a lifecycle policy of 14 days to manage old versions efficiently.
+
 Today, the FCOS pipeline is oriented towards having its own
 bucket; this will likely be fixed in the future.  But using your
 credentials, you should now do e.g.:
--- a/docs/s3-bucket-versioning.md
+++ b/docs/s3-bucket-versioning.md
@@ -1,0 +1,85 @@
+# **Steps to Create and Configure a Versioned S3 Bucket**
+
+This guide provides step-by-step instructions to create and configure an Amazon S3 bucket with versioning and lifecycle management to optimize storage and reduce costs. Versioning in S3 allows you to maintain multiple versions of an object. This feature is particularly useful in scenarios of:
+
+* **Accidental Overwrites**: If a file is accidentally overwritten, you can restore a previous version without data loss.  
+* **Accidental Deletions**: When an object is deleted, S3 creates a delete marker instead of immediately removing the object, enabling easy recovery of previous versions.  
+* **Garbage Collection and Storage Optimization**: Over time, old versions, non-current versions, and delete markers accumulate, taking up storage. Combining versioning with **lifecycle policies** automates the cleanup of outdated versions and unnecessary delete markers. This ensures your bucket stays organized and prevents unneeded storage costs from piling up.
+
+---
+
+## **Set a Variable for the Bucket Name**
+You can use the name of an existing bucket or specify a new bucket name if you are creating one:
+```bash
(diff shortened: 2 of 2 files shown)


docs/s3-bucket-versioning.md
```diff
@@ -0,0 +1,78 @@
+# **Steps to Create and Configure a Versioned S3 Bucket**
+
```
ravanelli, line 2: Would be good to have a summary about what this doc is about here:
Something like:

This file provides a step-by-step guide for creating and configuring an Amazon S3 bucket with versioning and lifecycle management to optimize storage usage and costs.

docs/s3-bucket-versioning.md
```diff
@@ -0,0 +1,78 @@
...
+## **Create a New S3 Bucket**
+Use the following command to create a bucket. Make sure to specify the correct region.
+
+```bash
+aws s3api create-bucket --bucket $BUCKET_NAME --region your-region
+```
+
+---
+
+## **Enable Versioning on the Bucket**
```
ravanelli, line 17: How can I see/track the changes/versions?
When should I change the version?
Would be good to add a section about it too.

docs/s3-bucket-versioning.md
```diff
@@ -0,0 +1,78 @@
+# **Steps to Create and Configure a Versioned S3 Bucket**
+
+## **Set a Variable for the Bucket Name**
+```bash
+BUCKET_NAME=temp-bucket-for-versioning
```
ravanelli, line 5: Is that the standard name we use for our bucket?
What is the diff for a normal bucket and this one?

gursewak1997: No there's no standard names for the buckets afaik. This is just an example for a name. I'll update the doc specifying the user to pick a name according to the needs.

docs/s3-bucket-versioning.md
gursewak1997: To track changes/versions, I have added a brief section down in **Verify Configuration**
Version changes happen automatically with each update or delete marker, so manual version changes aren't needed.

dustymabe:
Nice work! I like the explanations.

Maybe here let's also update https://github.com/coreos/fedora-coreos-pipeline/blob/main/HACKING.md#optional-allocating-s3-storage to mention "if you want to version and lifecycle the bucket see this doc for how to do so". and also mention that for FCOS fcos-builds we'll use X days as the lifecycle.

I'm thinking X of 14 days would be appropriate? 2 weeks is easy to remember and should be plenty of time I would think.

gursewak1997:
> I'm thinking X of 14 days would be appropriate? 2 weeks is easy to remember and should be plenty of time I would think.

2 weeks should be ample time to notice any discrepancy in `builds.json` for example. Updated.

HACKING.md
```diff
@@ -242,6 +242,8 @@ If you want to store builds persistently, now is a good time to allocate
 S3 storage.  See the [upstream coreos-assembler docs](https://github.com/coreos/coreos-assembler/blob/main/README-design.md)
 around build architecture.
 
+If you need to set up versioning and lifecycle management for the bucket, refer to [s3 bucket versioning doc](https://github.com/coreos/fedora-coreos-pipeline/blob/main/docs/s3-bucket-versioning.md) for detailed steps. For FCOS builds, we’ll apply a lifecycle policy of 14 days to manage old versions efficiently.
```
dustymabe, line 245: minor improvement:

```suggestion
If you need to set up versioning and lifecycle management for the bucket, refer to [s3 bucket versioning doc](https://github.com/coreos/fedora-coreos-pipeline/blob/main/docs/s3-bucket-versioning.md) for detailed steps. For FCOS production builds using the `fcos-builds` bucket we apply a lifecycle policy of 14 days to manage old versions efficiently.
```

HACKING.md
```diff
@@ -242,6 +242,8 @@ If you want to store builds persistently, now is a good time to allocate
...
+
```
dustymabe, line 246: should we update the `aws s3 mb` example below to use the new API? `aws s3api` ?

gursewak1997: Obviously, the existing command should work but I am happy to update to new API in another commit.

dustymabe: nvm - don't worry about this

gursewak1997: Hah, added the changes. Lmk they seem worth the change :eyes:

ravanelli (approved):
LGTM!

dustymabe (approved):

PR #1050, opened 2024-10-28
config.yaml: disable next-devel 2024-10-28

dustymabe:
Now that testing-devel is moving back over to F41 we don't have any differences in testing-devel and next-devel so we'll disable it.


aaradhak (approved):
LGTM

PR #1052, opened 2024-11-01
Garbage Collection: Update build description and add streams

gursewak1997:
Let's add the `dry_run` switch to the `build_description` for the garbage collection run. Also, expand the GC to `testing-devel` and `next-devel` for `cloud_uploads`.


dustymabe (approved):
LGTM

PR #1055, opened 2024-11-13
support conditionally building artifacts in an OSBuild stage

dustymabe:
With https://github.com/coreos/coreos-assembler/pull/3930 we can now build multiple artifacts/platforms in a single OSBuild call. Let's add support here for detecting what artifacts are supported to be built by OSBuild and build them using OSBuild.

We segregate here "experimental" versus "stable" OSbuild built artifacts and add a pipecfg knob for opting into the "experimental" ones being built by OSBuild.

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -427,6 +427,42 @@ def build_artifacts(pipecfg, stream, basearch) {
     // First get the list of artifacts to build from the config
     def artifacts = get_artifacts_to_build(pipecfg, stream, basearch)
 
+    // If `cosa osbuild` is supported then let's build what we can using OSBuild
+    if (shwrapRc("cosa shell -- test -e /usr/lib/coreos-assembler/cmd-osbuild") == 0) {
+        // Determine which platforms are OSBuild experimental versus
+        // stable (i.e. the default is to use OSBuild for them).
+        def experimental = shwrapCapture("cosa osbuild --supported-platforms").tokenize()
+        def stable = shwrapCapture('''
+            cosa shell -- bash -c '
+                for buildextend in /usr/lib/coreos-assembler/cmd-buildextend-*; do
+                    if [ "$(readlink -f ${buildextend})" == "/usr/lib/coreos-assembler/cmd-osbuild" ]; then
+                        # the 42 here chops off /usr/lib/coreos-assembler/cmd-buildextend-
+                        echo "${buildextend:42}"
+                    fi
+                done
+            '
+        ''').tokenize('\n')
+        // Based on the pipeline configuration we'll either use OSBuild for as
+        // much as we can (experimental) or just what it is the default for (stable)
+        def osbuild_supported_artifacts = stable
+        if (pipecfg.streams[stream].osbuild_experimental) {
+            osbuild_supported_artifacts = experimental
+        }
+        // Let's build separately the artifacts that can be built directly with OSBuild.
+        def osbuild_artifacts = []
+        for (artifact in artifacts) {
+            if (artifact in osbuild_supported_artifacts) {
+                osbuild_artifacts += artifact
+            }
+        }
+        if (!osbuild_artifacts.isEmpty()) {
+            artifacts.removeAll(osbuild_artifacts)
+            stage('💽:OSBuild') {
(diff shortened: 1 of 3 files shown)


dustymabe:
unfortunately building more in OSBuild is causing us to need a larger cache qcow:

https://github.com/coreos/coreos-assembler/pull/3942

dustymabe:
pushing up a new commit to see if we can make the linter happy - if it works I'll test it and squash.

dustymabe:
> pushing up a new commit to see if we can make the linter happy - if it works I'll test it and squash.

I don't think making the linter happy is going to happen. There are [plenty of other similar errors in the same file](https://app.codacy.com/gh/coreos/fedora-coreos-pipeline/files/130437888968/issues). I'm going to ignore since I already tested this code.

ravanelli (approved):
lgtm

utils.groovy
```diff
@@ -427,6 +427,41 @@ def build_artifacts(pipecfg, stream, basearch) {
...
+    // If `cosa osbuild` is supported then let's build what we can using OSBuild
+    if (shwrapRc("cosa shell -- test -e /usr/lib/coreos-assembler/cmd-osbuild") == 0) {
+        // Determine which platforms are OSBuild experimental versus
+        // stable (i.e. the default is to use OSBuild for them).
+        def experimental = shwrapCapture("cosa osbuild --supported-platforms").tokenize()
+        def stable = shwrapCapture('''
+            cosa shell -- bash -c '
+                for buildextend in /usr/lib/coreos-assembler/cmd-buildextend-*; do
+                    if [ "$(readlink -f ${buildextend})" == "/usr/lib/coreos-assembler/cmd-osbuild" ]; then
+                        echo "${buildextend:42}"
```
jlebon, line 439: That 42 is worth a comment. Or alternatively, I sometimes make it more self-documenting by putting the prefix into a variable, and then you can use it in the iteration as well as here via e.g. `${#prefix}`.

utils.groovy
```diff
@@ -427,6 +427,41 @@ def build_artifacts(pipecfg, stream, basearch) {
...
+                    fi
+                done
+            '
+        ''').tokenize('\n')
```
jlebon, line 443: Hmm, I wonder if we should fold this itself into `cosa osbuild`. E.g. we could have `cosa osbuild --supported-platforms=(all|experimental|stable)`.

jlebon (approved):
Some comments, but LGTM overall!

So is the idea here that the subsequent stages that build artifacts will still run but will no-op?

utils.groovy
dustymabe: I think I would if we planned this code to live long.. but I'm hoping to delete it in 6 months or less.

utils.groovy
dustymabe: to prevent re-testing in CI I'll just add a comment.

dustymabe:
> So is the idea here that the subsequent stages that build artifacts will still run but will no-op?

The platforms get removed from the list of artifacts to parallel buildextend on when they get built by OSBuild. So, no. The idea is that there isn't a no-op stage that happens later, however in practice right now both `metal` and `metal4k` will both run and get a no-op because the later code in this function force puts them back in when `live` is requested. That behavior will stop once we start building `live` using OSBuild too.

dustymabe:
just reporting some results now that we are building applehv gcp and hyperv using OSBuild in `rawhide`:

```
$ python3 diffbuildsizes.py 42.20241114.91.0 42.20241115.91.0
    platform     | compression  | 42.20241114.91.0 | 42.20241115.91.0
---------------------------------------------------------------------
...
applehv          | compressed   | 880MiB           | 881MiB          
applehv          | uncompressed | 10240MiB         | N/A             
...
hyperv           | compressed   | 872MiB           | 873MiB          
hyperv           | uncompressed | 1960MiB          | N/A             
...
gcp              | compressed   | 872MiB           | 881MiB          
gcp              | uncompressed | 1721MiB          | N/A             
```

Note the uncompressed size doesn't show up in the OSBuild releases
because we determined having the uncompressed size and sha256 weren't
important enough to perform open heart surgery on OSBuild to compute
that information.

PR #1057, opened 2024-11-15
jobs: consistently discard old builds

jlebon:
Most of our jobs enable build discard, but not all. Turn it on consistently across the board. Use the established setting of 100 for most of them, except for the garbage-collection and release jobs whose logs have more value.


jlebon:
> maybe for build-arch we should adjust it higher since for every build 3 buildarch jobs get called.

Though one difference is that `build` can no-op and so of the 100 we keep there, not all actually result in 3 `build-arch` jobs. I guess let's leave it at 100 and if we feel like we need more history there, we bump it?

PR #1058, opened 2024-11-18
Add containers Garbage Collection.

gursewak1997:
Rename `cloud-prune` to `coreos-prune` since we updated the script in cosa to `cmd-coreos-prune`. Update params to gc to take containers registry secret. Also, update the `gc-policy.yaml` to start containers pruning.

### DIFF
--- a/jobs/garbage-collection.Jenkinsfile
+++ b/jobs/garbage-collection.Jenkinsfile
@@ -56,13 +56,18 @@ lock(resource: "gc-${params.STREAM}") {
             def originalTimestamp = originalBuildsJson.timestamp
             def acl = pipecfg.s3.acl ?: 'public-read'
 
-            withCredentials([file(variable: 'GCP_KOLA_TESTS_CONFIG', credentialsId: 'gcp-image-upload-config')]) {
+            withCredentials([
+                file(variable: 'GCP_KOLA_TESTS_CONFIG', credentialsId: 'gcp-image-upload-config'),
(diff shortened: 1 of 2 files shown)


jobs/garbage-collection.Jenkinsfile
```diff
@@ -56,13 +56,17 @@ lock(resource: "gc-${params.STREAM}") {
...
+                file(variable: 'REGISTRY_SECRET', credentialsId: 'oscontainer-push-registry-secret')
+            ]) {
                 stage('Garbage Collection') {
                     pipeutils.shwrapWithAWSBuildUploadCredentials("""
                     cosa coreos-prune --policy ${new_gc_policy_path} \
                     --stream ${params.STREAM} ${dry_run} \
                     --gcp-json-key=\${GCP_KOLA_TESTS_CONFIG} \
                     --acl=${acl} \
+                    --registry-auth-file=\${REGISTRY_SECRET} \
                     --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG}
```
dustymabe, line 70: This isn't new in this PR, but I noticed it..

It's a bit odd to pass some secrets one way and some another.. maybe we should consider instead of using `shwrapWithAWSBuildUploadCredentials()` here to just explicitly mention the `AWS_BUILD_UPLOAD_CONFIG` one in the block above too.

dustymabe (approved):
LGTM, though I think 2w is a little aggressive here these are our development/mechanical streams and it's probably OK. If we may find it useful to have more of them hang around longer in the future we can just adjust the policy here in a new PR.

PR #1059, opened 2024-11-20
jobs/gc: Change the credential for containers deletion

gursewak1997:
We need `cosa-push-registry-secret` to delete quay containers. Updated it from `oscontainer-push-registry-secret`.


marmijo (approved):
LGTM

Issue #1061, opened 2024-11-22
NetworkNotFound error preventing server creation in VexxHost OpenStack


marmijo:
## Description
A `NetworkNotFound` error was seen in the FCOS OpenStack instance on VexxHost  today. The error caused all server creation to fail in OpenStack, either in the `kola-openstack` job or locally through the CLI. 

```
harness.go:1782: Cluster failed starting machines: waiting for instance to run: 
Server reported ERROR status: 
{500 2024-11-22 20:56:47 +0000 UTC  Build of instance 8837696b-6172-4721-aa4a-729e576573d5 aborted: Failed to allocate the network(s), not rescheduling.
```

The issue seems to be that the `private` network we attach to the servers is not available, but the network looks fine in the CLI and the cloud console.

When creating a server using `openstack server create --debug --network=private <other server creation arguments>` , the following debug message can be seen:

```
RESP BODY: {"NeutronError": {"type": "NetworkNotFound", "message": "Network private could not be found.", "detail": ""}}
```

The instance fails to launch and the error shows that the private network cannot be found. However:

- The private network exists and appears healthy (`openstack network show private` confirms this, as does the cloud console).
- The network has a valid subnet and sufficient IP addresses.
- Instances launched yesterday (2024-11-21) using this network were functional.

### OpenStack region doesn't seem to make a difference
The failure was seen using the `ca-ymq-1` region in OpenStack, but I also saw the error when I tried creating a server in `ams1` as well.

### Timing of failure
We saw a successful kola-openstack run on `2024-11-22 8:54 UTC`, but then saw the following error in a kola-openstack run at `2024-11-22 8:13 UTC`
```
[2024-11-22T08:11:42.185Z] + ore openstack --config-file=**** --region=ca-ymq-1 create-image --file=/home/jenkins/agent/workspace/kola-openstack/builds/41.20241119.20.1/aarch64/fedora-coreos-41.20241119.20.1-openstack.aarch64.qcow2 --name=kola-fedora-coreos-testing-devel-aarch64 --arch=aarch64
[2024-11-22T08:12:50.015Z] Couldn't create image: creating image: Expected HTTP response code [201] when accessing [POST https://image.public.mtl1.vexxhost.net/v2/images], but got 504 instead
[2024-11-22T08:12:50.015Z] <html>
[2024-11-22T08:12:50.015Z] <head><title>504 Gateway Time-out</title></head>
[2024-11-22T08:12:50.015Z] <body>
[...]

PR #1062, opened 2024-11-22
utils: temporarily disable kola-openstack in the pipeline

marmijo:
All openstack runs started failing today due to https://github.com/coreos/fedora-coreos-pipeline/issues/1061. 

Let's disable the mechanism to kick off openstack jobs for a few days to see if the issue is resolved after the weekend.


dustymabe:
looking at the code now it looks like another way to do this is to empty out the list in: 

https://github.com/coreos/fedora-coreos-pipeline/blob/2a109e590d261763077f5386f4cdbdcd65982112/config.yaml#L113

marmijo:
This should do the exact same thing since we don't run this test on RHCOS, but the manifest is probably a better place for this anyway. I'll make the change.

dustymabe:
actually you might have to do something like `['none']` here

I'm not sure if the code will go into the `if` if the list is empty (i.e. does an empty list evaluate to false in groovy)

aaradhak (approved):

marmijo:
> I'm not sure if the code will go into the if if the list is empty (i.e. does an empty list evaluate to false in groovy)

It should be fine as `test_architectures: []` : https://www.tutorialspoint.com/groovy/groovy_isempty.htm

marmijo:
/unhold
Yes, groovy will evaluate an empty list as false. I amended the commit to be `test_architectures = ['none']` so the list will evaluate to true, but have no matching arches.

dustymabe (approved):
LGTM

PR #1063, opened 2024-11-25
Rework build-mechanical Jenkins job and add  `scheduled: true` config knob

marmijo:
Changes include: 
- Rework build-mechanical Jenkins Job:
  - Change the logic of build-mechanical to continue to build all mechanical streams every 24 hours with the conditions set on the `scheduled: true` knob:
    - if any mechanical stream has the knob set, then only trigger those streams
    - if no mechanical stream has the knob set, then trigger all mechanical streams 
  - This will be used to build the rhel-9.6 and c9s streams in the RHCOS pipeline on a schedule. 
- `scheduled: true` knob:
  - used to flag streams that will be built on a schedule using the new `build-scheduled` Jenkins job. Specifically, this knob will be used to build two new RHCOS streams every two days.
- utils.groovy `scheduled_streams` method:
  - From a provided list, returns a subset of streams  with the `scheduled: true` knob set The method is used by the build-mechanical Jenkins job to build only streams with the knob set to true.

See: https://issues.redhat.com/browse/COS-3012

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -101,6 +101,11 @@ streams:
       # OPTIONAL: allow building using OSBuild for artifacts where OSBuild is
       # not yet the default in CoreOS Assembler.
       osbuild_experimental: true
+      # OPTIONAL: stream will be set as "scheduled"
+      # By default, `build-mechanical` builds all mechanical streams.
+      # When mechanical streams have `scheduled` on, then
+      # `build-mechanical` will only build those streams
+      scheduled: true
 
 # REQUIRED: architectures to build for other than x86_64
 additional_arches: [aarch64, ppc64le, s390x]
--- a/jobs/build-mechanical.Jenkinsfile
+++ b/jobs/build-mechanical.Jenkinsfile
@@ -19,28 +19,16 @@ properties([
 
 node {
     def mechanical_streams = pipeutils.streams_of_type(pipecfg, 'mechanical')
+    def scheduled_streams = pipeutils.scheduled_streams(pipecfg, mechanical_streams)
+    if (scheduled_streams) {
+        mechanical_streams = scheduled_streams
+    }
 
-    change = checkout(
-        [$class: 'GitSCM',
-         userRemoteConfigs: [[url: pipecfg.source_config.url]],
-         branches: pipeutils.streams_as_branches(mechanical_streams)
+    mechanical_streams.each{
+        echo "Triggering build for mechanical stream: ${it}"
+        build job: 'build', wait: false, parameters: [
+          string(name: 'STREAM', value: it),
+          booleanParam(name: 'EARLY_ARCH_JOBS', value: false)
         ]
-    )
-
-    if (pipeutils.triggered_by_push()) {
-        stream = pipeutils.stream_from_branch(change.GIT_BRANCH)
-        if (stream in mechanical_streams) {
-            build job: 'build', wait: false, parameters: [
-              string(name: 'STREAM', value: stream)
-            ]
-        }
-    } else {
-        // cron or manual build: build all mechanical streams
-        mechanical_streams.each{
-            build job: 'build', wait: false, parameters: [
-              string(name: 'STREAM', value: it),
-              booleanParam(name: 'EARLY_ARCH_JOBS', value: false)
-            ]
-        }
     }
 }
--- a/utils.groovy
+++ b/utils.groovy
@@ -384,6 +384,12 @@ def streams_of_type(config, type) {
     return config.streams.findAll{k, v -> v.type == type}.collect{k, v -> k}
 }
 
+// Returns a list of stream names from `streams_subset` that have `scheduled: true` set
+def scheduled_streams(config, streams_subset) {
+    return streams_subset.findAll{stream ->
+        config.streams[stream].scheduled}.collect{k, v -> k}
+}
+
 def get_streams_choices(config) {
     def default_stream = config.streams.find{k, v -> v['default'] == true}?.key
     def other_streams = config.streams.keySet().minus(default_stream) as List


marmijo:
Static code analysis is marking 3 minor issues for no whitespace around the opening and closing braces for the closures in this line:
https://github.com/coreos/fedora-coreos-pipeline/blob/f087c1995b9c87dd2980416740a01379af96c464/utils.groovy#L387-L389

However, I matched the code style of the another method in the same file:
https://github.com/coreos/fedora-coreos-pipeline/blob/f087c1995b9c87dd2980416740a01379af96c464/utils.groovy#L383-L385

I think we can ignore the Codacy Static Code Analysis warning.

jobs/build-scheduled.Jenkinsfile
```diff
@@ -0,0 +1,33 @@
...
+        artifactNumToKeepStr: '100'
+    )),
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+])
+
+node {
+
+    def scheduled_streams = pipeutils.scheduled_streams(pipecfg)
+
+    // unlink build-mechanical, only build the streams
```
jlebon, line 24: unlike?

jlebon (commented):
Hmm, OK sorry. You did _exactly_ what was written in the proposal. The code looks clean and simple, and I'd be fine merging it in.

That said, I thought `build-mechanical` was currently being triggered by both timers and git pushes, but that's actually not the case; it's only triggered by the timer. So this then makes the difference between `build-mechanical` and `build-scheduled` quite subtle.

I'm thinking, maybe instead it'd be cleaner if we _did_ just enable `build-mechanical` in the RHCOS pipeline instead. But then `scheduled: true` would tell `build-mechanical` to only trigger those mechanical streams, instead of all of them.

WDYT?

jobs/build-scheduled.Jenkinsfile
marmijo: I just updated this!

jlebon:
> That said, I thought `build-mechanical` was currently being triggered by both timers and git pushes, but that's actually not the case; it's only triggered by the timer.

I.e. I think this whole chunk of the `build-mechanical` job is no longer relevant:

https://github.com/coreos/fedora-coreos-pipeline/blob/b3771b252888c5cb04c3bb788b7a6e533d735fa2/jobs/build-mechanical.Jenkinsfile#L23C5-L37

and dates from a time when it _was_ triggered by git pushes, but then that was severed when we migrated out of the OpenShift Build Jenkins build strategy goop.

marmijo:
> That said, I thought build-mechanical was currently being triggered by both timers and git pushes, but that's actually not the case; it's only triggered by the timer.
....
and dates from a time when it was triggered by git pushes, but then that was severed when we migrated out of the OpenShift Build Jenkins build strategy goop.

I thought this was the case - we used to have it trigger by push as well, but I think it was bogging down the pipeline on development heavy days. It would be fine to rework build-mechanical to work in this case too but that job tiggers every 24 hours. Are we satisfied with building the new streams that frequently?

marmijo:
I'll look into how we can build only the `scheduled: true` streams every 48h, while the rest of the streams are built every 24h using the same job.

jlebon:
> It would be fine to rework build-mechanical to work in this case too but that job tiggers every 24 hours. Are we satisfied with building the new streams that frequently?

Ahh yes, good catch. I don't think we should somehow try to support different cron settings per stream here. But it could make sense to have a top-level e.g. `build-mechanical-cron` config knob so that it's a different setting in FCOS vs RHCOS. Though overall, I'm inclined to not worry about this for now and just leave it as 24h. We can add it in a follow-up if we feel it's worth it.

marmijo:
> Though overall, I'm inclined to not worry about this for now and just leave it as 24h.

I'm fine with that as well. 

We have the `type: mechanical` field in the RHCOS config for all streams. If we enable `build-mechanical` wont it trigger those every 24 hours, in addition to the triggers from ART? If that's the case, the easiest solution would probably be to add the `build-scheduled` job, or remove the `type: mechanical` field from the RHCOS config (though I'm not sure what other implications that might have).

jlebon:
> > Though overall, I'm inclined to not worry about this for now and just leave it as 24h.
> 
> I'm fine with that as well.
> 
> We have the `type: mechanical` field in the RHCOS config for all streams. If we enable `build-mechanical` wont it trigger those every 24 hours, in addition to the triggers from ART? 

No, that's what I mean with:

> But then `scheduled: true` would tell `build-mechanical` to only trigger those mechanical streams, instead of all of them.

The alternative which has slightly easier implementation logic is the opposite polarity, e.g. `skip-scheduled: true` for which then `build-mechanical` would trigger _all_ mechanical streams except those with `skip-scheduled`, but since most of our streams are ART-driven, it would require adding that to almost everything, which is annoying. So I'm thinking, just leaving it as `scheduled: true` seems cleaner.

So then the logic would be:
- if any stream has `scheduled: true`, then only trigger mechanical streams that have it
- if no stream has `scheduled: true`, then trigger all mechanical streams

marmijo:
> ...it would require adding that to almost everything, which is annoying. So I'm thinking, just leaving it as scheduled: true seems cleaner.

I agree, let's leave it as `scheduled: true`

> So then the logic would be:
if any stream has scheduled: true, then only trigger mechanical streams that have it
if no stream has scheduled: true, then trigger all mechanical streams

Yeah, this sounds good to me. I'll rework the PR to make it happen.

marmijo:
I updated this PR with the changes.

> So then the logic would be:
>
> if any stream has scheduled: true, then only trigger mechanical streams that have it
> if no stream has scheduled: true, then trigger all mechanical streams

I changed the conditions a little bit to:

- if any **mechanical** stream has `scheduled: true`, then only trigger those mechanical streams that have it
- if no **mechanical** stream has scheduled: true, then trigger all mechanical streams

Essentially, `scheduled: true` should only be used on mechanical streams. Meaning, of the set of mechanical streams, there's a subset of scheduled streams (or not).

WDYT?

docs/config.yaml
```diff
@@ -101,6 +101,10 @@ streams:
       # OPTIONAL: allow building using OSBuild for artifacts where OSBuild is
       # not yet the default in CoreOS Assembler.
       osbuild_experimental: true
+      # OPTIONAL: stream will be set as "scheduled"
+      # If using, `type: mechanical` must also be set on the stream
+      # This is used for some RHCOS streams to be built every day by the build-mechanical job
```
jlebon, line 106: Maybe we should document the semantics here. E.g.

```suggestion
      # OPTIONAL: stream will be set as "scheduled"
      # By default, `build-mechanical` builds all mechanical streams. When mechanical
      # streams have `scheduled` on, then `build-mechanical` will only build those streams.
```

?

jobs/build-mechanical.Jenkinsfile
```diff
@@ -19,28 +19,30 @@ properties([
...
-        stream = pipeutils.stream_from_branch(change.GIT_BRANCH)
-        if (stream in mechanical_streams) {
-            build job: 'build', wait: false, parameters: [
-              string(name: 'STREAM', value: stream)
-            ]
-        }
-    } else {
-        // cron or manual build: build all mechanical streams
-        mechanical_streams.each{
+    if (scheduled_streams) {
```
jlebon, line 24: Instead of this, I think we could do something like:

```groovy
def mechanical_streams = ...
def scheduled_streams = ...
if (scheduled_streams) {
    mechanical_streams = scheduled_streams
}

mechanical_streams.each{
    ...
}
```

jobs/build-mechanical.Jenkinsfile
```diff
@@ -19,28 +19,30 @@ properties([
...
               string(name: 'STREAM', value: it),
               booleanParam(name: 'EARLY_ARCH_JOBS', value: false)
             ]
         }
+    } else {
+        // only trigger all mechanical FCOS streams. There is another mechanism to trigger
+        // these streams for RHCOS. Since the brew section of the config is required for
+        // RHCOS, use it to decide which product we're building.
+        def is_rhcos = pipecfg.brew != null
+        if (!is_rhcos) {
```
jlebon, line 38: Hmm, there's no need for this. The proposed logic should work fine for both FCOS and RHCOS. In general, the pipeline shouldn't know about RHCOS vs FCOS.

utils.groovy
```diff
@@ -384,6 +384,12 @@ def streams_of_type(config, type) {
     return config.streams.findAll{k, v -> v.type == type}.collect{k, v -> k}
 }
 
+// Returns a list of stream names from `streams_subset` that have `scheduled: true` set
+def scheduled_streams(config, streams_subset) {
+    return streams_subset.findAll{stream ->
+        config.streams[stream].scheduled == true}.collect{k, v -> k}
```
jlebon, line 390:
```suggestion
        config.streams[stream].scheduled}.collect{k, v -> k}
```

jlebon (commented):
> I changed the conditions a little bit to:

Yeah sorry, that's exactly what I meant! :+1:

docs/config.yaml
marmijo: WFM!

utils.groovy
marmijo: Thanks! I'll make that change

jobs/build-mechanical.Jenkinsfile
jlebon: I think if there are no `scheduled` streams in RHCOS, we can just disable the job. It's already the case today that if somehow `build-mechanical` wasn't dormant (i.e. we ran it at least once so that it sets up its timer), it would trigger builds for every stream.

marmijo: Yeah, we are just likely to forget about it, though it should be pretty obvious when we have 10 builds all starting at the same time. I'll remove the check here and instead I'll add a comment to the rhcos pipecfg to disable the job if the knob is removed.

utils.groovy
marmijo: Looking at this again, shouldn't this be kept as `.scheduled == true` to collect only the mechanical streams that have `scheduled: true`?

jlebon: Hmm, not sure I follow. It should evaluate to false whether it's unset or set explicitly to `false`.

marmijo: I think you did. You explained why it doesn't need to be kept as I wrote it. 

Thanks! I suppose I was trying to be explicit here, but I see groovy still evaluates to true in this case. I'll update the PR with all the suggestions.

jlebon (approved):
Thanks, LGTM!

PR #1064, opened 2024-11-25
gc-policy: Expand GC to all streams

gursewak1997:
Expand the Garbage Collection to all streams available for FCOS.

### DIFF
--- a/gc-policy.yaml
+++ b/gc-policy.yaml
@@ -1,17 +1,46 @@
+# Production
+testing:
+    cloud-uploads: 24m
+    containers: 3m
+    # images: 30m
+    # images-keep: [qemu, live-iso]
+    # build: 36m # Upgrade test handling needed
+stable:
+    cloud-uploads: 24m
+    containers: 3m
+    # images: 30m
+    # images-keep: [qemu, live-iso]
+    # build: 36m # Upgrade test handling needed
+next:
+    cloud-uploads: 24m
+    containers: 3m
+    # images: 30m
+    # images-keep: [qemu, live-iso]
+    # build: 36m # Upgrade test handling needed
+# Development
 testing-devel:
-    cloud-uploads: 2y
+    cloud-uploads: 12m
     containers: 2w
+    images: 18m
+    images-keep: [qemu, live-iso]
+    build: 24m
 next-devel:
-    cloud-uploads: 2y
+    cloud-uploads: 12m
     containers: 2w
-branched:
-    cloud-uploads: 1y
-    containers: 2w
+    images: 18m
+    images-keep: [qemu, live-iso]
+    build: 24m
+# Mechanical
 rawhide:
-    cloud-uploads: 1y
+    cloud-uploads: 12m
     containers: 2w
+    images: 18m
+    images-keep: [qemu, live-iso]
+    build: 24m
+branched:
+    cloud-uploads: 12m
+    containers: 2w
+    build: 18m
 bodhi-updates:
     cloud-uploads: 1y
-    images: 58m
-    images-keep: [qemu, live-iso]
-    build: 62m
+    build: 2y


gursewak1997:
A quick draft of what we can implement for now. Open to comments for changes on how aggressive we should be for specific streams.
cc @dustymabe

gc-policy.yaml
```diff
@@ -1,15 +1,29 @@
...
     cloud-uploads: 2y
     containers: 2w
 next-devel:
     cloud-uploads: 2y
     containers: 2w
+# Mechanical
 branched:
     cloud-uploads: 1y
     containers: 2w
+    build: 4y
```
dustymabe, line 22: I see branched as much less useful than `rawhide` or even `testing-devel`/`next-devel`.. maybe cloud uploads 9m, images 1y, build 2y.

gc-policy.yaml
```diff
@@ -1,15 +1,29 @@
...
 rawhide:
     cloud-uploads: 1y
     containers: 2w
+    build: 4y
```
dustymabe, line 26: maybe add `images: 3y` ?

gc-policy.yaml
```diff
@@ -1,15 +1,29 @@
...
+next:
+    cloud-uploads: 2y
+    containers: 2m
+# Development
 testing-devel:
     cloud-uploads: 2y
     containers: 2w
 next-devel:
     cloud-uploads: 2y
     containers: 2w
```
dustymabe, line 17: probably should figure out some policy for these `-devel` streams too for images and build

dustymabe: though we can do that later

gc-policy.yaml
```diff
@@ -1,15 +1,29 @@
+# Production
+testing:
+    cloud-uploads: 2y
+    containers: 2m
+stable:
+    cloud-uploads: 2y
+    containers: 2m
+next:
+    cloud-uploads: 2y
+    containers: 2m
```
dustymabe, line 10: @jlebon - thoughts?

gc-policy.yaml
```diff
@@ -1,15 +1,31 @@
...
     cloud-uploads: 1y
     containers: 2w
-rawhide:
+    images: 3y
+    build: 4y
+branched:
     cloud-uploads: 1y
     containers: 2w
+    images: 1y
+    build: 1y
```
dustymabe, line 28: if these are the same value it doesn't mean anything does it?

gc-policy.yaml
```diff
@@ -1,15 +1,31 @@
...
     cloud-uploads: 2y
     containers: 2w
-branched:
+# Mechanical
+rawhide:
     cloud-uploads: 1y
     containers: 2w
-rawhide:
+    images: 3y
+    build: 4y
```
dustymabe, line 23: for rawhide I would probably do images 2y build 3y - so reduce a year off of both of these

gc-policy.yaml
gursewak1997: Nope technically, we could just delete `images` here since the `build` is gonna delete everything anyways.

gc-policy.yaml
jlebon: This SGTM. That said, I wonder if we should send out communications before we start pruning bootimages. We never promised that the bootimages would stick around, but as per [Hyrum's Law](https://www.hyrumslaw.com/), there's guaranteed to be people that rely on them. We shouldn't be afraid to break them, but a heads up would probably be appreciated. WDYT?

Less worried about the container images. We could start pruning those now. 2m WFM for those.

jlebon: Actually, as I was discussing this with @gursewak1997, one thing that I realized is that this probably could break scale up for OKD users due to the classic issue of bootimages never getting updated. Let me reach out to them.

dustymabe (approved):
LGTM

gursewak1997:
Let's hold it till Dec 9th as per our `coreos-status` message.

PR #1069, opened 2024-11-27
jenkins/plugins: update to latest versions

coreosbot-releng:


aaradhak:
created as part of testing

PR #1074, opened 2024-12-13
release: Run GC job only if gc-policy.yaml exists

gursewak1997:
Trigger the GC job only if `gc-policy.yaml` exists, which currently applies only to FCOS.


dustymabe (approved):
/lgtm

gursewak1997:
Not too sure what's happening with the CI check though:/

PR #1076, opened 2024-12-20
jobs/build-mechanical: schedule at off-peak and run builds serially

marmijo:
Adjust the cron schedule trigger to start at 2am UTC every day to run this job at off-peak times. Also, run the scheduled builds serially by waiting for each build to finish before starting the next one.

### DIFF
--- a/jobs/build-mechanical.Jenkinsfile
+++ b/jobs/build-mechanical.Jenkinsfile
@@ -7,8 +7,8 @@ node {
 
 properties([
     pipelineTriggers([
-        // run every 24h only for now
-        cron("H H * * *")
+        // run every 24h at 2am UTC
+        cron("0 2 * * *")
     ]),
     buildDiscarder(logRotator(
         numToKeepStr: '100',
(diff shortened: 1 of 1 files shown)


aaradhak (approved):
LGTM

jobs/build-mechanical.Jenkinsfile
```diff
@@ -25,10 +25,14 @@ node {
...
 
     mechanical_streams.each{
-        echo "Triggering build for mechanical stream: ${it}"
-        build job: 'build', wait: false, parameters: [
-          string(name: 'STREAM', value: it),
-          booleanParam(name: 'EARLY_ARCH_JOBS', value: false)
-        ]
+        try {
+            echo "Triggering build for mechanical stream: ${it}"
+            build job: 'build', wait: true, parameters: [
```
dustymabe, line 30: I think we can add `propagate: false` and we won't need a `try/catch` here.

https://www.jenkins.io/doc/pipeline/steps/pipeline-build-step/

marmijo: Ah, good point! That does seem like it'll be a simpler solution. 

https://github.com/coreos/fedora-coreos-pipeline/pull/1077

PR #1077, opened 2024-12-20
jobs/build-mechanical: use `propagate: false` to wait for build

marmijo:
If a scheduled build fails, the build-mechanical job will fail. Instead of the try/catch block to solve this, use `propagate: false`  instead to build each scheduled stream regardless of the build result.

follow-on to: https://github.com/coreos/fedora-coreos-pipeline/pull/1076


dustymabe (approved):
LGTM

PR #1080, opened 2025-01-22
Update the aws-region ticket details

aaradhak:
The command to view the current value of quota level is added and have also updated the text template.

Also, added the information on creating the DPP ticket in case of RHCOS related requests.

### DIFF
--- a/docs/aws-region-enable.md
+++ b/docs/aws-region-enable.md
@@ -27,11 +27,14 @@ This document provides the information on how to address and resolve missing AWS
     In the dialog box, review the informational text and choose Enable Region.
     Wait until the Region is ready to use.
 
-    We would also like to get AWS public access enabled for this region. Please use the following command:
+    We would like to get AWS public access enabled for this region. Please use the following command:
         "aws ec2 disable-image-block-public-access --region `<missing-region>`"
 
-    Let's also adjust the quota to match the other regions :
-        "aws service-quotas request-service-quota-increase --region=`<missing-region>` --service-code=ec2 --quota-code=L-0E3CBAB9 --desired-value 1000.00"
+    We would also like the "Public AMIs" quota level to be set to a minimum value. The quota codes can be viewed using the below command:
+	"aws service-quotas list-service-quotas --service-code=ec2 --quota-code=L-0E3CBAB9 --region=`<missing-region>`"
+
+    Let's make sure to adjust the quota to match the other regions if necessary. If the current quota is less than 1000, let's bump it to at least 1000 with the following request:
+        "aws service-quotas request-service-quota-increase --service-code=ec2 --quota-code=L-0E3CBAB9 --desired-value 1000.00 --region=`<missing-region>`"
 
     When do you need this to be done by? (YYYY/MM/DD)
     Preferably before our next FCOS release.
@@ -39,7 +42,12 @@ This document provides the information on how to address and resolve missing AWS
 
 ### 3. Submit the Ticket:
 
+In case of FCOS:
 - Ensure all relevant information is included and submit the ticket to the Fedora Infrastructure team.
+
+In case of RHCOS:
+- Follow the link https://devservices.dpp.openshift.com/support/, select "general AWS Question/request ticket" and submit ticket to the DPP team.
+  If more direct interactions would be needed with the folks handling the ticket, please check in the slack channel #forum-pge-cloud-ops
 
 ### 4. Follow Up:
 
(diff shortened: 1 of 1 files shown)


docs/aws-region-enable.md
```diff
@@ -30,21 +30,30 @@ This document provides the information on how to address and resolve missing AWS
     We would also like to get AWS public access enabled for this region. Please use the following command:
         "aws ec2 disable-image-block-public-access --region `<missing-region>`"
 
-    Let's also adjust the quota to match the other regions :
-        "aws service-quotas request-service-quota-increase --region=`<missing-region>` --service-code=ec2 --quota-code=L-0E3CBAB9 --desired-value 1000.00"
+    Let's also adjust the quota to match the other regions if necessary. If the current quota is less than 1000, let's bump it to at least 1000 with the following request:
+        "aws service-quotas request-service-quota-increase --region=ap-southeast-5 --service-code=ec2 --quota-code=L-0E3CBAB9 --desired-value 1000.00"
+
+    The quota codes can be viewed using the below command:
+	"aws service-quotas --service-code ec2 list-service-quotas"
```
dustymabe, line 37: Let's re-arrange things so that we first tell them to run the command to see the current value, and only then instruct them to make the request for increase if it is needed

aaradhak: Thats right, I changed the order .

docs/aws-region-enable.md
```diff
@@ -27,24 +27,33 @@ This document provides the information on how to address and resolve missing AWS
...
     Wait until the Region is ready to use.
 
-    We would also like to get AWS public access enabled for this region. Please use the following command:
+    We would like to get AWS public access enabled for this region. Please use the following command:
         "aws ec2 disable-image-block-public-access --region `<missing-region>`"
 
-    Let's also adjust the quota to match the other regions :
-        "aws service-quotas request-service-quota-increase --region=`<missing-region>` --service-code=ec2 --quota-code=L-0E3CBAB9 --desired-value 1000.00"
+    We would also like the quota level to be set, the quota codes can be viewed using the below command:
+        "aws service-quotas --service-code ec2 list-service-quotas"
```
dustymabe, line 34: The command order is wrong so let's fix that. and let's include specifying the region here like we do in the below command so we ensure when they check the quota number it is > `1000` they are looking at the right region.

```suggestion
        "aws service-quotas list-service-quotas --service-code=ec2 --quota-code=L-0E3CBAB9 --region=ap-southeast-5
```

docs/aws-region-enable.md
```diff
@@ -27,24 +27,33 @@ This document provides the information on how to address and resolve missing AWS
...
+
+    Let's make sure to adjust the quota to match the other regions if necessary. If the current quota is less than 1000, let's bump it to at least 1000 with the following request:
+        "aws service-quotas request-service-quota-increase --region=ap-southeast-5 --service-code=ec2 --quota-code=L-0E3CBAB9 --desired-value 1000.00"
```
dustymabe, line 37: Move region to the end so it's easy to update when copy/pasting.

```suggestion
        "aws service-quotas request-service-quota-increase  --service-code=ec2 --quota-code=L-0E3CBAB9 --desired-value 1000.00 --region=ap-southeast-5"
```

docs/aws-region-enable.md
```diff
@@ -27,24 +27,33 @@ This document provides the information on how to address and resolve missing AWS
     In the dialog box, review the informational text and choose Enable Region.
     Wait until the Region is ready to use.
 
-    We would also like to get AWS public access enabled for this region. Please use the following command:
+    We would like to get AWS public access enabled for this region. Please use the following command:
         "aws ec2 disable-image-block-public-access --region `<missing-region>`"
 
-    Let's also adjust the quota to match the other regions :
-        "aws service-quotas request-service-quota-increase --region=`<missing-region>` --service-code=ec2 --quota-code=L-0E3CBAB9 --desired-value 1000.00"
+    We would also like the quota level to be set, the quota codes can be viewed using the below command:
```
dustymabe, line 33: Mention what quota value we are setting
```suggestion
    We would also like the "Public AMIs" quota level to be set to a minimum value. The quota codes can be viewed using the below command:
```

docs/aws-region-enable.md
aaradhak: Changed the order of the command as suggested

docs/aws-region-enable.md
aaradhak: Updated the command to have the region at the end

docs/aws-region-enable.md
aaradhak: Updated the text as suggested

dustymabe (approved):
LGTM

PR #1081, opened 2025-01-23
jobs/kola-azure: overwrite an upload when using FORCE

nikita-dubrovskii:
This should fix pipeline failure, when blob exists:
```
platform/api/aws: found existing image ami-007a7e984851918dc, reusing
ore/azure: The blob exists. Pass --overwrite to force upload.
ERROR - Command returned bad exitcode
ERROR - COMMAND: ['ore', '--log-level', 'INFO', 'azure', 'upload-blob'
```

### DIFF
--- a/jobs/kola-azure.Jenkinsfile
+++ b/jobs/kola-azure.Jenkinsfile
@@ -31,6 +31,9 @@ properties([
              description: 'The exact config repo git commit to run tests against',
              defaultValue: '',
              trim: true),
+      booleanParam(name: 'FORCE',
+             defaultValue: false,
+             description: 'Whether to force an upload'),
     ]),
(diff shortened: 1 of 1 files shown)


dustymabe:
hmm. but he RHCOS pipeline wouldn't have hit this.. I would have thought the file that would need updating would be https://github.com/coreos/fedora-coreos-pipeline/blob/main/libcloud.groovy

dustymabe:
though I don't see where any of the other builduploads in libcloud.groovy support the force option either so I think maybe we just shouldnt do it here..

in the past when we've hit a problem we probably just restarted the x86_64 job with `FORCE=1` so we'd get a new build ID and work around the problem that way.

PR #1087, opened 2025-02-04
Update the TenantEgress on adding aws regions in RHCOS

aaradhak:
When a new aws-region is added in RHCOS, the tenantegress is required to be updated. The process to update the tenantegress is added for reference.

### DIFF
--- a/docs/aws-region-enable.md
+++ b/docs/aws-region-enable.md
@@ -52,6 +52,17 @@ In case of RHCOS:
 - Account: 531415883065
   ARN: arn:aws:iam::531415883065:user/art-rhcos-ci-s3-bot
   Refer 'RH-DEV AWS Account' in Bitwarden for User ID
+- Note:
+    - Ensure that the new AWS regions are added to the Tenant Egress configuration in the RHCOS ITUP PROD cluster to prevent any TCP timeout issues.
+    - Update the Tenant Egress directory in the internal Git repository with details of the new AWS regions.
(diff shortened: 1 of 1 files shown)


docs/aws-region-enable.md
```diff
@@ -52,6 +52,16 @@ In case of RHCOS:
 - Account: 531415883065
   ARN: arn:aws:iam::531415883065:user/art-rhcos-ci-s3-bot
   Refer 'RH-DEV AWS Account' in Bitwarden for User ID
+- Note:
+   - Make sure to add the new regions link in the Tenant Egress of the PROD Cluster to avoid any TCP timeout.
+   - Also, update the Tenant Egress directory with the details of the new aws regions and the DPP ticket.
```
mike-nguyen, line 57: What is the difference between this step and the bullet point above?

aaradhak: The first point is to update the tenantegress file in the cluster. The second is to update the Tenant egress directory in the internal repo. If it is not clear, I can rephrase it better.

marmijo: We should enforce the practice of only updating the TenantEgress file in the internal repo. We can then deploy the template to enact the change on the cluster after it's been reviewed in the repo first. 
So maybe this could be worded as something like "Follow the internal process to add the new region endpoint URL to the TenantEgress of the Prod Cluster."

aaradhak: I have rephrased it. Hope its worded better.

dustymabe (approved):

PR #1088, opened 2025-02-04
config: add `streams[].skip_disk_images`

jlebon:
As part of https://github.com/openshift/enhancements/pull/1637, we've moved OCP 4.19 to use bootimages with RHEL content only. This means that the bootimages built with OCP content will never be used in practice.

Add a `skip_disk_images` knob to disable building them.

We still generate the QEMU image for kola tests to run and because they're useful to debug, but we drop everything else. (Actually, we could also not generate the QEMU image either and sanity-check the OCI image with `kola run --oscontainer`, but that requires more rewiring.)

We don't generate live media since I don't think the test coverage from that is meaningfully different enough from the RHEL-only variants given that the additional OCP packages are unrelated.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -59,6 +59,10 @@ streams:
       variant: rhcos-9.0
       # OPTIONAL: override whether to use a versionary for this stream
       versionary: false
+      # OPTIONAL: Whether to skip disk image generation and upload. This
+      # is equivalent to listing all live artifacts and cloud platforms in
+      # `skip_artifacts`.
+      skip_disk_images: true
     stable:
       type: production
       # OPTIONAL: override cosa image to use for this stream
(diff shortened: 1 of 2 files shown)


jbtrystram (approved):

dustymabe:
I'll approve this but also question..

How long do we think we'll use this? i.e. soon we'll be building the Node Image on top of the base image and theoretically we won't have a stream entry at all (and thus not need this configuration) for the node level image?

dustymabe (approved):

jlebon:
> How long do we think we'll use this? i.e. soon we'll be building the Node Image on top of the base image and theoretically we won't have a stream entry at all (and thus not need this configuration) for the node level image?

Yes, ideally soon we won't have a need for this. Though even assuming everything goes smoothly, it'll still be a few weeks before it all lands. So this is just trying to not be wasteful meanwhile and generate/compress/upload things we don't care about.

PR #1089, opened 2025-02-06
config: enable branched stream

aaradhak:
As F42 has branched from rawhide, we are building the branched stream again

Ref: https://github.com/coreos/fedora-coreos-tracker/issues/1851


travier (approved):

config.yaml
```diff
@@ -17,10 +17,10 @@ streams:
...
       COSA_TESTISO_DEBUG: true
     osbuild_experimental: true
-  # branched:
-  #  type: mechanical
-  #  env:
-  #    COSA_USE_OSBUILD: true
+  branched:
+    type: mechanical
+    env:
+      COSA_USE_OSBUILD: true
```
dustymabe, line 23: This is old.. let's replace it with:

```suggestion
    osbuild_experimental: true
```

aaradhak: done, I have made the changes

dustymabe (approved):
/lgtm

PR #1091, opened 2025-02-12
release: allow pushing container images to multipes repositories

jbtrystram:
In RHCOS we need to publish the oscontainer image to the release repo and also the CI repo. Changing the registry_repo object to be an array instead of a map to allow multiple entries.

Patch best reviwed without whitespace changes

### DIFF
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -259,26 +259,28 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
         if (push_containers) {
             stage("Push Containers") {
                 parallel push_containers.collectEntries{configname, val -> [configname, {
-                    if (!registry_repos?."${configname}"?.'repo') {
+                    if (!registry_repos?."${configname}") {
                         echo "No registry repo config for ${configname}. Skipping"
                         return
                     }
                     withCredentials([file(variable: 'REGISTRY_SECRET',
                                           credentialsId: 'oscontainer-push-registry-secret')]) {
-                        def repo = registry_repos[configname]['repo']
+                        def repos = registry_repos[configname]
                         def (artifact, metajsonname) = val
-                        def tag_args = registry_repos[configname].tags.collect{"--tag=$it"}
-                        def v2s2_arg = registry_repos.v2s2 ? "--v2s2" : ""
-                        shwrap("""
-                        export COSA_SUPERMIN_MEMORY=1024 # this really shouldn't require much RAM
-                        cp \${REGISTRY_SECRET} tmp/push-secret-${metajsonname}
-                        cosa supermin-run /usr/lib/coreos-assembler/cmd-push-container-manifest \
-                            --auth=tmp/push-secret-${metajsonname} \
-                            --repo=${repo} ${tag_args.join(' ')} \
-                            --artifact=${artifact} --metajsonname=${metajsonname} \
-                            --build=${params.VERSION} ${v2s2_arg}
-                        rm tmp/push-secret-${metajsonname}
-                        """)
+                        for (repo in repos) {
+                            def tag_args = repo.tags.collect{"--tag=$it"}
+                            def v2s2_arg = repo.v2s2 ? "--v2s2" : ""
+                            shwrap("""
+                            export COSA_SUPERMIN_MEMORY=1024 # this really shouldn't require much RAM
+                            cp \${REGISTRY_SECRET} tmp/push-secret-${metajsonname}
+                            cosa supermin-run /usr/lib/coreos-assembler/cmd-push-container-manifest \
+                                --auth=tmp/push-secret-${metajsonname} \
+                                --repo=${repo.repo} ${tag_args.join(' ')} \
+                                --artifact=${artifact} --metajsonname=${metajsonname} \
+                                --build=${params.VERSION} ${v2s2_arg}
+                            rm tmp/push-secret-${metajsonname}
+                            """)
+                        }
                     }
                 }]}
             }
--- a/utils.groovy
+++ b/utils.groovy
@@ -516,26 +516,29 @@ def build_artifacts(pipecfg, stream, basearch) {
 
 def get_registry_repos(pipecfg, stream, version) {
     def registry_repos = pipecfg.registry_repos ?: [:]
-    // merge top-level registry_repos with stream-specific bits
-    registry_repos += pipecfg.streams[stream].additional_registry_repos ?: [:]
-    for (repo in (registry_repos.keySet() as List)) {
-        if (repo == 'v2s2') {
-            // this is a boolean option, not a registry repo
-            continue
-        }
-        if (registry_repos[repo].tags) {
-            def processed_tags = []
-            for (tag in registry_repos."${repo}".tags) {
-                tag = utils.substituteStr(tag, [STREAM: stream, VERSION: version])
-                if (pipecfg.hotfix) {
-                    // this is a hotfix build; include the hotfix name
-                    // in the tag suffix so we don't clobber official
-                    // tags
-                    tag += "-hotfix-${pipecfg.hotfix.name}"
+    // get stream-specific bits additional repos
+    def additional_repos = pipecfg.streams[stream].additional_registry_repos ?: [:]
+    for (repo_id in (additional_repos.keySet() as List)) {
+        // merge top-level registry_repos with additional repos
+        def merged_registry_repos = registry_repos[repo_id] ?: []
+        registry_repos[repo_id] = merged_registry_repos + (additional_repos[repo_id] ?: [])
+     }
+    for (repo_id in (registry_repos.keySet() as List)) {
+        for (repo in registry_repos[repo_id]) {
+            if (repo.tags) {
+                def processed_tags = []
+                for (tag in repo.tags) {
+                    tag = utils.substituteStr(tag, [STREAM: stream, VERSION: version])
+                    if (pipecfg.hotfix) {
+                        // this is a hotfix build; include the hotfix name
+                        // in the tag suffix so we don't clobber official
+                        // tags
+                        tag += "-hotfix-${pipecfg.hotfix.name}"
+                    }
+                    processed_tags += tag
                 }
-                processed_tags += tag
+                repo.tags = processed_tags
             }
-            registry_repos[repo]['tags'] = processed_tags
         }
     }
     return registry_repos
--- a/config.yaml
+++ b/config.yaml
@@ -40,11 +40,11 @@ s3:
 
 registry_repos:
   oscontainer:
-    repo: quay.io/fedora/fedora-coreos
-    tags: ["${STREAM}", "${VERSION}"]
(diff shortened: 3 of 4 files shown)


utils.groovy
```diff
@@ -516,26 +516,32 @@ def build_artifacts(pipecfg, stream, basearch) {
 
 def get_registry_repos(pipecfg, stream, version) {
     def registry_repos = pipecfg.registry_repos ?: [:]
-    // merge top-level registry_repos with stream-specific bits
-    registry_repos += pipecfg.streams[stream].additional_registry_repos ?: [:]
+    // get stream-specific bits additional repos
+    def additional_repos = pipecfg.streams[stream].additional_registry_repos ?: [:]
```
dustymabe, line 520: additional_registry_repos isn't added in the docs section of this PR

utils.groovy
```diff
@@ -516,26 +516,32 @@ def build_artifacts(pipecfg, stream, basearch) {
...
+    for (repo in ['oscontainer', 'legacy_oscontainer', 'extensions', 'kubevirt']) {
```
dustymabe, line 521: I think I would prefer it if we didn't hardcode these values here. Can we not just iterate through the keys of `additional_repos`?

utils.groovy
```diff
@@ -516,26 +516,32 @@ def build_artifacts(pipecfg, stream, basearch) {
...
-        if (registry_repos[repo].tags) {
-            def processed_tags = []
-            for (tag in registry_repos."${repo}".tags) {
-                tag = utils.substituteStr(tag, [STREAM: stream, VERSION: version])
-                if (pipecfg.hotfix) {
-                    // this is a hotfix build; include the hotfix name
-                    // in the tag suffix so we don't clobber official
-                    // tags
-                    tag += "-hotfix-${pipecfg.hotfix.name}"
+        for (entry in registry_repo[repo]) {
```
dustymabe, line 530: bug - `registry_repo` -> `registry_repos`

but might I suggest here that renaming the variables could help. 

I would change `repo` above to `repo_id` and then

here I would change to 
```suggestion
        for (repo in registry_repos[repo_id]) {
```

utils.groovy
```diff
@@ -516,26 +516,32 @@ def build_artifacts(pipecfg, stream, basearch) {
...
     def registry_repos = pipecfg.registry_repos ?: [:]
-    // merge top-level registry_repos with stream-specific bits
-    registry_repos += pipecfg.streams[stream].additional_registry_repos ?: [:]
+    // get stream-specific bits additional repos
+    def additional_repos = pipecfg.streams[stream].additional_registry_repos ?: [:]
+    for (repo in ['oscontainer', 'legacy_oscontainer', 'extensions', 'kubevirt']) {
+        //merge top-level registry_repos with addtional repos
+        registry_repos[repo] += additional_repos[repo] ?: []
+     }
     for (repo in (registry_repos.keySet() as List)) {
```
dustymabe, line 525: suggest to change this to repo_id (i.e. `oscontainer`, `extensions`, etc). It's just the key in the hash. 
```suggestion
    for (repo_id in (registry_repos.keySet() as List)) {
```

utils.groovy
dustymabe: also suggest to use `repo_id` as the loop var here.

utils.groovy
jbtrystram: Yeah it's not something I introduced there, but it's indeed not documented.
I'll add it

utils.groovy
jbtrystram: > Can we not just iterate through the keys of additional_repos?

I don't think so, because additional_repos and _registry_repos may contain different keys

utils.groovy
```diff
@@ -516,26 +516,32 @@ def build_artifacts(pipecfg, stream, basearch) {
 
 def get_registry_repos(pipecfg, stream, version) {
     def registry_repos = pipecfg.registry_repos ?: [:]
-    // merge top-level registry_repos with stream-specific bits
-    registry_repos += pipecfg.streams[stream].additional_registry_repos ?: [:]
+    // get stream-specific bits additional repos
+    def additional_repos = pipecfg.streams[stream].additional_registry_repos ?: [:]
+    for (repo in ['oscontainer', 'legacy_oscontainer', 'extensions', 'kubevirt']) {
+        //merge top-level registry_repos with addtional repos
+        registry_repos[repo] += additional_repos[repo] ?: []
```
jlebon, line 523: We need to handle the case here where `repo` is not defined on `registry_repos` so it's not a list yet we can append to.

jlebon (commented):
Some comments, but approach LGTM overall!

This changes the schema, so we'll need to make sure to have the changes for the RHCOS pipecfg as well lined up so they merge at the same time.

utils.groovy
jbtrystram: > iterate through the keys of additional_repos

Yeah that make way more sense actually, thanks for the suggestion

jobs/release.Jenkinsfile
```diff
@@ -259,26 +259,28 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
         if (push_containers) {
             stage("Push Containers") {
                 parallel push_containers.collectEntries{configname, val -> [configname, {
-                    if (!registry_repos?."${configname}"?.'repo') {
+                    if (!registry_repos?."${configname}"?.isEmpty()) {
```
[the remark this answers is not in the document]
jbtrystram: if I understand correctly, if `$configname` is missing the `?` will short circuit the expression

dustymabe (approved):
LGTM

utils.groovy
```diff
@@ -516,26 +516,29 @@ def build_artifacts(pipecfg, stream, basearch) {
...
-                if (pipecfg.hotfix) {
-                    // this is a hotfix build; include the hotfix name
-                    // in the tag suffix so we don't clobber official
-                    // tags
-                    tag += "-hotfix-${pipecfg.hotfix.name}"
+    // get stream-specific bits additional repos
+    def additional_repos = pipecfg.streams[stream].additional_registry_repos ?: [:]
+    for (repo_id in (additional_repos.keySet() as List)) {
+        //merge top-level registry_repos with addtional repos
+        def merged_registry_repos = registry_repos[repo_id] :? []
```
jlebon, line 523:
```suggestion
        def merged_registry_repos = registry_repos[repo_id] ?: []
```

utils.groovy
```diff
@@ -516,26 +516,29 @@ def build_artifacts(pipecfg, stream, basearch) {
...
-                tag = utils.substituteStr(tag, [STREAM: stream, VERSION: version])
-                if (pipecfg.hotfix) {
-                    // this is a hotfix build; include the hotfix name
-                    // in the tag suffix so we don't clobber official
-                    // tags
-                    tag += "-hotfix-${pipecfg.hotfix.name}"
+    // get stream-specific bits additional repos
+    def additional_repos = pipecfg.streams[stream].additional_registry_repos ?: [:]
+    for (repo_id in (additional_repos.keySet() as List)) {
+        //merge top-level registry_repos with addtional repos
```
jlebon, line 522:
```suggestion
        // merge top-level registry_repos with additional repos
```

utils.groovy
```diff
@@ -516,26 +516,29 @@ def build_artifacts(pipecfg, stream, basearch) {
...
+        def merged_registry_repos = registry_repos[repo_id] :? []
+        registry_repos[repo_id] = merged_registry_repos + (additional_repos[repo_id] ?: [])
+     }
+    for (repo_id in (registry_repos.keySet() as List)) {
+        for (repo in registry_repos[repo_id]) {
+            if (entry.tags) {
```
jlebon, line 528: repo

jlebon: And throughout the iteration block.

jlebon (commented):
This is now blocked on https://issues.redhat.com/browse/DPTP-4328.

jbtrystram:
> This is now blocked on https://issues.redhat.com/browse/DPTP-4328.

Is it ? The RHCOS change is, yes, but we can test this change before adding the extra repository in the RHCOS config

jlebon:
> > This is now blocked on [issues.redhat.com/browse/DPTP-4328](https://issues.redhat.com/browse/DPTP-4328).
> 
> Is it ? The RHCOS change is, yes, but we can test this change before adding the extra repository in the RHCOS config

Yes, that's fine assuming we remove that part of the pipecfg PR.

utils.groovy
jbtrystram: thanks for catching this :sweat_smile:

jobs/release.Jenkinsfile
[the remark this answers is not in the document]
jbtrystram: Did some testing and haven't had issues. However I see what you mean : it would not exit and fail afterwards I suppose.
WDYT of 
` if (!registry_repos?."${configname}"?) {`
since empty collections evaluate to false

jobs/release.Jenkinsfile
```diff
@@ -259,26 +259,28 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
...
-                            --build=${params.VERSION} ${v2s2_arg}
-                        rm tmp/push-secret-${metajsonname}
-                        """)
+                        for (repo in repos) {
+                            def tag_args = repo.tags.collect{"--tag=$it"}
+                            def v2s2_arg = repo.v2s2 ? "--v2s2" : ""
+                            shwrap("""
+                            export COSA_SUPERMIN_MEMORY=1024 # this really shouldn't require much RAM
+                            cp \${REGISTRY_SECRET} tmp/push-secret-${metajsonname}
+                            cosa supermin-run /usr/lib/coreos-assembler/cmd-push-container-manifest \
```
jlebon, line 276: Reviewing this with JB, one thing we remembered is that registry.ci doesn't actually support manifests. So we need to fall back to `cosa copy-container` instead here which has the magic "push manifested images as separate images with an arch suffix tag". (Or, add that logic to `cmd-push-container-manifest` which feels a bit weird given its name...)

jbtrystram: Made the change [c206a72ed604e10d364f981e29bc99f927615fc3](https://github.com/coreos/fedora-coreos-pipeline/pull/1091/commits/c206a72ed604e10d364f981e29bc99f927615fc3)
This will publish an extra tag on the destination registry as right now `copy-container` does not support differentiating source tag and destination tags. I think that's OK though as is the majority of case that's going to be `$VERSION` and will end up being the same.

https://github.com/coreos/coreos-assembler/blob/1f84d64cd85aa2401bdcf398d4210ea4954a0790/src/cmd-copy-container#L117

jbtrystram: cool !

jbtrystram: Hopefully I had made a separate commit

jlebon (approved):
LGTM! Will merge when ready to merge the pipecfg one.

jlebon:
FYI, follow-up to this in https://github.com/coreos/fedora-coreos-pipeline/issues/1099.

PR #1092, opened 2025-02-12
Update config to run Intel TDX test

HuijingHei:
Update config to run SNP test and add Intel TDX test. See https://github.com/coreos/fedora-coreos-config/pull/3356


jbtrystram (approved):
lgtm

PR #1095, opened 2025-02-17
bump-lockfile: consolidate image building; use OSBuild

dustymabe:
We haven't yet switched over the production streams to use OSBuild for the Live ISO/PXE yet, but I think it's time for us to go ahead and switch this job over (since `next` will be on Fedora 42 beta soon, which will be using OSBuild for live*).

While we are here let's just go ahead and build all disks up front, which means less OSBuild invocations to achieve the same result with less effort because we aren't jumping in and out of supermin multiple times.


marmijo (approved):
LGTM

PR #1096, opened 2025-02-20
utils: collect artifacts to build as a Set

marmijo:
Collect all artifacts in a Set to support the `skip_disk_images` knob. This avoids Jenkins script security errors when using `intersect()`.


jlebon (approved):
LGTM, thanks!

PR #1097, opened 2025-02-24
utils: convert collected as a Set artifact into List

nikita-dubrovskii:
This fixes introduced by previous commit issue:
```
hudson.remoting.ProxyException: groovy.lang.MissingMethodException: No signature of method: java.util.LinkedHashSet.add() is applicable for argument types: (java.lang.Integer, java.lang.String) values: [0, metal]
```


jlebon (approved):
Ahh yes, it needs to be a list in the end because we need to be able to order them for building. That's a detail that won't matter once we fully cut over to a single `cosa osbuild` invocation.

Issue #1079, opened 2025-01-07
Add a startup delay to kola-aws and kola-gcp jobs


jbtrystram:
We had an issue recently where kola-aws and kola-gcp where killed in the fetch metadata step because they started at the same time. 

Add a parametrized delay and have the `build` job pass a random delay between 0 and 120s

jlebon:
I would probably just use a `lock` for this instead of a random delay.

But... sigh ideally we do try to get to the bottom of the underlying issue.

dustymabe:
> I would probably just use a `lock` for this instead of a random delay.

good idea. that works too

PR #1098, opened 2025-02-25
jobs/build: lock buildfetch to avoid simultaneous jobs race condition

c4rt0:
jobs/kola: lock kola cloud jobs to avoid race condition
 
There were an issue recently where kola-aws and kola-gcp were
killed in the fetch metadata step because they started
at the same time. This aims to block cloud jobs since
the buildfetch doesn't take long. It shouldn't be too disruptive
to do this across all jobs that buildfetch at the beginning.
 
See: https://github.com/coreos/fedora-coreos-pipeline/issues/1079

### DIFF
--- a/jobs/kola-openstack.Jenkinsfile
+++ b/jobs/kola-openstack.Jenkinsfile
@@ -61,32 +61,34 @@ lock(resource: "kola-openstack-${params.ARCH}") {
     try {
 
         def openstack_image_name, openstack_image_filepath
-        stage('Fetch Metadata/Image') {
-            def commitopt = ''
-            if (params.SRC_CONFIG_COMMIT != '') {
-                commitopt = "--commit=${params.SRC_CONFIG_COMMIT}"
-            }
-            // Grab the metadata. Also grab the image so we can upload it.
-            withCredentials([file(variable: 'AWS_CONFIG_FILE',
-                                  credentialsId: 'aws-build-upload-config')]) {
-                def ref = pipeutils.get_source_config_ref_for_stream(pipecfg, params.STREAM)
-                def variant = stream_info.variant ? "--variant ${stream_info.variant}" : ""
-                shwrap("""
-                cosa init --branch ${ref} ${commitopt} ${variant} ${pipecfg.source_config.url}
-                time -v cosa buildfetch --build=${params.VERSION} --arch=${params.ARCH} \
-                    --url=s3://${s3_stream_dir}/builds --artifact=openstack
-                """)
-                pipeutils.withXzMemLimit(cosa_memory_request_mb - 512) {
-                    shwrap("cosa decompress --build=${params.VERSION} --artifact=openstack")
+        lock(resource: "kola-cloud-buildfetch") {
+            stage('Fetch Metadata/Image') {
+                def commitopt = ''
+                if (params.SRC_CONFIG_COMMIT != '') {
+                    commitopt = "--commit=${params.SRC_CONFIG_COMMIT}"
                 }
-                openstack_image_filepath = shwrapCapture("""
-                cosa meta --build=${params.VERSION} --arch=${params.ARCH} --image-path openstack
-                """)
+                // Grab the metadata. Also grab the image so we can upload it.
+                withCredentials([file(variable: 'AWS_CONFIG_FILE',
+                                    credentialsId: 'aws-build-upload-config')]) {
+                    def ref = pipeutils.get_source_config_ref_for_stream(pipecfg, params.STREAM)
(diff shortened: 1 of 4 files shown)


jobs/build.Jenkinsfile
```diff
@@ -225,26 +225,27 @@ lock(resource: "build-${params.STREAM}") {
...
-        stage('BuildFetch') {
-            if (s3_stream_dir) {
-                pipeutils.shwrapWithAWSBuildUploadCredentials("""
-                cosa buildfetch --arch=${basearch} \
-                    --url s3://${s3_stream_dir}/builds \
-                    --aws-config-file \${AWS_BUILD_UPLOAD_CONFIG}
-                """)
-                if (parent_version != "") {
-                    // also fetch the parent version; this is used by cosa to do the diff
+        lock(resource: "build-${params.STREAM}") {
```
jbtrystram, line 228: This will lock any other builds in the `build` stage, we should have a lock dedicated for this fetch operation: 
```suggestion
        lock(resource: "buildfetch") {
```

Since multiple streams jobs can start at the same time let's also have a global lock. It's less than <20s so it's not going to stale the whole pipeline for long.

jbtrystram (changes requested):
Can you add more explanation in the commit message as well ?

jobs/build.Jenkinsfile
c4rt0: Ok, that makes sense. Thanks!

dustymabe:
the buildfetch that needs adjusting are the ones that download artifacts. Openstack/Azure download the actual image and the aws/gcp ones download the ostree. So basically we need the lock on the *cloud* kola jobs. We think this problem is exacerbated by all of these jobs getting kicked off at the exact same time.

So let's add a global lock to just those jobs. maybe call it `cloud-buildfetch` or something.

dustymabe:
A few things:

1. The lock needs to go just around the "fetch" stage. not around the whole job
2. The lock needs to be named something common (otherwise we're not actually making the run exclusive). My suggestion above was something like `kola-cloud-buildfetch`. 
3. let's add the lock to ALL kola "cloud" jobs that get kicked off, not just AWS/GCP.

c4rt0:
>     The lock needs to be named something common (otherwise we're not actually making the run exclusive). My suggestion above was something like `kola-cloud-buildfetch`.
Ok now this makes sense. I've been wondering what is the relation between resource and the lock.

Thanks.

jobs/kola-openstack.Jenkinsfile
```diff
@@ -92,20 +92,22 @@ lock(resource: "kola-openstack-${params.ARCH}") {
...
-                # First delete it if it currently exists, then create it.
-                ore openstack --config-file=\${OPENSTACK_KOLA_TESTS_CONFIG} \
-                    --region=${region} \
-                    delete-image --id=${openstack_image_name} || true
-                ore openstack --config-file=\${OPENSTACK_KOLA_TESTS_CONFIG} \
-                    --region=${region} \
-                    create-image --file=${openstack_image_filepath} \
-                    --name=${openstack_image_name} --arch=${params.ARCH}
-                """)
+            lock(resource: "kola-cloud-buildfetch") {
```
dustymabe, line 95: This lock needs to go higher up around the `stage('Fetch Metadata/Image')` stage and not here.

dustymabe (changes requested):
Very close. Needs one change!

jobs/kola-openstack.Jenkinsfile
c4rt0: Ouch, that sheds some light on my attention to detail this evening. My bad, thanks for catching this.

c4rt0: Resolved in: https://github.com/coreos/fedora-coreos-pipeline/commit/f2cf30f234696e6a0e84ce5e5a7ff4d6d4b8f2bd

jobs/kola-openstack.Jenkinsfile
```diff
@@ -105,7 +107,8 @@ lock(resource: "kola-openstack-${params.ARCH}") {
                     --name=${openstack_image_name} --arch=${params.ARCH}
                 """)
             }
-        
+
+
```
dustymabe, line 111: only thing I would delete is remove this blank line addition

dustymabe (approved):
LGTM - will merge soon

jobs/kola-openstack.Jenkinsfile
c4rt0: Done

dustymabe (approved):

PR #1100, opened 2025-03-06
jobs: Add build-node-image.Jenkinsfile 

ravanelli:
- As part of openshift/os#799, we'll want to build the "OCP node" image as a layered image on top of the RHCOS base image.
 - Create a separete job where we build it and push to the registry

### DIFF
--- a/jobs/build-node-image.Jenkinsfile
+++ b/jobs/build-node-image.Jenkinsfile
@@ -1,0 +1,147 @@
+def gitref, commit, shortcommit
+
+node {
+    checkout scm
+    // these are script global vars
+    pipeutils = load("utils.groovy")
+    pipecfg = pipeutils.load_pipecfg()
+}
+
+properties([
+    pipelineTriggers([cron('H H * * *')]),
+    parameters([
+      choice(name: 'RELEASE',
+             choices: pipeutils.get_streams_choices(pipecfg, true),
+             description: 'CoreOS stream to build'),
+      string(name: 'FROM',
+             description: 'Image FROM',
+             defaultValue: '',
+             trim: true),
+      string(name: 'ARCHES',
+             description: 'Space-separated list of target architectures',
+             defaultValue: "x86_64 aarch64 ppc64le s390x",
+             trim: true),
+      string(name: 'COREOS_ASSEMBLER_IMAGE',
+             description: 'Override the coreos-assembler image to use',
+             defaultValue: "quay.io/coreos-assembler/coreos-assembler:latest",
+             trim: true),
+    ]),
+    buildDiscarder(logRotator(
+        numToKeepStr: '100',
+        artifactNumToKeepStr: '100'
+    )),
+    durabilityHint('PERFORMANCE_OPTIMIZED')
+])
+
+def stream_info = pipecfg.ocp_node_builds.release[params.RELEASE]
+def src_config_ref = stream_info.source_config.ref
+def src_config_url = stream_info.source_config.url
+
+lock(resource: "build-node-image") {
+    cosaPod(image: params.COREOS_ASSEMBLER_IMAGE,
+            memory: "512Mi", kvm: false,
+            serviceAccount: "jenkins") {
+    timeout(time: 15, unit: 'MINUTES') {
+    try {
+
+        def output = shwrapCapture("git ls-remote ${src_config_url} ${src_config_ref}")
+        commit = output.substring(0,40)
+        shortcommit = commit.substring(0,7)
+        currentBuild.description = "[${params.RELEASE}@${shortcommit}] Running"
+
+        // Get the list of requested architectures to build for
+        def arches = params.ARCHES.split() as Set
+        def archinfo = arches.collectEntries{[it, [:]]}
+        def container_registry_staging_repo = pipecfg.ocp_node_builds.registries.staging
+        def container_registry_repo_and_tag = utils.substituteStr(pipecfg.ocp_node_builds.registries.prod,
+                                                                 [RELEASE: params.RELEASE])
+
+        // add any additional root CA cert before we do anything that fetches
+        pipeutils.addOptionalRootCA()
+
+        stage('Init') {
+            shwrap("""git clone ${stream_info.yumrepo.url} yumrepos""")
+        }
+
+        stage('Build Layered Image') {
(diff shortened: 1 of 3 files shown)


dustymabe:
I'm somewhat against adding a new job that does something really similar to a bunch of other jobs here.

There's

- build-cosa
- build-fcos-buildroot
- build-podman-os (which we can delete now)
- build-kola-containers

I feel like we need to start consolidating the "build container for all arches" flow into a library and use that.

jobs/build-node-image.Jenkinsfile
```diff
@@ -0,0 +1,215 @@
...
+// kola parallelism from that. We leave 512M for overhead and VMs are at most
+// 1.5G each (in general; root reprovisioning tests require 4G which is partly
+// why we run them separately).
+// XXX: https://github.com/coreos/coreos-assembler/issues/3118 will make this
+// cleaner
+def ncpus = ((cosa_memory_request_mb - 512) / 1536) as Integer
+
+// runtime parameter always wins
+def cosa_img = params.COREOS_ASSEMBLER_IMAGE
+cosa_img = cosa_img ?: pipeutils.get_cosa_img(pipecfg, params.RELEASE)
```
jlebon, line 57: Let's drop this. We can always add release-level cosa overrides if needed down the line, but hopefully we've moved over to Konflux before that.

jobs/build-node-image.Jenkinsfile
```diff
@@ -0,0 +1,215 @@
...
+            def variant = stream_info.variant ? "--variant ${stream_info.variant}" : ""
+            shwrap("""
+            cosa init --force --branch ${src_config_ref} --commit=${commit} ${yumrepos} ${variant} ${src_config_url}
+            """)
+        }
+
+        stage('Build Layered Image') {
+            withCredentials([file(credentialsId: 'oscontainer-push-registry-secret', variable: 'REGISTRY_SECRET')]) {
+                def src_config_from = params.FROM ?: stream_info.from
+                parallel arches.collectEntries{arch -> [arch, {
```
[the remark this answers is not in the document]
ravanelli: A think you are right, we only did it for fcos! We could launch an instance in AWS for it, but since we are trying to reduce costs, we probably will need to run it in the job itself.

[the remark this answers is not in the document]
jlebon: Yeah, that's going to be messy to have it be built one way for x86_64 and another for the other arches. Esp. once it becomes a library call.

I guess see how it comes out. Another option actually would be to just launch an FCOS QEMU VM right here and set it up for SSH and that'll be our "builder" for this run.

jobs/build-node-image.Jenkinsfile
```diff
@@ -0,0 +1,215 @@
...
+                shwrap("""cp ${REGISTRY_SECRET} push-secret""")
+
+                // We need to run all buildah commands in the same superminvm to not lose
+                // the manifest creation/additions
+                shwrap("""cosa supermin-run bash -c \
+                    "${cmds} && \
+                    /usr/bin/buildah manifest push --authfile ./push-secret --all ${manifest} \
+                    docker://${manifest}"
+                """)
+                shwrap("""rm -f  push-secret""")
```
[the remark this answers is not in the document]
ravanelli: in my mind we had an issue with it, looking back it was related to the osbuild, changed.

jobs/build-node-image.Jenkinsfile
```diff
@@ -0,0 +1,183 @@
...
+                          noTags: true,
+                          reference: '',
+                          shallow: true]]
+        ]
+    )
+
+    def output = shwrapCapture("git rev-parse HEAD")
+    commit = output.substring(0,40)
+    shortcommit = commit.substring(0,7)
+}
```
jlebon, line 78: Is it even worth cloning the repo here? I think instead we could just `git ls-remote` from within the cosa pod.

jobs/build-node-image.Jenkinsfile
```diff
@@ -0,0 +1,183 @@
...
+                }
+            }
+        }
+        withCredentials([file(credentialsId: 'oscontainer-push-registry-secret', variable: 'REGISTRY_SECRET')]) {
+           stage("Release Manifest") {
+                // Release the manifest and container images
+                shwrap("""
+                    skopeo copy --all --authfile \$REGISTRY_SECRET   \
+                    docker://${container_registry_staging_repo}:node-image-${params.RELEASE}-${shortcommit} \
+                    docker://${container_registry_repo}:${params.RELEASE}-node
```
jlebon, line 147: Let's indent those so it's clearer that it's a continuation.

jlebon: And similar for the other `skopeo` invocations below.

jobs/build-node-image.Jenkinsfile
```diff
@@ -0,0 +1,183 @@
...
+                def images = ""
+                for (arch in arches"]) {
+                    images += " --image=docker://${container_registry_staging_repo}:${params.RELEASE}-${arch}-${shortcommit}"
+                }
+                // arbitrarily selecting the x86_64 builder; we don't run this
+                // locally because podman wants user namespacing (yes, even just
+                // to push a manifest...)
+                pipeutils.withPodmanRemoteArchBuilder(arch: "ppc64le") {
+                    shwrap("""
+                    cosa push-container-manifest --v2s2 \
```
jlebon, line 134: Let's drop `--v2s2`.

jobs/build-node-image.Jenkinsfile
```diff
@@ -0,0 +1,183 @@
...
+        withCredentials([file(credentialsId: 'oscontainer-push-registry-secret', variable: 'REGISTRY_SECRET')]) {
+            stage("Push Manifest") {
+                def images = ""
+                for (arch in arches"]) {
+                    images += " --image=docker://${container_registry_staging_repo}:${params.RELEASE}-${arch}-${shortcommit}"
+                }
+                // arbitrarily selecting the x86_64 builder; we don't run this
+                // locally because podman wants user namespacing (yes, even just
+                // to push a manifest...)
+                pipeutils.withPodmanRemoteArchBuilder(arch: "ppc64le") {
```
jlebon, line 132: Comment says x86_64 but here we use ppc64le. That said, if we had to pick a random builder, I'd probably pick `s390x` since usually the fastest.

PR #1101, opened 2025-03-12
 Remove outdated ITPAAS references 

ravanelli:


ravanelli:
The `builder-splunk.bu` is also broken, I had to remove the `builder-splunk.bu` merge to be able to launch a new instance in AWS. It is failing in the ignition part, when it has the splunk sutff added.  @marmijo let me know if you have cycles to look at it, if not, I can try to look after.

marmijo (approved):
LGTM

PR #1107, opened 2025-03-18
Add container extension build and add new functions to utils

ravanelli:


utils.groovy
```diff
@@ -820,4 +820,92 @@ def matrixSend(message) {
     }
 }
 
+def build_remote_image(arches, commit, url, repo, tag, yumrepo=None, from=None,
+                        extraArgs=None) {
+     // Note we can pass ARCH_REPLACE keyword in the tag,
+     // in this way we make the process easier to handle arch tag names.
+    withCredentials([file(credentialsId: 'oscontainer-push-registry-secret', variable: 'REGISTRY_SECRET')]) {
```
jlebon, line 827: We shouldn't hardcode the secret name here. I think probably best is to leave that to the calling job, which can pass in the registry file path as another parameter?

jobs/build-node-image.Jenkinsfile
```diff
@@ -71,16 +71,32 @@ lock(resource: "build-node-image") {
                                           ["--security-opt label=disable", "--mount-host-ca-certs", "--force"])
         }
 
-        stage("Push Manifest") {
-            pipeutils.push_manifest(arches, container_registry_staging_repo, container_registry_staging_image_tag, container_registry_staging_manifest_tag)
+        stage('Build Extension Container') {
+            pipeutils.build_remote_image(arches, commit, src_config_url, container_registry_staging_repo,
+                                         "${container_registry_staging_image_tag}-extensions", stream_info.yumrepo.file, false,
```
jlebon, line 76: One gotcha here that I hit when testing extensions building locally is that the extensions container currently needs the RHEL 9.4 repo as well: https://github.com/openshift/os/blob/26ecd2c01600e8b2532ea6edf6322c60dcac1846/extensions-ocp-rhel-9.6.yaml#L17

I think what we can do here is concatenate the `rhel-9.4.repo` and the `rhel-9.6.repo` repo into one file, and then pass that as the secret. Hmm, I guess to represent that in the pipecfg, maybe it should be a list of files that we concatenate instead of just one file?

jobs/build-node-image.Jenkinsfile
```diff
@@ -71,16 +71,32 @@ lock(resource: "build-node-image") {
...
+                                         ["--security-opt label=disable", "--mount-host-ca-certs", "--force",
+                                          "--git-file extensions/Dockerfile"] )
```
jlebon, line 78: `--git-containerfile` is what landed in the end

jobs/build-node-image.Jenkinsfile
jlebon, line 76: The `FROM` needs to be the built node image. Maybe simplest is to build the layered node image fully, but pushed to the staging repo, then build the extensions container (passing the manifest listed node image as the `FROM`), then push both to their final location in the prod repo? That way, we reduce the chance of one getting pushed but not the other because something went wrong.

To reconcile this with my earlier comment about having one high-level utility function. Let's call it e.g. `build_and_push_image(parameters..., push_destination)`. Then do something like (pseudocode):

```
build_and_push_image(params to build node image..., "registry.ci.openshift.org/coreos/node-staging:4.19-9.6-node-image")
build_and_push_image(params to build extensions image FROM node image..., "registry.ci.openshift.org/coreos/node-staging:4.19-9.6-node-image-extensions")
copy_image("registry.ci.openshift.org/coreos/node-staging:4.19-9.6-node-image", "quay.io/openshift-release-dev/ocp-v4.0-art-dev:${RELEASE}-node-image")
copy_image("registry.ci.openshift.org/coreos/node-staging:4.19-9.6-node-image-extensions", "quay.io/openshift-release-dev/ocp-v4.0-art-dev:${RELEASE}-node-image-extensions")
```

?

utils.groovy
```diff
@@ -820,4 +820,92 @@ def matrixSend(message) {
...
+             shwrap("""
+             cosa push-container-manifest \
+                 --auth=\$REGISTRY_SECRET --tag ${manifest_tag} \
+                 --repo ${repo} ${images}
+             """)
+         }
+     }
+}
+
+def release_manifest(staging_repo, staging_tag, prod_repo_and_tag) {
```
jlebon, line 880: Wonder if this would be clearer as just `copy_image()` instead? And probably the parameters can just be `src_image` and `dest_image`?

jobs/build-node-image.Jenkinsfile
```diff
@@ -55,7 +55,8 @@ lock(resource: "build-node-image") {
         def container_registry_staging_repo = pipecfg.ocp_node_builds.registries.staging
         def container_registry_repo_and_tag = utils.substituteStr(pipecfg.ocp_node_builds.registries.prod,
                                                                  [RELEASE: params.RELEASE])
-
+        def container_registry_staging_manifest_tag = "${params.RELEASE}-${shortcommit}"
+        def container_registry_staging_image_tag = "${params.RELEASE}-ARCH_REPLACE-${shortcommit}"
```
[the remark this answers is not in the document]
ravanelli: We can, maybe we should have it as default and have an override as well. What I think makes more sense.

utils.groovy
```diff
@@ -820,4 +820,79 @@ def matrixSend(message) {
...
+    if (secret) {
+         build_args += ["--secret", secret]
+    }
+    if (from) {
+        build_args += ["--from", from]
+    }
+    parallel arches.collectEntries { arch ->
+        [arch, {
+            pipeutils.withPodmanRemoteArchBuilder(arch: arch) {
+                def final_args = build_args + ["--arch", arch, "--tag", "${tag}-${arch}"] + extra_args
```
jlebon, line 837: Minor: could append `extra_args` once before entering `parallel`.

utils.groovy
```diff
@@ -820,4 +820,79 @@ def matrixSend(message) {
...
+    pipeutils.withPodmanRemoteArchBuilder(arch: "s390x") {
+        shwrap("""
+        cosa push-container-manifest \
+            --auth=${registry_secret_path} --tag ${manifest_tag} \
+            --repo ${repo} ${images}
+        """)
+    }
+}
+
+def copy_image(src_image, dest_image, registry_secret_path) {
```
jlebon, line 869: Let's make `registry_secret_path` optional? The caller might be already be passing it in via `REGISTRY_AUTH_FILE` (which... we could just use that too in `build-node-image`).

utils.groovy
```diff
@@ -820,4 +820,79 @@ def matrixSend(message) {
...
+
+    // Release the manifest and container images
```
jlebon, line 871: Would just drop these lines.

utils.groovy
```diff
@@ -820,4 +820,79 @@ def matrixSend(message) {
...
+    shwrap("""
+        skopeo copy --all --authfile=${registry_secret_path} \
+            docker://${src_image} \
+            docker://${dest_image}
+    """)
+}
+
+def delete_tags(arches, repo, image_tag, manifest_tag, registry_secret_path) {
+
```
jlebon, line 880: Minor: empty line

utils.groovy
```diff
@@ -820,4 +820,79 @@ def matrixSend(message) {
...
+        shwrap("""
+            export STORAGE_DRIVER=vfs # https://github.com/coreos/fedora-coreos-pipeline/issues/723#issuecomment-1297668507
+            skopeo delete --authfile=\$registry_secret_path \
+                docker://${repo}:${image_tag}-${arch}
+        """)
+    }]}
+}
+
+def build_and_push_image(arches, commit, url, staging_repo, staging_image_tag, staging_manifest_tag, registry_secret_path,
+                         secret=None, from=None, extra_args=None) {
```
jlebon: I think this would probably be cleaner using the options dict pattern. E.g. have a single `params = [:]` parameter, much like in a bunch of coreos-ci-lib calls.

Though totally fine too to leave it as is for now!

jlebon: `extra_args` -> `extra_build_args`?

jobs/build-node-image.Jenkinsfile
```diff
@@ -74,9 +76,28 @@ lock(resource: "build-node-image") {
...
 
-        stage("Release Manifest") {
+        stage('Build Extensions Container') {
+            withCredentials([file(credentialsId: 'oscontainer-push-registry-secret', variable: 'REGISTRY_SECRET')]) {
+                // Use the node image as from
+                def build_from = container_registry_staging_manifest
+                // The Extesions container is using two repositories for now, we need to merge them to pass it as a secret
+                def repo_1 = "\$(pwd)/yumrepos/${stream_info.yumrepo.file}"
+                def repo_2 = "\$(pwd)/yumrepos/rhel-9.4.repo"
+                shwrap(""" cat ${repo_1} ${repo_2} > /tmp/final_repo """)
```
jlebon, line 86: I think instead of this, let's change `stream_info.yumrepo.file` to be e.g. `stream_info.yumrepo.files` instead (i.e. a list). Which we then concatenate if it's > 1.

That way the OCP release-specific repo hack and context live in the pipecfg, not directly in the pipeline.

utils.groovy
```diff
@@ -835,16 +835,24 @@ def build_remote_image(arches, commit, url, repo, tag, registry_secret_path, sec
         [arch, {
             pipeutils.withPodmanRemoteArchBuilder(arch: arch) {
                 def final_args = build_args + ["--arch", arch, "--tag", "${tag}-${arch}"] + extra_args
+                if (store_digest) {
+                     final_args += ["--write-digest-to-file", "${tag}-${arch}"]
```
jlebon, line 839: Hmm, a bit odd that we "leak" the concept of that file across two separate functions.

How about instead, we have `build_remote_image` return the list of digests (and it can delete those files as well). And then in `build_and_push_image`, we just take that list and pass it to `push_manifest`?

I think if we do that, let's just have the logic always do it. I.e. let's not have any `store_digest` parameter on any of the functions.

jlebon (commented):
Love the direction this is going!

jobs/build-node-image.Jenkinsfile
```diff
@@ -74,16 +76,40 @@ lock(resource: "build-node-image") {
...
                                                 extra_build_args: ["--security-opt label=disable", "--mount-host-ca-certs", "--force"])
             }
         }
 
-        stage("Release Manifest") {
+        stage('Build Extensions Container') {
+            withCredentials([file(credentialsId: 'oscontainer-push-registry-secret', variable: 'REGISTRY_SECRET')]) {
+                // Use the node image as from
+                def build_from = container_registry_staging_manifest
+                // The Extesions container is using two repositories for now, we need to merge them to pass it as a secret
```
jlebon, line 89: Let's just drop this comment. It lives better in the pipecfg repo where we'll have to add the `rhel-9.4.repo` file to the list.

jobs/build-node-image.Jenkinsfile
```diff
@@ -65,67 +68,48 @@ lock(resource: "build-node-image") {
...
+
+        stage('Build Extensions Container') {
+            withCredentials([file(credentialsId: 'oscontainer-push-registry-secret', variable: 'REGISTRY_SECRET')]) {
+                // Use the node image as from
+                def build_from = container_registry_staging_manifest
+                // The Extesions container is using two repositories for now, we need to merge them to pass it as a secret
+                def repos = stream_info.yumrepo.files
+                for (repo in repos) {
+                    def yumrepo = "\$(pwd)/yumrepos/${repo}"
+                    shwrap(""" cat ${yumrepo} >> /tmp/final_repo """)
```
jlebon, line 93:
```suggestion
                    shwrap("cat yumrepos/${repo} >> /tmp/final_repo")
```

jobs/build-node-image.Jenkinsfile
```diff
@@ -65,67 +68,48 @@ lock(resource: "build-node-image") {
...
+                                               src_commit: commit,
+                                               src_url: src_config_url,
+                                               staging_repository: container_registry_staging_repo,
+                                               image_tag_staging: "${container_registry_staging_image_tag}-extensions",
+                                               manifest_tag_staging: "${container_registry_staging_manifest_tag}-extensions",
+                                               registry_secret_path: REGISTRY_SECRET,
+                                               secret: "id=yumrepos,src=/tmp/final_repo",
+                                               from: build_from,
+                                               extra_build_args: ["--security-opt label=disable", "--mount-host-ca-certs",
+                                               "--git-containerfile extensions/Dockerfile", "--force"])
```
jlebon, line 105:
```suggestion
                                               "--git-containerfile", "extensions/Dockerfile", "--force"])
```

utils.groovy
```diff
@@ -899,8 +904,8 @@ def build_and_push_image(params = [:]) {
     def from = params.get('from', "");
     def extra_build_args = params.get('extra_build_args', "");
 
-    build_remote_image(params['arches'], params['src_commit'], params['src_url'], params['staging_repository'],
-                       params['image_tag_staging'], params['registry_secret_path'], secret, from, extra_build_args)
+    def digests = build_remote_image(params['arches'], params['src_commit'], params['src_url'], params['staging_repository'],
+                                     params['image_tag_staging'], params['registry_secret_path'], secret, from, extra_build_args)
 
     stage("Push Manifest") {
         push_manifest(params['arches'], params['staging_repository'], params['image_tag_staging'],
```
jlebon, line 911: Don't we need to pass `digests` here?

jobs/build-node-image.Jenkinsfile
```diff
@@ -65,67 +68,48 @@ lock(resource: "build-node-image") {
...
-                    }
-                }]}
+                 def build_from = params.FROM ?: stream_info.from
+                 pipeutils.build_and_push_image(arches: arches,
+                                                src_commit: commit,
+                                                src_url: src_config_url,
+                                                staging_repository: container_registry_staging_repo,
+                                                image_tag_staging: container_registry_staging_image_tag,
+                                                manifest_tag_staging: container_registry_staging_manifest_tag,
+                                                registry_secret_path: REGISTRY_SECRET,
```
jlebon, line 78: Minor/optional: does it not work to just change `REGISTRY_SECRET` above to `REGISTRY_AUTH_FILE` and then drop this?

jlebon (commented):
Some minor comments, and but LGTM overall!

jlebon (approved):
This looks really great, thank you!

I took the liberty to push some tweaks on top. Assuming you agree and you've successfully tested this, feel free to merge!

PR #1108, opened 2025-03-19
jobs/build-development: make development streams build daily

dustymabe:
Over time I'm thinking the build on every push to the git repo is a bit heavy. Often times the changes will be insignificant to warrant building, testing, and pushing all that we do. On the other hand, if there is a change we want we can easily click a button and start a build if we don't want to wait.

This commit switches the development streams to build daily similarly to the mechanical ones (i.e. rawhide). It also changes the timing of the build-mechanical and bump-lockfiles jobs and adds a stage to the bump-lockfile job to trigger a build once it has pushed.

The timings are chosen based on:

- bump-lockfiles at 04:00 UTC
    - The bodhi/pungi updates composes seem to run around 0:00 UTC and finish within an hour or two. This gives some extra time so we hopefully trigger a bump-lockfile with the latest content.
- build-development at 05:00 UTC
    - Triggered soon after the bump-lockfiles job it will now wait if there are bump-lockfile jobs running before kicking off the development stream builds.
    - If the bump-lockfile runs failed (and thus didn't trigger a new build at the end of the run) this daily trigger will make sure we still get a build. If the bump-lockfile run finished and a new build was started then this build will wait and hopefully result in "No New Build".
- build-mechanical at 10:00 UTC
    - The rawhide composes seem to run around 5:00 UTC and finish around 7:00 UTC or 8:00 UTC. This runs a rawhide soon after with an hour or so of grace in case the pungi runs start taking more time.

### DIFF
--- a/jobs/build-development.Jenkinsfile
+++ b/jobs/build-development.Jenkinsfile
@@ -6,7 +6,10 @@ node {
 }
 
 properties([
-    pipelineTriggers(pipeutils.get_push_trigger()),
+    pipelineTriggers([
+        // run every 24h at 5:00 UTC
+        cron("0 5 * * *")
+    ]),
     buildDiscarder(logRotator(
         numToKeepStr: '100',
         artifactNumToKeepStr: '100'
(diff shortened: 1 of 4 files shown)


dustymabe:
ok this is an expanded version of https://github.com/coreos/fedora-coreos-pipeline/pull/973 with the same goal.

to bring forward a conversation piece from [last time](https://github.com/coreos/fedora-coreos-pipeline/pull/973#issuecomment-2026103903). @jlebon wrote:

Had a chat with Dusty about this. Things we mentioned:
- we're currently not building multiple times per day very often
- part of that though is probably due to Dusty manually intervening with the build job
- until we have multi-arch CI, this would mean that multi-arch bugs in newly introduced code/tests will not be found until e.g. the next day, vs a few hours later
- on promotion day, the latest testing-devel that we'd promote might not have been built and tested at all yet
- a good compromise would be to leave it to trigger on git pushes, but still force a rate-limit of e.g. 20h in the job itself

dustymabe:
> * we're currently not building multiple times per day very often
> * part of that though is probably due to Dusty manually intervening with the build job

agree. A lot of times I kill jobs because I know the changes are meaningless. I want to stop doing that.

 
> * until we have multi-arch CI, this would mean that multi-arch bugs in newly introduced code/tests will not be found until e.g. the next day, vs a few hours later

I think if we think this is important we should do multi-arch on CI instead of doing it later. I'm pretty sure we are capable of this now considering https://github.com/coreos/coreos-ci/blob/63a5f7e0beb42e0403d463e013618627453f07aa/jobs/test-override.Jenkinsfile#L170-L195
> 
> * on promotion day, the latest testing-devel that we'd promote might not have been built and tested at all yet

ehh. I'm not too concerned about this. the proposal here will have the bump-lockfile (which does tests) and the builds happening in early UTC and we don't usually start builds until later in the day UTC.

> 
> * a good compromise would be to leave it to trigger on git pushes, but still force a rate-limit of e.g. 20h in the job itself

I think I value predictability here. This ultimately is an attempt to gain some sanity in the monitoring of the pipeline. If we see a failure we know its the only one for that day and we can focus on it versus there being more builds coming and we have to navigate those too while we are investigating an already obsolete build.

jlebon (approved):
LGTM! Schedules seem sane. We can always tweak things as we go if needed.

PR #1109, opened 2025-03-21
node-image-build: add HOTFIX parameters

jbtrystram:
Allow specifying a HOTFIX reference and url for
pipecfg.

This will always append the hotfix name to the pushed tags so we don't overwrite anything.

See https://issues.redhat.com/browse/COS-3174

### DIFF
--- a/jobs/build-node-image.Jenkinsfile
+++ b/jobs/build-node-image.Jenkinsfile
@@ -25,13 +25,25 @@ properties([
              description: 'Override the coreos-assembler image to use',
              defaultValue: "quay.io/coreos-assembler/coreos-assembler:latest",
              trim: true),
-    ]),
+    ]+ pipeutils.add_hotfix_parameters_if_supported()),
     buildDiscarder(logRotator(
         numToKeepStr: '100',
         artifactNumToKeepStr: '100'
     )),
     durabilityHint('PERFORMANCE_OPTIMIZED')
 ])
+
+def build_description = "[${params.RELEASE}]"
+
+// Reload pipecfg if a hotfix build was provided. The reason we do this here
+// instead of loading the right one upfront is so that we don't modify the
+// parameter definitions above and their default values.
+if (params.PIPECFG_HOTFIX_REPO || params.PIPECFG_HOTFIX_REF) {
+    node {
+        pipecfg = pipeutils.load_pipecfg(params.PIPECFG_HOTFIX_REPO, params.PIPECFG_HOTFIX_REF)
+        build_description = "[${params.RELEASE}-${pipecfg.hotfix.name}]"
+    }
+}
 
 def stream_info = pipecfg.ocp_node_builds.release[params.RELEASE]
 def src_config_ref = stream_info.source_config.ref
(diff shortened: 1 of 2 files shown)


utils.groovy
```diff
@@ -550,6 +550,20 @@ def get_registry_repos(pipecfg, stream, version) {
     return registry_repos
 }
 
+def get_ocp_node_registry_repo(pipecfg, release) {
+    def staging_repo = pipecfg.ocp_node_builds.registries.staging
+    def repo_and_tag = utils.substituteStr(pipecfg.ocp_node_builds.registries.prod,
+    [RELEASE: release])
```
jlebon, line 556: Indentation

jobs/build-node-image.Jenkinsfile
```diff
@@ -134,12 +148,12 @@ lock(resource: "build-node-image") {
...
         if (currentBuild.result == 'SUCCESS') {
-            currentBuild.description = "[${params.RELEASE}@${shortcommit}] ⚡"
+            currentBuild.description = "[${build_description}@${shortcommit}] ⚡"
         } else {
-            currentBuild.description = "[${params.RELEASE}@${shortcommit}] ❌"
+            currentBuild.description = "[${build_description}@${shortcommit}] ❌"
         }
         if (currentBuild.result != 'SUCCESS') {
-            message = ":openshift: build-node-image #${env.BUILD_NUMBER} <${env.BUILD_URL}|:jenkins:> <${env.RUN_DISPLAY_URL}|:ocean:> [${params.RELEASE}][${src_config_ref}@${shortcommit}]"
+            message = ":openshift: build-node-image #${env.BUILD_NUMBER} <${env.BUILD_URL}|:jenkins:> <${env.RUN_DISPLAY_URL}|:ocean:> ${build_description}"
```
jlebon, line 156: This loses the git ref and shortcommit info.

I think instead the way I would structure this is that we keep all the info in `build_description`, and then whenever we update `currentBuild.description`, it's only to append the "state" part of it. I.e. [near the start](https://github.com/jbtrystram/fedora-coreos-pipeline/blob/43cfa03d6c776e3e49e1f5e243463f5ef007f355/jobs/build-node-image.Jenkinsfile#L63):

```
currentBuild.description = "${build_description} Running"
```

and then at the beginning of the `finally` just above

```
currentBuild.description = "${build_description} ⚡"
```

and then here we can just use `build_description` only as you're doing.

jobs/build-node-image.Jenkinsfile
```diff
@@ -25,14 +25,27 @@ properties([
...
+// Reload pipecfg if a hotfix build was provided. The reason we do this here
+// instead of loading the right one upfront is so that we don't modify the
+// parameter definitions above and their default values.
+if (params.PIPECFG_HOTFIX_REPO || params.PIPECFG_HOTFIX_REF) {
+    node {
+        pipecfg = pipeutils.load_pipecfg(params.PIPECFG_HOTFIX_REPO, params.PIPECFG_HOTFIX_REF)
+        build_description = "[${params.RELEASE}-${pipecfg.hotfix.name}]"
+    }
+}
+
```
jlebon, line 47: Minor: extra empty line.

jlebon (approved):
LGTM!

This conflicts with https://github.com/coreos/fedora-coreos-pipeline/pull/1107 which is also getting close. Let's chat about the ordering OOB.

PR #1111, opened 2025-03-22
multi-arch-builders/splunk: remove outdated RH Root CA Cert

marmijo:
The 2015 IT Root CA cert is no longer available and is causing the RHCOS multi-arch builders to fail to provision. Let's remove it from the butane config since the 2022 cert is sufficient.


dustymabe (approved):
LGTM

PR #1114, opened 2025-03-31
jobs/build-node-image: bump timeout

jlebon:
All arches finish the node image build in 2 or 3 minutes; ppc64le takes 14 minutes... We really need to get better performance there.

Anyway, since we also need to be able to build the extensions container, we should bump this anyway.

### DIFF
--- a/jobs/build-node-image.Jenkinsfile
+++ b/jobs/build-node-image.Jenkinsfile
@@ -53,7 +53,7 @@ lock(resource: "build-node-image") {
     cosaPod(image: params.COREOS_ASSEMBLER_IMAGE,
             memory: "512Mi", kvm: false,
             serviceAccount: "jenkins") {
-    timeout(time: 15, unit: 'MINUTES') {
+    timeout(time: 45, unit: 'MINUTES') {
     try {
 
         def output = shwrapCapture("git ls-remote ${src_config_url} ${src_config_ref}")


ravanelli:
It's tricky. I checked the server a long time ago, and the only issue I found was emulation with the 4K disk, even though it was already using a 4K disk. That has since been fixed. This is a borrowed server, so we can look for a new one.

We could also check how the new KVM setup works for Power. If I’m not mistaken, it’s not pure KVM but involves additional emulation. We’ll probably need to run some benchmarks to understand its behavior better. But at least it is not a limitation for P9 with kvm anymore.

jlebon:
At least for the node image build, there shouldn't be any KVM involved here at all. But yeah, in the other jobs, KVM settings might be a factor to its slowness.

jlebon:
For now at least, let's just get this one in?

ravanelli:
We probably need to understand where the new KVM is supported, looking at https://github.com/coreos/coreos-assembler/issues/2473 we have an example: `Power10 Rainier using firmware 1060.10 ` .  IBM cloud still an issue, I would guess, since the last time I tried,  it was a 'hell' to get things working there. 

https://github.com/coreos/fedora-coreos-tracker/issues/1627
https://github.com/coreos/fedora-coreos-tracker/issues/1628

ravanelli (approved):
LGTM

PR #1115, opened 2025-03-31
jobs/build-node-image: add `notsecret` tag to yumrepos secret line

jlebon:
Got an alert from a scanner that thought those lines contained secrets. Mark as `notsecret` to pacify it.


dustymabe (approved):
We would have git history.. but adding maybe one more word or two might help make it more readable/immediately obvious

// code scanner notsecret

PR #1116, opened 2025-03-31
jobs/build-node-image: send Slack message on success too

jlebon:
I think it's useful when a new job comes up to start with notifying Slack on success and only later downgrading to on failure only if warranted.

This gives us more insight into how often it runs, whether a rerun passes, or a patch to fix a recent failure worked, etc. And just in general time to evaluate whether that information is useful to us.


ravanelli (approved):
LGTM

PR #1118, opened 2025-04-04
jobs/release: support triggering build-node-image

jlebon:
We want the ability to trigger build-node-image from the release job
instead of the latter being cron triggered. That way we reduce the
chance of having compose skew when in pre-GA state, and we get fresh
node images soon after the base image is updated.

I had initially made the mechanism more generic, but meh, it didn't seem
worth the complexity. We can always generalize it if another "follow-up
trigger" case shows up.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -63,6 +63,9 @@ streams:
       # is equivalent to listing all live artifacts and cloud platforms in
       # `skip_artifacts`.
       skip_disk_images: true
+      # OPTIONAL: make the release job trigger build-node-image for the
+      # following OCP releases once successful
+      build_node_images: ["4.19-9.6"]
     stable:
       type: production
       # OPTIONAL: override cosa image to use for this stream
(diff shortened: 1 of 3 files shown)


ravanelli (approved):

docs/config.yaml
```diff
@@ -316,3 +291,32 @@ misc:
...
+               # REQUIRED: source config ref to build
+               ref: master
+               # REQUIRED: repo file name to build
+               file: rhel-9.6.repo
+    # REQUIRED: knobs related to the source config
+    registries:
+        # REQUIRED: staging repo to push the final image built
+        staging: registry.ci.openshift.org/coreos/node-staginear
+        # REQUIRED: prodution repo to push the final image built
+        prod: quay.io/openshift-release-dev/ocp-v4.0-art-dev:{RELEASE}-node-image
```
jbtrystram, line 321: needs to be updated / rebased with the tags changes from https://github.com/coreos/fedora-coreos-pipeline/pull/1120

jbtrystram (changes requested):
this also requires an update in the pipecfg

jbtrystram (approved):

PR #1120, opened 2025-04-08
utils: Add timestamp-based tag to avoid image cleanup issues in get_o…

ravanelli:
…cp_node_registry_repo

- Quay's garbage collection removes untagged images, which can result in lost node images when the same tag is reused. To ensure images persist throughout the CI lifecycle, we need a unique and stable tag.

- While multiple node images may use the same image base (rhel-9.6-coreos), we can't rely solely on that to differentiate them.

- This change introduces a timestamp-based tag (yyyyMMddHHmm) to guarantee uniqueness for each node image build.

- The timestamp is generated at runtime and locked, ensuring no collisions and preserving CI consistency.

### DIFF
--- a/jobs/build-node-image.Jenkinsfile
+++ b/jobs/build-node-image.Jenkinsfile
@@ -67,7 +67,7 @@ lock(resource: "build-node-image") {
         def archinfo = arches.collectEntries{[it, [:]]}
         def now = java.time.LocalDateTime.now()
         def timestamp = now.format(java.time.format.DateTimeFormatter.ofPattern("yyyyMMddHHmm"))
-        def (container_registry_staging_repo, container_registry_repo_and_tag) = pipeutils.get_ocp_node_registry_repo(pipecfg, params.RELEASE, timestamp)
+        def (container_registry_staging_repo, container_registry_repo, prod_tags) = pipeutils.get_ocp_node_registry_repo(pipecfg, params.RELEASE, timestamp)
         def container_registry_staging_manifest_tag = "${params.RELEASE}"
         def container_registry_staging_image_tag = "${params.RELEASE}"
         def container_registry_staging_manifest = "${container_registry_staging_repo}:${container_registry_staging_manifest_tag}"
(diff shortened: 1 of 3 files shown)


jobs/build-node-image.Jenkinsfile
```diff
@@ -98,7 +100,8 @@ lock(resource: "build-node-image") {
                                                 manifest_tag_staging: container_registry_staging_manifest_tag,
                                                 secret: "id=yumrepos,src=${yumrepos_file}", // notsecret (for secret scanners)
                                                 from: build_from,
-                                                extra_build_args: ["--security-opt label=disable", "--mount-host-ca-certs", "--force"])
+                                                extra_build_args: ["--security-opt label=disable", "--mount-host-ca-certs", "--force", 
+                                                "--label=org.opencontainers.image.version=${container_registry_staging_image_tag}.${timestamp}"])
```
jbtrystram, line 104: Should we also set `io.openshift.build.versions` ?

jbtrystram (approved):
LGTM ! But i would squash the two commits as some code is added then deleted.

dustymabe:
What I'd really like to do here is push to the same place we were in the past, but then also to this extra timestamp based tag. Can we do that?

jobs/build-node-image.Jenkinsfile
joepvd: Both of these are important:
```
  "io.openshift.build.version-display-names": "machine-os=Red Hat Enterprise Linux CoreOS",
  "io.openshift.build.versions": "machine-os=419.96.202503251418-0",
```
...as the get consumed by `oc adm release new` to create the `displayVersions."machine-os"` attribute in the resulting payload.

joepvd: ☝🏼 is wrong. It is intentional that these two labels are retained from the rhel layer.

ravanelli:
lgtm JB changes as well!

jobs/build-node-image.Jenkinsfile
```diff
@@ -119,9 +121,20 @@ lock(resource: "build-node-image") {
...
+                // that it's ready for release. So we want all the expected artifacts
+                // to be available when the ART tooling kicks in.
+                for ( tag in prod_tags ) {
+                    pipeutils.copy_image("${container_registry_staging_manifest}-extensions",
+                                     "${container_registry_repo}:${tag}-extensions")
+                }
+
+                // Skopeo does not support pushing multiple tags at the same time
+                // So we just recopy the same image multiple times.
+                // https://github.com/containers/skopeo/issues/513
```
dustymabe, line 134: I feel like this comment should be at the top of this function

PR #1121, opened 2025-04-10
jobs/build-node-image: Add a tag field for the staging repo

ravanelli:
- Add a tag field for the staging repo in the node image: We're currently using the same repo for both production and staging, which causes the manifest to be updated in devel tests what makes it unreliable for production builds.

### DIFF
--- a/jobs/build-node-image.Jenkinsfile
+++ b/jobs/build-node-image.Jenkinsfile
@@ -66,10 +66,12 @@ lock(resource: "build-node-image") {
         def archinfo = arches.collectEntries{[it, [:]]}
         def now = java.time.LocalDateTime.now()
         def timestamp = now.format(java.time.format.DateTimeFormatter.ofPattern("yyyyMMddHHmm"))
-        def (container_registry_staging_repo, container_registry_repo, prod_tags) = pipeutils.get_ocp_node_registry_repo(pipecfg, params.RELEASE, timestamp)
-        def container_registry_staging_manifest_tag = "${params.RELEASE}"
-        def container_registry_staging_image_tag = "${params.RELEASE}"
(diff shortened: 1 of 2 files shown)


jobs/build-node-image.Jenkinsfile
```diff
@@ -66,8 +66,8 @@ lock(resource: "build-node-image") {
         def archinfo = arches.collectEntries{[it, [:]]}
         def now = java.time.LocalDateTime.now()
         def timestamp = now.format(java.time.format.DateTimeFormatter.ofPattern("yyyyMMddHHmm"))
-        def (container_registry_staging_repo, container_registry_repo, prod_tags) = pipeutils.get_ocp_node_registry_repo(pipecfg, params.RELEASE, timestamp)
-        def container_registry_staging_manifest_tag = "${params.RELEASE}"
+        def (container_registry_staging_repo, container_registry_staging_manifest_tag,
```
jbtrystram, line 69: variables names are getting long. what about 
```suggestion
        def (staging_image, staging_tags,
        prod_image, prod_tags) 
```

jbtrystram (dismissed):
I think there is a confusion between string and list for `tags`

utils.groovy
```diff
@@ -927,7 +937,7 @@ def build_and_push_image(params = [:]) {
     def digests = build_remote_image(params['arches'], params['src_commit'], params['src_url'], params['staging_repository'],
```
dustymabe, line 937: can we delete `manifest_tag_staging` as an argument to this function. It's really not needed.

ravanelli: Where?

dustymabe: I think maybe I'll just open a new PR next week. We can go with this for now.

dustymabe (approved):

PR #1122, opened 2025-04-11
build: adjust timeout to 300m

dustymabe:
The build-arch job is already at 300 and we just lost a built because the signing of the OSTree commit took an hour. Sigh..


dustymabe:
Merging this because I'm angry.

PR #1123, opened 2025-04-11
config: enable osbuild_experimental for testing/testing-devel

dustymabe:
F42 GA is GO for next Tuesday so we'll go ahead and enable more artifacts for OSBuild building for those streams.


marmijo (approved):
LGTM

PR #1125, opened 2025-04-16
jobs:build-node-image: support publishing v2s2 manifests

jbtrystram:
Some ART tooling does not support OCI manifests yet, so allow creating the manifests in v2s2 support.

This is already supported by COSA so it's just
passing the correct flag and tie it to a knob.
Defaults to false.


jlebon (approved):
Thanks!

PR #1128, opened 2025-04-17
utils.groovy: use `--preserve-digests` during `skopeo copy`

jlebon:
We want to ensure that the same digests from the base layers are kept so that we get good deduplication for the initial pivot and also across updates. Make sure that the digests are retained.

Part of https://issues.redhat.com/browse/COS-3144.


jbtrystram (approved):
LGTM

PR #1129, opened 2025-04-17
jobs/build-node-image: label images with manifest list tag

jlebon:
ART needs to be able to, from an x86_64 nightly, derive an s390x version of it. They used to be able to do this before the layering work by looking at the build ID, but now the source of truth is the manifest list, which is a moving target.

Enable this by adding the manifest list tag to all the built images. This feels like a layering violation of course. I think the more appropriate fix for this is to better keep track of the manifest lists used from the start when building arch-specific nightlies, but for now at least let's not break current workflows. And anyway, all this will have to be reworked for Konflux.

### DIFF
--- a/jobs/build-node-image.Jenkinsfile
+++ b/jobs/build-node-image.Jenkinsfile
@@ -68,6 +68,17 @@ lock(resource: "build-node-image") {
         def timestamp = now.format(java.time.format.DateTimeFormatter.ofPattern("yyyyMMddHHmm"))
         def (registry_staging_repo, registry_staging_tags, registry_prod_repo, registry_prod_tags) = pipeutils.get_ocp_node_registry_repo(pipecfg, params.RELEASE, timestamp)
 
+        // Get the tag that's unique
+        def unique_tag = ""
+        for (tag in registry_prod_tags) {
+            if (tag.contains(timestamp)) {
+                if (unique_tag != "") {
+                    error("Multiple unique tags found: ${registry_prod_tags}")
+                }
+                unique_tag = tag
+            }
+        }
+
         // `staging_tags` is a list to stay consistent with the `prod` objects,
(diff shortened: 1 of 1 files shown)


ravanelli (approved):
LGTM

jobs/build-node-image.Jenkinsfile
```diff
@@ -68,6 +68,20 @@ lock(resource: "build-node-image") {
...
+            if (tag.contains(timestamp)) {
+                if (unique_tag != "") {
+                    error("Multiple unique tags found: ${registry_prod_tags}")
+                }
+                unique_tag = tag
+            }
+        }
+        if (unique_tag == "") {
+            error("No unique tag found: ${registry_prod_tags}")
+        }
```
dustymabe, line 83: hmm. I think we configured the devel RHCOS pipeline to not push a unique tag (i.e. we actually wanted GC to clean up).

If that's true this will fail in the devel RHCOS pipeline.

dustymabe (approved):
LGTM assuming the concern I raised isn't a problem

jobs/build-node-image.Jenkinsfile
ravanelli: In this case the unique tag, is the one using the timestamp, so it won't be GC

dustymabe: I think you are missing my point.

In the configuration for the devel pipeline there is no unique (unique to this `build-node-image` run) label.

So the result after we merge this PR would be:

```
ERROR: No unique tag found: 4.19-node-image-devel
```

jlebon: Fixed, thanks for catching!

jlebon:
Just going to merge this in to try to unblock things on the ART side. We can do fixups as followups if needed.

PR #1130, opened 2025-04-21
config: drop osbuild_experimental from most streams

dustymabe:
Since https://github.com/coreos/coreos-assembler/pull/4074 most platforms have now been switched to default to osbuild for image building. Let's drop `osbuild_experimental: true` from all streams except for `rawhide` where the next round of experimental osbuild platforms should land first.

Here, we also add it to `stable` stream just in case we do the next build of `stable` (i.e. the first F42 build) with the same COSA that we used for the first F42 `testing` build, which wouldn't have the changed defaults in it.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -1,25 +1,21 @@
 streams:
   stable:
     type: production
+    osbuild_experimental: true
   testing:
     type: production
-    osbuild_experimental: true
   next:
     type: production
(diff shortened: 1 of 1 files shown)


marmijo (approved):
LGTM

aaradhak (approved):

PR #1131, opened 2025-04-21
config.yaml: disable next-devel 2025-04-21

dustymabe:
Now that testing-devel is moving back over to F42 we don't have any differences in testing-devel and next-devel so we'll disable it.


marmijo (approved):
LGTM

PR #1134, opened 2025-04-23
libcloud: build and replicate aws-winli images for supported streams

marmijo:
Expand the replicate to clouds functionality to build AWS Windows License Included (winli) images for streams that support it before replicating AMIs to other regions. `cosa aws-replicate` will now replicate both traditional AMIs and aws-winli AMIs if present in the metadata.

See: https://github.com/coreos/coreos-assembler/pull/4069

Also add a `winli` knob to the pipeline config clouds.aws section to only build winli images for specific streams.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -119,6 +119,8 @@ streams:
       # When mechanical streams have `scheduled` on, then
       # `build-mechanical` will only build those streams
       scheduled: true
+      # OPTIONAL: Create and replicate AWS Windows License Included images for this stream.
+      create_and_replicate_winli_ami: true
 
 # REQUIRED: architectures to build for other than x86_64
 additional_arches: [aarch64, ppc64le, s390x]
--- a/libcloud.groovy
+++ b/libcloud.groovy
@@ -6,7 +6,9 @@ def replicate_to_clouds(pipecfg, basearch, buildID, stream) {
         cosa meta --build=${buildID} --arch=${basearch} --dump
     """))
     def replicators = [:]
+    def builders = [:]
     def credentials
+    def stream_info = pipecfg.streams[stream]
 
     credentials = [file(variable: "ALIYUN_IMAGE_UPLOAD_CONFIG",
                         credentialsId: "aliyun-image-upload-config")]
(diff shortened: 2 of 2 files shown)


docs/config.yaml
```diff
@@ -208,6 +208,9 @@ clouds:
       # OPTIONAL: boolean: whether or not to set the image/snapshot as public on
       # initial upload (AMIs of released builds are always made public)
       public: true
+    # OPTIONAL: list of streams to build and replicate AWS Windows License Included images for
+    winli:
+      - "4.19-9.6"
```
marmijo, line 213: Is there a better way to implement this? AIUI, our stream approach might not make much sense anymore in the RHCOS case.

libcloud.groovy
```diff
@@ -46,10 +48,76 @@ def replicate_to_clouds(pipecfg, basearch, buildID, stream) {
...
+        def creds = [file(variable: "AWS_CONFIG_FILE", credentialsId: credentialId)]
+        withCredentials(creds) {
+            utils.syncCredentialsIfInRemoteSession(["AWS_CONFIG_FILE"])
+            def c = config
+
+            // Discover the latest Windows Server AMI to use as the winli-builder instance.
+            // The AMI rotates frequently with Windows updates and is not persistant in AWS
+            // for very long, so we need to find the most recent AMI ID.
+            // Windows Server 2022 was selected here because Windows Server 2025 does not
+            // support legacy bios. WS2022 has an EOL date of 2026-10-13
```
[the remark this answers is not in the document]
marmijo: I believe it might be. The ROSA team was not able to properly create a cluster with a winli image based on windows server 2025.

libcloud.groovy
```diff
@@ -46,10 +48,76 @@ def replicate_to_clouds(pipecfg, basearch, buildID, stream) {
...
+            }
+            shwrap("""
+                cosa ${image_upload_cmd}-aws \
+                    --upload \
+                    --winli \
+                    --windows-ami=${windows_server_ami} \
+                    --arch=${basearch} \
+                    --build=${buildID} \
+                    --region=${c.primary_region} \
+                    --bucket=s3://${c.bucket} \
```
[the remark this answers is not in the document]
marmijo: We don't need to upload it to S3, and it actually doesn't do anything in the winli case. This option is required for the traditional aws upload and is pretty baked into the code. I opted to just leave it instead of creating more if/else branches in ore.

libcloud.groovy
```diff
@@ -46,10 +48,76 @@ def replicate_to_clouds(pipecfg, basearch, buildID, stream) {
...
+            // Discover the latest Windows Server AMI to use as the winli-builder instance.
+            // The AMI rotates frequently with Windows updates and is not persistant in AWS
+            // for very long, so we need to find the most recent AMI ID.
+            // Windows Server 2022 was selected here because Windows Server 2025 does not
+            // support legacy bios. WS2022 has an EOL date of 2026-10-13
+            // https://learn.microsoft.com/en-us/lifecycle/products/windows-server-2022
+            def windows_server_ami_name = ""
+            windows_server_ami_name = "Windows_Server-2022-English-Full-Base-*"
+            def windows_server_ami = shwrapCapture("""
+                export AWS_CONFIG_FILE=\${AWS_BUILD_UPLOAD_CONFIG}
```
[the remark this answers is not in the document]
marmijo: I'm not 100% sure. I copied the format of other shWrap type calls in utils.groovy: [example](https://github.com/coreos/fedora-coreos-pipeline/blob/8761827ec89afb2ef7e340def1e1cf15fae1e99d/utils.groovy#L188), [example](https://github.com/coreos/fedora-coreos-pipeline/blob/8761827ec89afb2ef7e340def1e1cf15fae1e99d/utils.groovy#L241)

libcloud.groovy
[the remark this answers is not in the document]
dustymabe: Is there an issue that exists to track that limitation (not on our side necessarily), but on the ROSA side?

Either way we're going to bump up into that EOL at some point and having some place here in the comment to click through to would be nice.

libcloud.groovy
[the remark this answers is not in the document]
dustymabe: yeah. those use `shwrapWithAWSBuildUploadCredentials()`, which uses `withAWSBuildUploadCredentials()` which sets `variable: 'AWS_BUILD_UPLOAD_CONFIG'`. 

Here you are setting `def creds = [file(variable: "AWS_CONFIG_FILE", credentialsId: credentialId)]` so `AWS_CONFIG_FILE` should already be populated.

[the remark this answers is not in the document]
marmijo: Perfect! I'll remove the export. Thanks.

libcloud.groovy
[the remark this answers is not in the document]
dustymabe: if we delete `--bucket` here will `cosalib/aws.py` then just pass `--bucket ""` to `ore`? Will that be OK since `ore` does populate a default value for uploadBucket?

Having it in the code here is misleading IMO. Having things be confusing can lead to wasting time.

In the very least we should add a comment here about why we are adding it when it's not really going to be used.

libcloud.groovy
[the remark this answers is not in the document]
marmijo: @dustymabe and I spoke about this. The Windows Server 2025 AMI requires booting in UEFI mode, which restricts the possible instance types to those that support UEFI. This boot mode isn't a requirement for our RHCOS AMIs, and we would like to continue supporting as many instance types as possible. We'll continue using Windows Server 2022 and if we need to support BIOS for our winli images when it's no longer available, we can use one of our `aws-winli` images as the "builder" by passing it to `--windows-ami` using `cosa imageupload-aws`.

dustymabe (approved):
LGTM

libcloud.groovy
```diff
@@ -46,10 +48,92 @@ def replicate_to_clouds(pipecfg, basearch, buildID, stream) {
...
+                extraArgs += c.grant_users.collect{"--grant-user-snapshot=${it}"}
+            }
+            if (c.tags) {
+                extraArgs += c.tags.collect { "--tags=${it}" }
+            }
+            if (c.public) {
+                extraArgs += "--public"
+            }
+            shwrap("""
+                cosa ${image_upload_cmd}-aws \
```
dustymabe, line 113: Since we don't need to worry about backwards compat here we can juts go ahead and:
```suggestion
                cosa imageupload-aws \
```

libcloud.groovy
```diff
@@ -46,10 +48,92 @@ def replicate_to_clouds(pipecfg, basearch, buildID, stream) {
...
+            // fetch the aws vmdk image from s3 because ore upload still requires it in the
+            // builds directory. We're already working in an initialized cosa working
+            // directory, so we only need to perform the buildfetch and decompress here.
+            def s3_stream_dir = pipeutils.get_s3_streams_dir(pipecfg, stream)
+            shwrap("""
+                cosa buildfetch --build=${buildID} \
+                --arch=x86_64 --artifact=aws --url=s3://${s3_stream_dir}/builds \
+                --aws-config-file \${AWS_CONFIG_FILE} \
+                && cosa decompress --build=${buildID} --artifact=aws
+            """)
```
dustymabe, line 74: Since this is an artificial requirement in this case how about we just `touch /path/to/foo.vmdk.gz` here and then delete the empty file when we're done with the upload

marmijo: SGTM, I updated the PR to grab the file name from the metadata and then `touch` it just before creating the AMI.

libcloud.groovy
```diff
@@ -46,10 +48,89 @@ def replicate_to_clouds(pipecfg, basearch, buildID, stream) {
...
+    // aws-winli is only supported on x86_64.
+    def awsWinLIBuildClosure = { config, aws_image_name, credentialId ->
+        def creds = [file(variable: "AWS_CONFIG_FILE", credentialsId: credentialId)]
+        withCredentials(creds) {
+            utils.syncCredentialsIfInRemoteSession(["AWS_CONFIG_FILE"])
+            def c = config
+
+            // Since we are not uploading anything, let's just touch the vmdk image
+            // file to satisfy ore aws upload, which still requires the file in the
+            // cosa working dir.
```
dustymabe, line 63: to be clear here. it's not actually `ore` it's the cosalib wrapper that checks to see if we "have the artifact" before attempting to upload.

libcloud.groovy
```diff
@@ -46,10 +48,89 @@ def replicate_to_clouds(pipecfg, basearch, buildID, stream) {
...
+            def s3_stream_dir = pipeutils.get_s3_streams_dir(pipecfg, stream)
+            shwrap("""
+                touch builds/${buildID}/${basearch}/${aws_image_name}
```
dustymabe, line 66: Should we clean this up when we're done running `cosa imageupload-aws`? Not sure if it might cause any issues (i.e. wouldn't want this empty file to get uploaded back overtop the actual file somehow).

dustymabe (approved):
LGTM.

WDYT about waiting to merge this when the FCOS releases are done?

libcloud.groovy
marmijo: That's a good idea. I updated the PR to include a `rm` command on it. I dont think it would be uploaded since this is the replicate portion of the code, but it's better to be safe!

libcloud.groovy
marmijo: Thanks! I updated the comments.

libcloud.groovy
```diff
@@ -46,10 +48,95 @@ def replicate_to_clouds(pipecfg, basearch, buildID, stream) {
...
+    def awsWinLIBuildClosure = { config, aws_image_name, credentialId ->
+        def creds = [file(variable: "AWS_CONFIG_FILE", credentialsId: credentialId)]
+        withCredentials(creds) {
+            utils.syncCredentialsIfInRemoteSession(["AWS_CONFIG_FILE"])
+            def c = config
+
+            // Since we are not uploading anything, let's just touch the vmdk image
+            // file to satisfy the cosa ore wrapper, which still requires the file 
+            // in the cosa working dir.
+            aws_image_path = "builds/${buildID}/${basearch}/${aws_image_name}"
```
dustymabe, line 64:
```suggestion
            def aws_image_path = "builds/${buildID}/${basearch}/${aws_image_name}"
```

??

dustymabe (approved):

PR #1135, opened 2025-04-28
bump-lockfile: sync timeout with recent change to build.Jenkinsfile

nikita-dubrovskii:
The build job is already at 300, adjust timeout here to the same value.

### DIFF
--- a/jobs/bump-lockfile.Jenkinsfile
+++ b/jobs/bump-lockfile.Jenkinsfile
@@ -63,7 +63,7 @@ def getLockfileInfo(lockfile) {
 // Keep in sync with build.Jenkinsfile
 def cosa_memory_request_mb = 10.5 * 1024 as Integer
 def ncpus = ((cosa_memory_request_mb - 512) / 1536) as Integer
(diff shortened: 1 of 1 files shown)


dustymabe:
> The build job is already at 300, adjust timeout here to the same value.

Hmm.

The bump-lockfile job does much less than the `build` job does, though. It doesn't need to compress all the artifacts, wait on signing, or upload to clouds. Do you know why we need to bump this?

nikita-dubrovskii:
>  Do you know why we need to bump this?

According to comment https://github.com/coreos/fedora-coreos-pipeline/blob/main/jobs/bump-lockfile.Jenkinsfile#L63 
```
// Keep in sync with build.Jenkinsfile
```

dustymabe:
ok, but is that the only reason?

I'm trying to understand if you have seen cases where the existing timeout has been hit

nikita-dubrovskii:
> ok, but is that the only reason?

Yes, this is just to adjust the timeout value among `build-ABC`

dustymabe (approved):

PR #1137, opened 2025-04-29
jobs/nodeimage: use the base image digest for FROM

jbtrystram:
Instead of relying on the tag of the base image, first grab the manifest digest than pass that value to the builders.
This will avoid reusing stale images that may be present in the local containers storage when building.

### DIFF
--- a/jobs/build-node-image.Jenkinsfile
+++ b/jobs/build-node-image.Jenkinsfile
@@ -118,6 +118,7 @@ lock(resource: "build-node-image") {
                                                 secret: "id=yumrepos,src=${yumrepos_file}", // notsecret (for secret scanners)
                                                 from: build_from,
                                                 v2s2: v2s2,
+                                               check_base_match: true,
                                                 extra_build_args: ["--security-opt label=disable", "--mount-host-ca-certs", "--force"] + label_args)
             }
         }
--- a/utils.groovy
+++ b/utils.groovy
@@ -933,19 +933,54 @@ def build_and_push_image(params = [:]) {
     // src_url:              string  -- Source Git URL.
     // staging_repository:   string  -- Repository URL for the staging repo.
     // v2s2:                 bool    -- Use docker v2s2 format rather than OCI. Default to false.
+    // check_base_match      bool    -- Wether to check if the base image is the same accross all arches. Default to false.
 
     def secret = params.get('secret', "");
     def from = params.get('from', "");
     def manifest_digest = ""
     def extra_build_args = params.get('extra_build_args', "");
     def v2s2 = params.get('v2s2', false);
+    def arches = params['arches']
+    def check_base_match = params.get('check_base_match', false)
 
-    def digests = build_remote_image(params['arches'], params['src_commit'], params['src_url'], params['staging_repository'],
+    def from_manifest = skopeo_inspect(from)
+    if (!from.contains('@')) {
+        // pin the FROM with the manifest digest, to avoid arch remote builders reusing stale images
+        from = "${from.split(':')[0]}@${from_manifest["Digest"]}"
+    }
+    // also save the base image buildid
+    def base_build_id = from_manifest["Labels"]["org.opencontainers.image.version"]
+
+    def digests = build_remote_image(arches, params['src_commit'], params['src_url'], params['staging_repository'],
                                      params['image_tag_staging'], secret, from, extra_build_args)
+
     stage("Push Manifest") {
         manifest_digest = push_manifest(digests, params['staging_repository'], params['manifest_tag_staging'], v2s2)
     }
+
+    // sanity check: all the images should have the same base image
+    if (check_base_match) {
+        for (arch in arches) {
+            def manifest = skopeo_inspect("${params['staging_repository']}@${manifest_digest}", arch=arch)
+            def id = manifest["Labels"]["org.opencontainers.image.version"]
+            if (id != base_build_id) {
+                throw new Exception("Mismatched base image build id for ${arch}. Expected ${base_build_id}, got ${id}")
+            }
+        }
+    }
     return manifest_digest
+}
+
+def skopeo_inspect(image, arch = "") {
+
+    // goarch -> podman arches
+    def podarch  = ["aarch64":"arm64", "x86_64":"amd64", "s390x":"s390x", "ppc64le":"ppc64le"]
+    def arch_arg = ""
+    if (arch != "") {
+        arch_arg = "--override-arch ${podarch[arch]}"
+    }
+
+    return readJSON(text: shwrapCapture("skopeo inspect --authfile \$REGISTRY_AUTH_FILE --no-tags ${arch_arg} docker://${image}"))
 }
 
 return this


dustymabe (approved):
This LGTM

One question I have is what happens if the most recent run of `rhel-9.6` pushed only for some architectures? I imagine this whole process fails (and should fail)?

jobs/build-node-image.Jenkinsfile
```diff
@@ -109,16 +109,41 @@ lock(resource: "build-node-image") {
...
 
+                // pull the base RHCOS image and get it's digest and build ID
+                def base_image_manifest = readJSON(text: shwrapCapture("""
+                                                    skopeo inspect --authfile \$REGISTRY_AUTH_FILE --no-tags \
+                                                    docker://$build_from
+                                                    """))
+                def base_image_digest = base_image_manifest["Digest"]
+                def base_build_id = base_image_manifest["Labels"]["org.opencontainers.image.version"]
+
+                build_from = build_from.split(':')[0]
```
ravanelli, line 120: minor, fix indentation

jobs/build-node-image.Jenkinsfile
```diff
@@ -109,16 +109,41 @@ lock(resource: "build-node-image") {
...
+                                                skopeo inspect --authfile \$REGISTRY_AUTH_FILE --no-tags \
+                                                --override-arch ${podarch[arch]} \
+                                                docker://${registry_staging_repo}@${node_image_manifest_digest} \
+                                                | jq -r '.Labels."org.opencontainers.image.version"'
+                                            """)
+                     if (base != base_build_id) {
+                        throw new Exception("Mismatched base image build id for ${arch}. Expected ${base_build_id}, got ${base}")
+                    }
+                }
+
```
ravanelli, line 146: Same, here, we should move this part to the library-  utils

jobs/build-node-image.Jenkinsfile
ravanelli, line 120: It should be part of `build_and_push_image/build_remote_image`, the idea is to keep as much as possible in the libs. Let's create a new function to get it.

jbtrystram:
> I will note here that both this and #1121 could have been handled if we just used `--pull=always` in the underlying libraries when we do the build, but having extra checks here is good too.

This was obviously the first thing i tried, but it requires passing through the arg in the cosa wrapper. So i went this route, which makes it extra safer :)

utils.groovy
```diff
@@ -939,13 +939,44 @@ def build_and_push_image(params = [:]) {
...
+        def manifest = skopeo_inspect("${params['staging_repository']}@${manifest_digest}", arch=arch)
+        def id = manifest["Labels"]["org.opencontainers.image.version"]
+        if (id != base_build_id) {
+            throw new Exception("Mismatched base image build id for ${arch}. Expected ${base_build_id}, got ${id}")
+        }
+    }
     return manifest_digest
 }
 
+def skopeo_inspect(image, arch = "") {
```
ravanelli, line 970: How about we return everything we need here? and  do something like: 

```suggestion 
def getBaseImageInfo(buildFrom, arch "") {
   ...
   
    def manifestJson = shwrapCapture("""
        skopeo inspect --authfile ${registryAuthFile} --no-tags ${arch_arg}  docker://${image}
    """)
    ...
    
    def manifest = readJSON(text: manifestJson)
    def digest = manifest["Digest"]
    def buildId = manifest["Labels"]["org.opencontainers.image.version"]

    return [digest: digest, buildId: buildId]
} 
```

In this way we do not need to look twice for ```manifest["Labels"]["org.opencontainers.image.version"]``` for example.

jbtrystram: the only thing this really allows to refactor is `["Labels"]["org.opencontainers.image.version"]` at the cost of the helper not being generic anymore

utils.groovy
```diff
@@ -933,19 +933,54 @@ def build_and_push_image(params = [:]) {
...
+
+    // goarch -> podman arches
+    def podarch  = ["aarch64":"arm64", "x86_64":"amd64", "s390x":"s390x", "ppc64le":"ppc64le"]
+    def arch_arg = ""
+    if (arch != "") {
+        arch_arg = "--override-arch ${podarch[arch]}"
+    }
+
+    return readJSON(text: shwrapCapture("skopeo inspect --authfile \$REGISTRY_AUTH_FILE --no-tags ${arch_arg} docker://${image}"))
```
jlebon, line 983: Shouldn't need `--authfile $REGISTRY_AUTH_FILE` here; skopeo already knows about that environment variable.

utils.groovy
```diff
@@ -933,19 +933,54 @@ def build_and_push_image(params = [:]) {
...
     def extra_build_args = params.get('extra_build_args', "");
     def v2s2 = params.get('v2s2', false);
+    def arches = params['arches']
+    def check_base_match = params.get('check_base_match', false)
 
-    def digests = build_remote_image(params['arches'], params['src_commit'], params['src_url'], params['staging_repository'],
+    def from_manifest = skopeo_inspect(from)
+    if (!from.contains('@')) {
+        // pin the FROM with the manifest digest, to avoid arch remote builders reusing stale images
+        from = "${from.split(':')[0]}@${from_manifest["Digest"]}"
```
jlebon, line 949: This will be an arch-specific digest. We need the manifest list digest.

Which actually... I don't see a way to get that from skopeo.

OK, this seems to work:

```
$ skopeo inspect --raw -n docker://quay.io/openshift-release-dev/ocp-v4.0-art-dev:4.19-9.6-node-image > manifest.json
$ skopeo manifest-digest manifest.json
sha256:5fbb06ed936e20174a301c9c4e2dcf17a2e739fa223630440c2868cb7f45c1c5
```

Let's drop the `skopeo_inspect` function and replace it with a `get_manifest_digest()`?

utils.groovy
```diff
@@ -933,19 +933,54 @@ def build_and_push_image(params = [:]) {
...
+
+    def digests = build_remote_image(arches, params['src_commit'], params['src_url'], params['staging_repository'],
                                      params['image_tag_staging'], secret, from, extra_build_args)
+
     stage("Push Manifest") {
         manifest_digest = push_manifest(digests, params['staging_repository'], params['manifest_tag_staging'], v2s2)
     }
+
+    // sanity check: all the images should have the same base image
+    if (check_base_match) {
```
jlebon, line 962: I think if we use the manifest list digest, we shouldn't need this sanity-check.

utils.groovy
jlebon, line 949: This will run even if no `from` was provided.

utils.groovy
```diff
@@ -933,19 +933,54 @@ def build_and_push_image(params = [:]) {
...
+        for (arch in arches) {
+            def manifest = skopeo_inspect("${params['staging_repository']}@${manifest_digest}", arch=arch)
+            def id = manifest["Labels"]["org.opencontainers.image.version"]
```
jlebon, line 965: This is assuming that the image has a version tag and that it's meaningful. E.g. the cosa container just has `42`.

I think let's drop this whole `check_base_match` thing. The manifest list digest enforcement is stronger than this.

dustymabe:
obsoleted by https://github.com/coreos/fedora-coreos-pipeline/pull/1138

utils.groovy
jlebon: OK sorry, I was wrong here. The `.Digest` field of a `skopeo inspect docker://manifest/listed/image` _is_ the manifest list digest.

One interesting semi-gotcha though is that `--override-arch` doesn't affect the reported digest at all (i.e. it's still the digest of the manifest list), but the manifests are clearly different. Which I guess makes sense but is a bit surprising.

jbtrystram: This is https://github.com/containers/skopeo/issues/1554 I think

PR #1138, opened 2025-04-29
jobs/nodeimage: use the base image digest for FROM

ravanelli:
Instead of relying on the tag of the base image, first grab the manifest digest than pass that value to the builders.
This will avoid reusing stale images that may be present in the local containers storage when building.

Continuation of: https://github.com/coreos/fedora-coreos-pipeline/pull/1137


jlebon (approved):
Just did some minor tweaks.

LGTM, thanks all!

PR #1141, opened 2025-05-05
Fix the emojis for the build-node-image job

cverna:
Use the same emojis as for the build job and use the Slack syntax for them. Also use the currentBuild.description which has the emojis in the message.


jbtrystram (approved):
LGTM

PR #1142, opened 2025-05-05
config.yaml: allow stream-level source config URL overrides

jlebon:
For https://github.com/openshift/os/issues/1775, we want the new el-only
streams to build from coreos/rhel-coreos-config. But for the older OCP
releases, we want to continue building from openshift/os to minimize
disruption.

Make the source config URL configurable at the stream level to allow
this. This matches the source config ref.


dustymabe (approved):
LGTM

PR #1164, opened 2025-06-13
[staging] config: add azure testing gallery and remove arch limitation

marmijo:
Sync with the main branch by adding the Azure testing gallery name to the config and removing the limitation to test only on x86_64, so we can also test aarch64 in staging.


dustymabe (approved):
LGTM

PR #1166, opened 2025-06-16
 feat(build-node-image): add option to skip brew upload stage 

Roshan-R:
Adds an option to skip the Brew upload stage when building the Node image. Useful for test builds where publishing to the Brew repository is not needed.

### DIFF
--- a/docs/config.yaml
+++ b/docs/config.yaml
@@ -317,6 +317,8 @@ ocp_node_builds:
                ref: master
                # REQUIRED: repo file name to build
                file: rhel-9.6.repo
+           # OPTIONAL: skip upload to brew. Defaults to false
+           skip_brew_upload: false
     # REQUIRED: knobs related to the source config
     registries:
         # REQUIRED: staging repo to push the final image built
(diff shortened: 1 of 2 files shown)


docs/config.yaml
```diff
@@ -317,6 +317,8 @@ ocp_node_builds:
                ref: master
                # REQUIRED: repo file name to build
                file: rhel-9.6.repo
+           # OPTIONAL: skip upload to brew
```
jbtrystram, line 320: we should write what's the default value

jobs/build-node-image.Jenkinsfile
```diff
@@ -147,10 +147,12 @@ lock(resource: "build-node-image") {
                                                                   "--add-openshift-build-labels"] + label_args)
             }
         }
-        stage("Brew Upload") {
-            // Use the staging since we already have the disgests
-            pipeutils.brew_upload(arches, params.RELEASE, registry_staging_repo, node_image_manifest_digest,
-                                  extensions_image_manifest_digest, timestamp, pipecfg)
+        if (!stream_info.skip_brew_upload){
```
jbtrystram, line 150: as the field is optional in the pipecfg you should make sure it's initialized first (and assign the default value if needed)

jbtrystram (changes requested):

jobs/build-node-image.Jenkinsfile
```diff
@@ -147,10 +147,12 @@ lock(resource: "build-node-image") {
...
+            stage("Brew Upload") {
+                // Use the staging since we already have the disgests
```
c4rt0, line 152: // Use the staging since we already have the digests

c4rt0 (changes requested):

PR #1168, opened 2025-06-16
Add pipeline-graph-view plugin

aaradhak:
Apparently Blue Ocean Jenkins view is deprecated. There is a new "Pipeline Graph View" plugin that is recommended instead. The pipeline-graph-view version `419.v946ff4a_49e1c `is compatible with our current jenkins version.

Ref: https://github.com/jenkinsci/dark-theme-plugin/issues/89#issuecomment-2939077690


dustymabe:
Thanks @aaradhak

PR #1171, opened 2025-06-23
config: limit proxmoxve to `testing/stable` for now

dustymabe:
We're still working out a little bit what the best format is for the uploaded disk images in https://github.com/coreos/fedora-coreos-tracker/issues/1652


jdoss (approved):
LGTM!

PR #1173, opened 2025-06-25
config: disable ppc64le architecture for now

dustymabe:
We are moving to a new datacenter and don't have a ppc64le builder set up there yet. This will aid in the migration.


marmijo (approved):
LGTM

PR #1174, opened 2025-06-25
utils: add new FCOS jenkins to protected jenkinses

marmijo:
We've moved the FCOS pipeline to a new cluster, so we need to update the
list of protected jenkinses to mark this cluster as the official one.

We also have access to the ppc64le builder from the new cluster, so we can
add ppc64le back into the list of additional arches.


dustymabe (approved):
LGTM

PR #1183, opened 2025-07-17
kola-azure: add azure_managed_identity to jenkins

prestist:
With the new argument which was added to cosa, lets use it while running on Jenkins.

ref: https://github.com/coreos/coreos-assembler/pull/4223

I think this is all that is needed; We need to update the yaml to include the var we are defining, but should be good? @marmijo

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -212,6 +212,19 @@ oc annotate secret/azure-kola-tests-config \
     jenkins.io/credentials-description="Azure kola tests credentials config"
 ```
 
+If you want to establish an authorized identity to pre-emptively grant access
+to other Azure resources, enabling the startup process to leverage Azure's default
+credentials, use the azure-kola-managed-identity:
+
+```
+oc create secret generic azure-kola-managed-identity \
+    --from-file=text="${MANAGED_IDENTITY_ID}"
+oc label secret/azure-kola-managed-identity \
+    jenkins.io/credentials-type=secretText
+oc annotate secret/azure-kola-managed-identity \
+    jenkins.io/credentials-description="Azure managed identity credentials"
+```
+
 NOTE: For the prod pipeline these secrets can be found in BitWarden
 
 ### [OPTIONAL] Creating OpenStack credentials configs


HACKING.md
```diff
@@ -210,6 +210,13 @@ oc label secret/azure-kola-tests-config \
     jenkins.io/credentials-type=secretFile
 oc annotate secret/azure-kola-tests-config \
     jenkins.io/credentials-description="Azure kola tests credentials config"
+
+oc create secret generic azure-managed-identity \
```
marmijo, line 214: Would you mind adding a comment above explaining the purpose of this secret? Or perhaps we should separate it into it's own `[OPTIONAL]` section. WDYT?

prestist: Sure, I will make another optional config section, I think that fits best.

prestist: Hmm, I added a comment to it; idk does nto quite fit the same form that was pre-existing wdyt? Maybe remove the ###Optional?

HACKING.md
```diff
@@ -212,6 +212,20 @@ oc annotate secret/azure-kola-tests-config \
...
 ```
 
+#### [OPTIONAL] Using shared Azure managed identity for kola tests
+If you want to establish an authorized identity to pre-emptively grant access
+to other Azure resources, enabling the startup process to leverage Azure's default
+credentials, use the azure-kola-managed-identity:
+
+```
+oc create secret generic azure-kola-managed-identity \
+    --from-file=text=/run/user/$UID/azureManagedIdentity.txt
```
marmijo, line 222: I should have looked more closely at other secrets when we were defining this one earlier. It looks like we could define it a little easier:
```
MANAGED_IDENTITY_ID=<azure managed identity resource id>
oc create secret generic azure-kola-managed-identity  --from-literal=text="${MANAGED_IDENTITY_ID}"
```

This also defines the format of the secret, which is consistent the rest of this doc.

jobs/kola-azure.Jenkinsfile
```diff
@@ -149,6 +151,7 @@ cosaPod(memory: "${cosa_memory_request_mb}Mi", kvm: false,
                      platformArgs: """-p=azure                               \
                          --azure-credentials \${AZURE_KOLA_TESTS_CONFIG}     \
                          --azure-location $region                            \
+                         --azure-managed-identity \${AZURE_MANAGED_IDENTITY}   \
```
marmijo, line 154:
```suggestion
                         --azure-managed-identity \${AZURE_KOLA_MANAGED_IDENTITY}   \
```

marmijo (commented):
I added a few more comments. 
I also tested this in  [kola-azure run 58](https://jenkins-fedora-coreos-pipeline.apps.ocp.fedoraproject.org/blue/organizations/jenkins/kola-azure/detail/kola-azure/58/pipeline/) and confirmed the identity is assigned to all VMs! :tada:

jobs/kola-azure.Jenkinsfile
prestist: Nice catch

marmijo (approved):
LGTM, thanks for working on this!

PR #1185, opened 2025-07-31
jobs/build: support importing OCI image using `cosa import` and skipping build stages

Roshan-R:
This allows the build pipeline to skip the fetch and the build ostree stages if source_oci_image is defined in the stream config. In that case, the image is imported using cosa import.

latest build logs: 
[jenkins](https://jenkins-fedora-coreos-pipeline.apps.ocp.stg.fedoraproject.org/job/rosh-build/15/)
[blue ocean](https://jenkins-fedora-coreos-pipeline.apps.ocp.stg.fedoraproject.org/blue/organizations/jenkins/rosh-build/detail/rosh-build/15/pipeline)

The failed tests are tracked here:
[#1996](https://github.com/coreos/fedora-coreos-tracker/issues/1996) 
[#1995](https://github.com/coreos/fedora-coreos-tracker/issues/1995)

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -17,6 +17,10 @@ properties([
       choice(name: 'STREAM',
              choices: pipeutils.get_streams_choices(pipecfg),
              description: 'CoreOS stream to build'),
+      string(name: 'SOURCE_OCI_IMAGE',
+             description: 'Override source_oci_image image to use. If set, the STREAM parameter will be ignored.',
+             defaultValue: "",
+             trim: true),
       string(name: 'VERSION',
              description: 'Override default versioning mechanism',
              defaultValue: '',
@@ -109,6 +113,9 @@ def cosa_memory_request_mb = 10.5 * 1024 as Integer
 // XXX: https://github.com/coreos/coreos-assembler/issues/3118 will make this
 // cleaner
 def ncpus = ((cosa_memory_request_mb - 512) / 1536) as Integer
+
+def source_oci_image = params.SOURCE_OCI_IMAGE ?: stream_info.get("source_oci_image", "")
+boolean import_oci = source_oci_image != ""
 
 echo "Waiting for build-${params.STREAM} lock"
 currentBuild.description = "${build_description} Waiting"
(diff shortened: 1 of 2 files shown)


config.yaml
```diff
@@ -7,6 +7,8 @@ streams:
     type: production
   testing-devel:
     type: development
+    # temporarily using sha256 tag as konflux builds do not have the latest tag
+    source_oci_image: docker://quay.io/konflux-fedora/coreos-tenant/fedora-coreos-rawhide@sha256:a0268014587489cbf36564c9a70948fb16bbaede01dc68cbf6bddc17e41e3f6b
```
jcapiitao, line 11: Could we add the ability to pull the latest OCI image from `quay.io/konflux-fedora/coreos-tenant/fedora-coreos-rawhide` easily ? 
I'm going to check if we can configure Konflux to tag with `latest` in build-stage. Other solution would be to add a minimal release pipeline doing so.

jbtrystram, line 11: you need to reflect that in the `docs/config.yaml`

jobs/build.Jenkinsfile
```diff
@@ -247,84 +258,93 @@ lock(resource: "build-${params.STREAM}") {
...
+                // And we're done!
+                return
+            }
+        newBuildID = buildID
+        } else{
+            stage("Import OCI image"){
+                shwrap("""
+                    cosa import docker://${source_oci_image}
+                """)
+                newBuildID = shwrapCapture("basename \$(realpath builds/latest)")
```
jbtrystram, line 344: let's pull the buildID from the builds.json: 

https://github.com/coreos/fedora-coreos-pipeline/blob/8b5f5f4bc563b5f6ae0e44d39d79eb8184f5d358/jobs/build.Jenkinsfile#L306-L307

jobs/build.Jenkinsfile
```diff
@@ -106,6 +106,17 @@ def cosa_memory_request_mb = 10.5 * 1024 as Integer
...
+def source_oci_image = null
+boolean skip_build = false
+
+if (stream_info.source_oci_image) {
+    skip_build = true
+    source_oci_image = stream_info.source_oci_image
+}
+
+echo "Skip Build: $skip_build"
+echo "Source OCI Image: $source_oci_image"
```
jbtrystram, line 118: This should be inside the if statement

jbtrystram (changes requested):

dustymabe:
> The "trigger vs pull" bit jumped out to me as well. I _think_ importing the Konflux builds that already happen on f-c-c pushes is fine, though it's a huge conceptual shift. It immediately makes the Jenkins pipeline feel "secondary" and the Konflux pipeline primary. E.g. the job name `build` doesn't really feel correct anymore. Again, not a bad thing (and my preference long-term as I mentioned is that we fully move to Konflux), but just calling this out.

I think it would be OK to pull from Konflux builds that have already happened, but I don't know if I really like it. I kind of like the point of control being the jenkins build for now.

That being said, even if we wanted to poll and import the already built thing, we have strong semanitcs around our version string. How do we control that if we "pull/poll" versus "trigger"? Wouldn't that be a blocker?

jbtrystram:
> But anyway, we won't be able to switch to it until we have https://gitlab.com/fedora/infrastructure/konflux/infra-deployments/-/issues/9, no?


Yes. Everything is prep work and we cannot do any meaningful switch before we have that. Hopefully it will happen soon ish.

> Shorter term, I'd like to switch rawhide to doing the container build itself (via https://github.com/coreos/coreos-assembler/pull/4242) so we at least start exercising that path. And then once we have multi-arch support, we can switch to Konflux (and get back signing).

I'd like to go straight to building the container with Konflux. What's the point of spending time changing things if we switch again to konflux a month later ?


Regarding the push vs pull discussion :

We can trigger Jenkins from konflux and have every konflux build trigger a Jenkins run with a specific image. BUT (with the current config) konflux will build each commit, so I don't want to choke the pipeline with 10 builds a day and block other jobs. I know there are lockfiles, but that would be quite an increase in churn. That's why I suggested we keep the daily mechanical build and pull the latest. We could very well make a check on the commit ref from fcc (since that's what the image will be tagged with to ensure consistency).

That said there is a lot of room for adjustment here and that's really up for discussion. 
So far our (@jcapiitao , @Roshan-R and me) goal was to set things up with the least possible disruption, to get things going and build up a "parallel world" to exercise things, while keeping the config straightforward.

But we could very well make some different choices. One idea that comes to mind is to have konflux trigger Jenkins on specific commits path (e.g. lockfiles) or git tags to avoid unnecessary churn. 

@dustymabe konflux is very much "triggered" ! It's simply that the trigger is git events (commits, pr, tags, probably other I don't know of). There are a lot of knobs we can tweak, so let's discuss.

I'm sure @jcapiitao will have other ideas to bring up here as well.

jbtrystram:
Regarding the version string yeah, good point, thanks for bringing that up, it's an oversight from me.

I'll have look to learn more and try to think of something that make sense.

jlebon:
> I'd like to go straight to building the container with Konflux. What's the point of spending time changing things if we switch again to konflux a month later ?

I'd like that as well, but it's not clear when this will happen. AIUI, one month is a lower bound estimate. We don't need to block on this before cutting over rawhide to the container build path and continuing to root out issues.

A primary goal of https://github.com/coreos/coreos-assembler/pull/4242 is to require minimal pipeline changes (ideally just `env: {COSA_BUILD_WITH_BUILDAH: 1}` in the pipecfg).

jlebon:
For the version string, we'll have to run versionary in an earlier step and somehow be able to feed that as a build argument to the build.

The X.Y.Z components are trivial and only need the git repo. The N component is trickier. It currently relies on build history in `builds.json`, but that won't get updated by Konflux but Jenkins. I think things could be made to work with the current scheme, but I think it'd be conceptually cleaner if N in that new flow changed to use Konflux build history instead?

dustymabe:
> For the version string, we'll have to run versionary in an earlier step and somehow be able to feed that as a build argument to the build.

but that implies triggering konflux (from somewhere, jenkins?) versus letting konflux drive itself. I'm happy with that, but was just trying to tease out the answer to poll versus trigger in my questions above.

> Alternatively, I'd be comfortable temporarily dropping s390x and ppc64le in rawhide though interested what others think.

Yeah. I mean since we control our pipeline we should just be able to add conditional logic to pull some builds from konflux and do some builds ourselves using `COSA_BUILD_WITH_BUILDAH`.

I wouldn't want to do this for production, but for rawhide it seems reasonable to go hybrid.

jcapiitao:
> I'm more in favor of triggering Jenkins pipeline with Konflux

Problem with that approach is that I doubt we'll have time to implement it before the F43 cutoff.
But, as said above, not having s390x and ppc64le is in the critical path. If building x86 only is ok, then let's go with the pull option. The push option would be handled later on.

> Alternatively, I'd be comfortable temporarily dropping s390x and ppc64le in rawhide though interested what others think.
I don't have strong opinion on that TBH. IIUC we would trade arch builds to signed OCI image ?

jbtrystram:
> Problem with that approach is that I doubt we'll have time to implement it before the F43 cutoff. 

Does it matter if we don't plan on releasing anything anyway ?

config.yaml
```diff
@@ -122,3 +125,6 @@ misc:
   generate_release_index: true
   run_extended_upgrade_test_fcos: true
   allow_missing_architectures_when_autotriggering_release_job: true
+
+hotfix:
+  name: rchandar-import-konflux-build
```
jbtrystram, line 130: Move to a separate DNM commit

config.yaml
```diff
@@ -31,7 +32,9 @@ source_config:
   url: https://github.com/coreos/fedora-coreos-config
 
 s3:
-  bucket: fcos-builds
+  # TODO: change the bucket url once finished
+  # bucket: fcos-builds
+  bucket: prod-account-match-fcos-builds
```
jbtrystram, line 37: Have that to a separate DNM commit

jobs/build.Jenkinsfile
```diff
@@ -251,84 +259,94 @@ lock(resource: "build-${params.STREAM}") {
...
-            // And we're done!
-            return
+                // And we're done!
+                return
+            }
+        newBuildID = buildID
+        } else{
+            stage("Import OCI image"){
+                echo "Skiping Build"
+                echo "Source OCI Image: $source_oci_image"
```
jbtrystram, line 343:
```suggestion
        } else {
            stage("Import OCI image") {
                echo ""
                echo "Skiping build : Importing OCI : $source_oci_image"
```

jobs/build.Jenkinsfile
```diff
@@ -110,6 +115,9 @@ def cosa_memory_request_mb = 10.5 * 1024 as Integer
 // cleaner
 def ncpus = ((cosa_memory_request_mb - 512) / 1536) as Integer
 
+def source_oci_image = params.SOURCE_OCI_IMAGE ?: stream_info.get("source_oci_image", "")
+boolean skip_build = source_oci_image != ""
```
jbtrystram, line 119: let's name that with something more explicit, like "import_oci"

jbtrystram (changes requested):
Some small comments

jobs/build.Jenkinsfile
```diff
@@ -101,6 +105,8 @@ def strict_build_param = stream_info.type == "mechanical" ? "" : "--strict"
 // XXX bump an extra 2G (to 10.5) because of an error we are seeing in
 // testiso: https://github.com/coreos/fedora-coreos-tracker/issues/1339
 def cosa_memory_request_mb = 10.5 * 1024 as Integer
+// TODO: remove if not needed
+// def cosa_memory_request_mb = 16 * 1024 as Integer
```
jbtrystram, line 109: we can remove this

jbtrystram (approved):
LGTM

jobs/build.Jenkinsfile
```diff
@@ -17,6 +17,10 @@ properties([
       choice(name: 'STREAM',
              choices: pipeutils.get_streams_choices(pipecfg),
              description: 'CoreOS stream to build'),
+      string(name: 'SOURCE_OCI_IMAGE',
+             description: 'Override source_oci_image image to use. If set, the STREAM parameter will be ignored.',
```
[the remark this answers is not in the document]
jbtrystram: @jlebon I started working on that but wasn't really satisfied about how the stream tag is very fcos specific. 
I wanted to chat with you about it, but the opportunity did not come up yet.

See https://github.com/jbtrystram/fedora-coreos-pipeline/commit/828e545d86da60955161f4345e966e63f90b833b for more details

PR #1188, opened 2025-08-04
jobs/build-node-image: update s3 stream directory

marmijo:
Append "rhel-" to the `rhel_stream` variable to get the correct s3 stream directory for the `cosa buildfetch` command.


dustymabe (approved):
LGTM

PR #1194, opened 2025-08-18
config.yaml: turn on COSA_BUILD_WITH_BUILDAH for rawhide

jlebon:
As per https://github.com/coreos/fedora-coreos-tracker/issues/1969, we are switching to the container-native path for building FCOS in Fedora 43.

Enable this using the ready-made `COSA_BUILD_WITH_BUILDAH` env var.

Fedora 43 has already brached so once this bakes a bit in Rawhide, we should do the same to the `branched` stream as well.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -12,6 +12,9 @@ streams:
     # type: development # do not touch; line managed by `next-devel/manage.py`
   rawhide:
     type: mechanical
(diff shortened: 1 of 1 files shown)


jlebon:
Doing a final test in stg before merging this one.

Issue #1186, opened 2025-08-01
release: optionnaly skip pushing incomplete OCI manifests


jbtrystram:
We historically support publishing incomplete releases (not including all architectures).

Recently we started to publish OCI images to the ART repos to allow the `build-node-image` job to derive the OCP container image.  A corner case here is that publishing an incomplete OCI manifest triggers the build-node-image which will fail because the manifest is incomplete, 
until the next release job successfully push a complete manifest.

This is problematic because ART need to be able to reliably rebuild the OCP images if the openshift bits change, so they need a reliabe base manifest. 

I suggest to skip publishing the manifest if it's incomplete. Let's add a pipecfg knob e.g. `push-partial-manifest: false` under `registry_repos`

dustymabe:
I see two options here:

1. Modify the build-node-image job to also just build for the architectures that have been built
2. Modify the calls to the release job to stop setting `ALLOW_MISSING_ARCHES` to true

prestist:
Which one would be preferred? @dustymabe 

I was thinking we could add the arg mentioned above per registry (defined in the config.yaml) defaulting to true except on whichever one we don't want partial manifests to, and pivoting in the push step for release.jenkins; or is that not suitable?

dustymabe:
basically the tradeoff between `1.` and `2.` is that with `2.` if one architecture is failing for an extended period of time OpenShift gets no updates to be testing in the meantime for all the other successful architectures.

jlebon:
AIUI, ART will just not use builds which are incomplete. So we might as well do option 2.

New e.g. openshift-kubelet RPM builds at least would not be blocked from going into the nightly/CI payloads.

dustymabe:
In that case I think we just need to change the logic slightly and also the comment in

https://github.com/coreos/fedora-coreos-pipeline/blob/f9e562ee3c64b032311a873caaede49e9d52ecdb/jobs/build.Jenkinsfile#L571-L586

I think we do still want `ALLOW_MISSING_ARCHES=true` for FCOS development/mechanical streams, but for RHCOS we can just always set it to `false`.

jbtrystram:
> basically the tradeoff between `1.` and `2.` is that with `2.` if one architecture is failing for an extended period of time OpenShift gets no updates to be testing in the meantime for all the other successful architectures.

Yeah, I think ART have some tooling that blocks (or at least mark as errors) payloads that have version inconsistencies between arches .

[...]

PR #1198, opened 2025-08-20
build: conditionally allow missing arches

prestist:
Publishing incomplete OCI manifests results in failures for OCP images. Update build to identify and pivot based on target.

#1186

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -131,6 +131,10 @@ def add_hotfix_parameters_if_supported() {
     return []
 }
 
+def is_rhcos_jenkins() {
+    return env.JENKINS_URL && env.JENKINS_URL.contains('rhcos')
+}
+
 def get_source_config_for_stream(pipecfg, stream) {
     def url = pipecfg.source_config.url
     if (pipecfg.streams[stream].source_config_url) {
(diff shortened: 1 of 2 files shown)


utils.groovy
```diff
@@ -131,6 +131,10 @@ def add_hotfix_parameters_if_supported() {
     return []
 }
 
+def is_rhcos_jenkins() {
+    return env.JENKINS_URL && env.JENKINS_URL.contains('rhcos')
```
dustymabe, line 135: I did something similar in https://github.com/coreos/fedora-coreos-pipeline/pull/1197 but it for detecting when we were running against FCOS.


```
if (pipecfg.source_config.url != 'https://github.com/coreos/fedora-coreos-config')
    // Only skip builds in FCOS for now                                           
    return false                                                                  
}                                                                                 

```


@jlebon @marmijo @aaradhak is there a method you prefer here?

jlebon: Hmm, can we shorten it to `allow_missing_arches_default`?

dustymabe: > Hmm, can we shorten it to `allow_missing_arches_default`?

yeah, but it's slightly more complicated than that implies. Right now where we are using it is when triggering a later job rather than actually setting the default for that parameter in that job.

so I do like the shorter name, but it doesn't tell the whole story.

jlebon: Yeah, naming is tricky. I'm OK with something wordier, but at some point it's better to capture the nuance in a docstring rather than trying to capture it all in the variable name.

prestist: Okay I will go-ahead and do that thank you!

marmijo (approved):
LGTM

PR #1204, opened 2025-08-27
libcloud: trim .gz from aws file name in winli closure

marmijo:
The AWS Windows License Included (winli) closure grabs the aws image path from the meta.json file, but the path could include the `.gz` file extension from compression. Trim the extension so the ore wrapper can accept the file name, even though no operations are performed on it.

### DIFF
--- a/libcloud.groovy
+++ b/libcloud.groovy
@@ -101,10 +101,11 @@ def replicate_to_clouds(pipecfg, basearch, buildID, stream) {
         if (pipecfg.clouds?.aws &&
             utils.credentialsExist(credentials)) {
 
-            // grab the aws vmdk image name from the metadata to pass to the winli closure.
-            // the cosa ore wrapper still requires the image to exist, but we dont upload 
-            // anything so we'll just touch the file in the cosa working dir.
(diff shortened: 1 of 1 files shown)


dustymabe:
I think it would really be preferable if we could change the underlying code so this hack isn't needed. Do we know how hard that would be?

dustymabe:
Let me restate. I remember discussing this before, but forget the exact reason other than something along the lines of "it makes the code easier". 

Since the file isn't being used here and the source snapshot is, why do we need to pass `--file` to `ore`. If we didn't do that I don't think we would have hit this bug.

I think we should change [cmd-ore-wrapper](https://github.com/coreos/coreos-assembler/blob/af13093f6de4748a722019f8020969aeb3f5e292/src/cmd-ore-wrapper#L124-L125) to not do that check if 
`--winli` is passed and then change [aws.py](https://github.com/coreos/coreos-assembler/blob/af13093f6de4748a722019f8020969aeb3f5e292/src/cosalib/aws.py#L174) to not pass `--file` (and also `upload.go` to not require it if source snapshot is passed in).

PR #1205, opened 2025-08-28
jobs/build-arch: add --autolock to the `cosa build` call

dustymabe:
In the build-with-buildah workflow `cosa fetch` doesn't do anything so we need to add the `--autolock=VERSION` argument to the `cosa build` call.

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -247,7 +247,7 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
             def version = "--version ${params.VERSION}"
             def force = params.FORCE ? "--force" : ""
             shwrap("""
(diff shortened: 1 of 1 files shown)


jlebon (approved):
Ahh yeah, makes sense.

jlebon:
(We later realized this was not actually necessary because `cmd-build-with-buildah`, like `cmd-build`, automatically uses autolock if it detects that it's being used to "complete" a pre-existing build which we already buildfetched (which is the case here).

aaradhak:
> (We later realized this was not actually necessary because `cmd-build-with-buildah`, like `cmd-build`, automatically uses autolock if it detects that it's being used to "complete" a pre-existing build which we already buildfetched (which is the case here).
Do we need to revert this change?

dustymabe:
Yes. will do

PR #1206, opened 2025-08-29
config.yaml: switch branched to buildah path

jlebon:
We have stabilized rawhide enough now to also do this on branched.


dustymabe:
fixed spacing for the other keys under `branched`.

dustymabe (approved):
LGTM

PR #1216, opened 2025-09-10
jobs/release: simplify pushed image signing

jlebon:
Now that the manifest list is part of `meta.json`, we no longer need to use `--write-digest-to-file` and `--manifest-list-digest`. This fixes a subtle issue: if `cosa push-container-manifest` no-ops (because the images were already pushed), no digestfile is written, and so signing would fail.

By having it be in `meta.json`, we make it robust against re-running the release job, which is designed to be idempotent. (Though worth noting that the signing operation itself is not yet idempotent; we always re-sign, overwriting any old signature.)

But since signing can commonly fail, let's move it to after we upload the updated `meta.json` to S3 so that `cosa push-container-manifest` can at least keep no-op'ing when we rerun.

### DIFF
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -289,27 +289,12 @@ lock(resource: "release-${params.STREAM}", extra: locks) {
                                 --auth=tmp/push-secret-${metajsonname} \
                                 --repo=${repo.repo} ${tag_args.join(' ')} \
                                 --artifact=${artifact} --metajsonname=${metajsonname} \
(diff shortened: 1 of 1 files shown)


dustymabe (approved):
LGTM

jlebon:
Hmm, this was meant to be merged with https://github.com/coreos/coreos-assembler/pull/4306. But actually... I don't think it actually breaks FCOS to not have the cosa patch. We just won't sign the manifest list digest.

PR #1224, opened 2025-09-15
jobs/release: take stream lock first

dustymabe:
It appears that jenkins will wait until all the locks are available before taking any locks, which makes sense, but causes the desired outcome of 49e05be to not be achieved. Let's take the toplevel stream based lock first by itself so the kola-upgrade job will actually wait for the release job to finish before kicking off.

### DIFF
--- a/jobs/release.Jenkinsfile
+++ b/jobs/release.Jenkinsfile
@@ -85,12 +85,14 @@ def stream_info = pipecfg.streams[params.STREAM]
 build_description += "[${basearches.join(' ')}][${params.VERSION}]"
 currentBuild.description = "${build_description} Waiting"
 
-// We just lock here out of an abundance of caution in case somehow two release
-// jobs run for the same stream, but that really shouldn't happen. Anyway, if it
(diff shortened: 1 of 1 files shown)


jobs/release.Jenkinsfile
```diff
@@ -85,12 +85,14 @@ def stream_info = pipecfg.streams[params.STREAM]
 build_description += "[${basearches.join(' ')}][${params.VERSION}]"
 currentBuild.description = "${build_description} Waiting"
 
-// We just lock here out of an abundance of caution in case somehow two release
-// jobs run for the same stream, but that really shouldn't happen. Anyway, if it
-// *does*, this makes sure they're run serially.
+// Take a lock for the stream to make sure only one of these run at a
+// time and also synchronize with the kola-upgrade job. Take this lock
+// early (i.e. don't wait for all the arch specific locks) and hold it
+// so we can synchronize wit hthe kola-upgrade job.
```
jlebon, line 91: with the

dustymabe: fixed

PR #1230, opened 2025-09-17

marmijo:
Fedora 43 is now being tracked in the `next-devel` branch, so we can disable the `branched` stream now.


dustymabe (approved):
LGTM

PR #1231, opened 2025-09-18
Move build-mechnical to 04:00 UTC

sdodson:
The goal being to ensure build-mechnical and the downstream build-node-image jobs finish before we typically see content changed, which tends to be EMEA to US East AM hours. In situations where we're hoping to pick up late changes this may cause delays, if that becomes an issue I'd propose we consider adding a second run at 17:00 UTC (EDT 12:00 + 5hr), but ideally we'd limit that to only versions where we hope to have rapid turn around. Those versions would be 9.6 today, then 9.6, 9.8, and 10.2 starting approximately May 2026.

### DIFF
--- a/jobs/build-mechanical.Jenkinsfile
+++ b/jobs/build-mechanical.Jenkinsfile
@@ -7,8 +7,8 @@ node {
 
 properties([
     pipelineTriggers([
-        // run every 24h at 10:00 UTC
-        cron("0 10 * * *")
+        // run every 24h at 04:00 UTC
+        cron("0 4 * * *")
     ]),
     buildDiscarder(logRotator(
         numToKeepStr: '100',


travier (approved):

dustymabe:
This time was originally chosen (1da5e54) based on when Fedora composes would finish for `rawhide`.

dustymabe:
> ensure build-mechnical and the downstream build-node-image jobs finish before we typically see content changed

why? wouldn't you want to wait until after new content was available?

jlebon:
> All of the last 10 rhel-9.6 jobs showing up in Jenkins were triggered by timer or me. I can check to see why we're not getting builds triggered on change by ART for at least rhel-9.6.

To be clear I wasn't implying that ART is doing this today. But that it could.

jlebon:
> > > All of the last 10 rhel-9.6 jobs showing up in Jenkins were triggered by timer or me. I can check to see why we're not getting builds triggered on change by ART for at least rhel-9.6.
> > 
> > 
> > To be clear I wasn't implying that ART is doing this today. But that it could.
> 
> I'd actually prefer ART didn't. Since RHEL content shouldn't really change more than once a day, is it really needed?

Don't we today already trigger once a day? ISTM the ask here is to have better timing wrt RHEL, not necessarily have more builds. (This would actually be a reduction in builds if RHEL content changes less often than that... though I'm not sure in practice how often that happens.) But I guess if we go this way, we should probably also ask ART to cap the rate as well in case it sometimes churns faster for some reason.

sdodson:
Just summarizing.

Right now build-mechanical is only triggered by timer and it initiates builds for c10s, c9s, rhel-10.1 and rhel-9.6, in that order. It's never initiated by ART. Builds of c10s, c9s, rhel-10.1 and rhel-9.6 are pretty much exclusively triggered by build-mechanical or a human replaying a failure. I guess also Fedora pipelines happen too but I don't know about those.

build-mechanical has a median run time of six hours and a maximum runtime of 13 hours. We should expect that runtime to grow whenever we add 9.8 and 10.2, though we'll remove 10.1 at roughly the same time. There are almost always changes to c10s and c9s Tuesday through Saturday, so this means that rhel-10.1 or rhel-9.6 typically start 13:00-14:00 UTC.

build-node-image IS triggered by ART whenever OCP content changes, potentially multiple times a day.

I'm not sure how we can please everyone with just one knob to set the start time across two pipelines and three OSes publishing schedules. Do we know if moving four hours earlier would be too early for Fedora? It seems like [CS9 composes](https://composes.stream.centos.org/stream-9/production/) are complete by 04:00, [CS10 composes](https://composes.stream.centos.org/stream-10/production/) by 04:30, [Rawhide](https://kojipkgs.fedoraproject.org/compose/rawhide/) by 5:30. RHEL is a bit of a wildcard unfortunately, but lets assume done 16:00 UTC is generally done.

With just one knob and no further changes I'd probably propose moving things two hours **later** in hopes of RHEL pushes having completed by the time we start RHEL builds but still with enough time to complete the process before the daily cut-off. Any appetite for that change? An alternative would be to try to go two hours earlier but I'd like to try later first.

dustymabe:
> [Rawhide](https://kojipkgs.fedoraproject.org/compose/rawhide/) by 5:30.

Typically the time when the compose completes isn't the time that directory was created. For [today's](https://kojipkgs.fedoraproject.org/compose/rawhide/Fedora-Rawhide-20251001.n.0/) rawhide build I'd use the timestamp on [this](https://kojipkgs.fedoraproject.org/compose/rawhide/Fedora-Rawhide-20251001.n.0/STATUS) file which says:

```
STATUS                  2025-10-01 07:01   20
```

> Do we know if moving four hours earlier would be too early for Fedora?

Another option here (I think) would be to change the order the `c10s`, `c9s`, `rhel-10.1` and `rhel-9.6` builds are triggered. If the `rhel-*` builds were to go first would that help achieve the goal here? I assume the punctuality here is more focused on the `rhel-*` builds and not so much the CentOS ones?

> With just one knob and no further changes I'd probably propose moving things two hours later in hopes of RHEL pushes having completed by the time we start RHEL builds but still with enough time to complete the process before the daily cut-off.

Yeah. Later by two hours seems fine. The only problem I see here is that it leans more heavily on more western timezone folks to investgate failures because it will be later in the day before failures trickle in (more towards the end of eastern TZ folks' day).

If you don't mind, can we clear up what the actual goal is? In your original description you mention:

> The goal being to ensure build-mechnical and the downstream build-node-image jobs finish before we typically see content changed, which tends to be EMEA to US East AM hours.

but I don't really understand why we wouldn't want to pick up the newest content sooner after it's available rather than later. I would think the goal would be "to have the newest content available by XYZ cutoff time on W day of the week". Can you put it in those terms?

FTR I'm not opposed to having multiple knobs here. It just makes things slightly more complicated.

PR #1234, opened 2025-09-24
jobs/bump-lockfile: fix setting env for remote builders

dustymabe:
The setting of the COSA_BUILD_WITH_BUILDAH env variable gets lost here when proxying it across to a remote builder. Which means that only x86_64 is getting lockfile bumps [1] (i.e. the only arch that isn't using a remote builder). Let's wrap it in a `cosa shell` to fix this.

Fixes 66b653b

[1] https://github.com/coreos/fedora-coreos-config/commit/a94a13aebc341a9e17f407b0766a6bc511202230


yasminvalim (approved):
/lgtm

PR #1235, opened 2025-09-24
jobs/kola-upgrade: bump timeout to 150m

dustymabe:
The test itself [1] was just modified to have a longer timeout so let's adjust here to twice that time since there's an initial run and also a rerun that could happen.

[1] https://github.com/coreos/fedora-coreos-config/commit/7f21e359fd27a26edff30894f72857e22a577f2b


yasminvalim (approved):
/LGTM

PR #1236, opened 2025-09-25
skip build by importing konflux built OCI Image in build-arch

Roshan-R:
Should be similar to https://github.com/coreos/fedora-coreos-pipeline/pull/1185/

### DIFF
--- a/jobs/build.Jenkinsfile
+++ b/jobs/build.Jenkinsfile
@@ -301,14 +301,6 @@ lock(resource: "build-${params.STREAM}") {
                 shwrap("""
                 cosa build ostree ${strict_build_param} --skip-prune ${force} ${version_arg} ${parent_arg}
                 """)
-
-                // Insert the parent info into meta.json so we can display it in
-                // the release browser and for sanity checking
-                if (parent_version) {
-                    shwrap("""
-                    cosa meta --set fedora-coreos.parent-version=${parent_version}
-                    """)
-                }
             }
 
             def buildID = shwrapCapture("readlink builds/latest")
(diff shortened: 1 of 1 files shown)


Roshan-R:
latest successful build logs: https://jenkins-rhcos--devel-pipeline.apps.int.preprod-stable-spoke1-dc-iad2.itup.redhat.com/job/rosh-build-arch/24/

jobs/build-arch.Jenkinsfile
```diff
@@ -237,33 +244,40 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
...
+                    shwrap("""
+                    cosa meta --set fedora-coreos.parent-version=${parent_version}
+                    """)
+                }
+            }
+        } else {
+            stage("Import OCI image") {
+                echo "Skipping build : Importing OCI : $source_oci_image"
+                shwrap("cosa import docker://${source_oci_image} --skip-prune")
+                newBuildID = shwrapCapture("cosa meta --dump | jq '.buildid'")
```
joelcapitao: We don't need this line as we reuse the ID of the `x86_64` build

joelcapitao (changes requested):
I realized we don't chain our OCI with previous build i.e `cosa meta --set fedora-coreos.parent-version=${parent_version}` in `build.json` and here. @jbtrystram would it break the release browser ? I guess we use that `fedora-coreos.parent-version` metadata to list the `Upgraded:` NEVRAs but I'm unsure.

joelcapitao:
> I realized we don't chain our OCI with previous build i.e `cosa meta --set fedora-coreos.parent-version=${parent_version}` in `build.json` and here. @jbtrystram would it break the release browser ? I guess we use that `fedora-coreos.parent-version` metadata to list the `Upgraded:` NEVRAs but I'm unsure.

According to https://github.com/coreos/fedora-coreos-streams/blob/main/.github/ISSUE_TEMPLATE/stable.md#sanity-check-the-build we need the parent version. So let's add it as well in this PR for both `build-arch` and `build` jobs.

joelcapitao (approved):
LGTM

PR #1238, opened 2025-09-29
kola-upgrade: rework architecture detection logic

dustymabe:
Similar to c25521d we need to update the logic in the upgrade test
for finding if a given release is built for the given architecture.


tlbueno (approved):
lgtm

PR #1240, opened 2025-10-02
release: fix the lock for releases containing 1 arch

jbtrystram:
In 3051996 we added a second level of locking on a per-arch basis. However if the release job is started only for 1 architecture, using `1..-1` results in an index out of bounds as the inital array is one. Use a remove in place operation instead.


joelcapitao (approved):
LGTM

PR #1242, opened 2025-10-03
libcloud.groovy: drop outdated cleanup step

marmijo:
This code was supposed to be removed in an earlier change`[1]`. Removing it now fixes the cloud-replicate job, which is failing due to the undeclared variable.

`[1]`: https://github.com/coreos/fedora-coreos-pipeline/pull/1217


aaradhak (approved):
/lgtm

PR #1243, opened 2025-10-03
libcloud.groovy: support winli flag in AWS replicate closure

marmijo:
Extend `awsReplicateClosure` to accept a `winli` argument (default: false). When set, the closure will pass `--winli` to `cosa aws-replicate` so that only WinLI AMIs are replicated.

Add an additional replicator entry that invokes the closure with `winli = true`, ensuring WinLI replication is handled explicitly and kept separate from standard AMI replication.


dustymabe (approved):
LGTM

PR #1244, opened 2025-10-06
libcloud.groovy: ensure winli closure is only called for x86_64

marmijo:
Move the call statement for the aws winli replicate closure to the winli 'if statement' to ensure it's only executed on x86_64 and when the winli knob is set in the pipeline config.


aaradhak (approved):
LGTM

PR #1246, opened 2025-10-09
utils: add `non_konflux_driven_streams` to filter out Konflux-managed streams

Roshan-R:
Adds a new utility function `non_konflux_driver_streams(config, streams_subset)` that returns a list of stream names from the given subset which are *not* managed by Konflux (i.e., streams that do not have `konflux_driver: true` in the config).

This is useful for excluding Konflux-managed streams from Jenkins crontab jobs, allowing those to be managed separately by Konflux.

see: https://issues.redhat.com/browse/COS-3637

### DIFF
--- a/utils.groovy
+++ b/utils.groovy
@@ -408,7 +408,8 @@ def get_streams_choices(config, node = null) {
 
     def default_stream = stream_source.find { k, v -> v['default'] == true }?.key
     def other_streams = stream_source.keySet().minus(default_stream) as List
-    return [default_stream] + other_streams
+    def non_konflux_driven_streams = non_konflux_driven_streams(config, other_streams)
+    return [default_stream] + non_konflux_driven_streams
 }
 
 // Returns the default trigger for push notifications. This will trigger builds


jobs/build-mechanical.Jenkinsfile
```diff
@@ -19,10 +19,13 @@ properties([
 
 node {
     def mechanical_streams = pipeutils.streams_of_type(pipecfg, 'mechanical')
+
     def scheduled_streams = pipeutils.scheduled_streams(pipecfg, mechanical_streams)
     if (scheduled_streams) {
         mechanical_streams = scheduled_streams
     }
+    // Filter out streams that are managed by konflux
+    mechanical_streams = pipeutils.non_konflux_driven_streams(pipecfg, mechanical_streams)
```
jbtrystram: This could be called from `pipeutils.scheduled_streams()`

utils.groovy
```diff
@@ -397,12 +397,19 @@ def scheduled_streams(config, streams_subset) {
...
+}
+
 def get_streams_choices(config, node = null) {
     def stream_source = node ? config.ocp_node_builds.release : config.streams
 
     def default_stream = stream_source.find { k, v -> v['default'] == true }?.key
     def other_streams = stream_source.keySet().minus(default_stream) as List
-    return [default_stream] + other_streams
+    def non_konflux_driven_streams = non_konflux_driven_streams(config, other_streams)
+    return [default_stream] + non_konflux_driven_streams
```
jbtrystram: We should still be able to manually start a job for a given stream, at least until we have https://issues.redhat.com/browse/COS-3604 and can dynamically figure out what stream the konflux image is from

jbtrystram (changes requested):

utils.groovy
joelcapitao: I'm handling this already in the Konflux configuration c.f https://gitlab.com/fedora/infrastructure/konflux/tenants-config/-/merge_requests/73/diffs 
```
    - name: jenkins-job-params
      value: "NO_UPLOAD=true IMPORT_OCI_IMAGE=$CONTAINER_IMAGE_URL STREAM=$BRANCH"
```

IMO we should not be able to manually start a job from Jenkins anymore, once switched to Konflux.

jobs/build-mechanical.Jenkinsfile
Roshan-R: If we want to do that then i.g we would have to change the name for scheduled_streams as according to https://github.com/coreos/fedora-coreos-pipeline/blob/8bc217b087eecd12e8362673b254d02e47f8bb83/utils.groovy#L394-L398 it is only used to filter out streams which have `scheduled:true`

PR #1254, opened 2025-10-27
config.yaml: disable next-devel 2025-10-27

marmijo:
Now that testing-devel is tracking Fedora 43`[1]`* there is no difference between testing-devel and next-devel, so we can disable it.

`[1]`: https://github.com/coreos/fedora-coreos-config/pull/3852

`*` PR hasn't merged yet


dustymabe:
LGTM

PR #1255, opened 2025-11-01
jenkins/plugins: update and pin compatible versions

coreosbot-releng:


Rolv-Apneseth (approved):
LGTM

c4rt0 (approved):
I would probably rename the PR to something like `jenkins/plugins: update and pin compatible versions` since we are not exactly upgrading these to the latest.

Other than that
/lgtm

PR #1256, opened 2025-11-06
jobs/bump-lockfile: enable for mechanical streams

joelcapitao:
As Konflux is commit-based, enabling lockfiles for all the streams (not only development ones) will ensure us to have a new build when necessary i.e: when new tested packages set exists. 
This addresses part of the plan described in [1].

[1] https://github.com/coreos/fedora-coreos-tracker/issues/2038#issuecomment-3498258143


dustymabe (approved):
LGTM

PR #1258, opened 2025-11-11
build-node-image: label images with `io.openshift.os.streamclass`

marmijo:
Append a stream class label to both the node and extensions images
derived from the RHEL major version of the release (e.g. `rhel-9`,
`rhel-10`). The MCO can then match on that label to pick the correctly
tagged payload with either RHEL9 or RHEL10 content when operating in a
dual stream environment.

### DIFF
--- a/jobs/build-node-image.Jenkinsfile
+++ b/jobs/build-node-image.Jenkinsfile
@@ -53,6 +53,14 @@ def basearches = params.ARCHES.split() as Set
 def timeout_mins = 300
 
 def cosa_img = params.COREOS_ASSEMBLER_IMAGE
+// Derive the stream class once so both node and extensions builds use
+// the `io.openshift.os.streamclass` label. The streamclass is based on
+// the rhel major version and is used by the MCO to later select the right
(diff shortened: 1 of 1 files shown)


jobs/build-node-image.Jenkinsfile
```diff
@@ -122,8 +135,11 @@ lock(resource: "build-node-image") {
             withCredentials([file(credentialsId: 'oscontainer-push-registry-secret', variable: 'REGISTRY_AUTH_FILE')]) {
                  def build_from = params.FROM ?: stream_info.from
                  def label_args = []
+                 if (rhel_coreos_tag) {
+                     label_args += ["--label", "io.coreos.oscontainerimage.osstream=${rhel_coreos_tag}"]
```
travier:
```suggestion
                     label_args += ["--label", "com.coreos.oscontainerimage.osstream=${rhel_coreos_tag}"]
```

travier:
Should we set those directly when building it in the Containerfile?

travier:
Don't we already have a version as a label somewhere that could be used by the MCO to pick that info?

PR #1260, opened 2025-11-12
jobs/build: pass the OCI image to the multi-arch job

jbtrystram:
If the build is importing an OCI image, the forked multi-arch builds should too.
Pass on the source OCI parameter to the multiarch build.


marmijo (approved):
LGTM

PR #1285, opened 2025-12-22
builder-common.bu: Prune untagged rhel-coreos-base and node-staging images

ravanelli:
- Explicitly target untagged images with repository names that bypass the default dangling filter. Applied 12h retention policy for these specific images.

### DIFF
--- a/multi-arch-builders/builder-common.bu
+++ b/multi-arch-builders/builder-common.bu
@@ -99,6 +99,14 @@ storage:
           ExecStart=podman volume prune --force --filter="label!=persistent"
           ExecStart=podman image prune --force
           ExecStart=podman image prune --all --external --force --filter until=48h
+          # Added specific cleanup for rhel-coreos-base and node-staging
+          # untagged images older than 12h:
+          ExecStart=/usr/bin/bash -c "podman images \
+           --filter 'reference=registry.ci.openshift.org/coreos/*' \
+           --filter 'until=12h' \
+           --format '{{.ID}} {{.Tag}}' \
+           | awk '$2 == \"<none>\" {print $1}' \
+           | xargs -r podman rmi -f"
     - path: /home/builder/.config/systemd/user/prune-container-resources.timer
       mode: 0644
       user:


HuijingHei (approved):
LGTM, not sure if we should sync the time with above `48h`.

dustymabe:
> * Explicitly target untagged images with repository names that bypass the default dangling filter. Applied 12h retention policy for these specific images.

hmm. I'm not sure I understand why the `podman image prune --force` isn't pruning untagged images?

What is the "default dangling filter" and why isn't it sufficient here?

multi-arch-builders/builder-common.bu
```diff
@@ -99,6 +99,14 @@ storage:
           ExecStart=podman volume prune --force --filter="label!=persistent"
           ExecStart=podman image prune --force
           ExecStart=podman image prune --all --external --force --filter until=48h
+          # Added specific cleanup for rhel-coreos-base and node-staging
+          # untagged images older than 12h:
+          ExecStart=/usr/bin/bash -c "podman images \
+           --filter 'reference=registry.ci.openshift.org/coreos/*' \
+           --filter 'until=12h' \
+           --format '{{.ID}} {{.Tag}}' \
+           | awk '$2 == \"<none>\" {print $1}' \
```
c4rt0, line 108: Hi @ravanelli! I noticed Gemini [flagged an issue](https://github.com/coreos/fedora-coreos-pipeline/pull/1285#discussion_r2640151707) with the awk. I ran a quick test and it does seem like this is a bug:

```
adamsky@lenovo ~/Work/butane $ echo "banana <none>" | bash -c "awk '\$2 == \"<none>\" {print \$1}'"            
banana
 
adamsky@lenovo ~/Work/butane $ echo "banana <none>" | bash -c "awk '$2 == \"<none>\" {print $1}'"  
awk: cmd. line:1:  == "<none>" {print }
awk: cmd. line:1:  ^ syntax error
```

Rolv-Apneseth: Would `--format '{{if eq .Tag "<none>"}}{{.ID}}{{end}}'` work? If so we could just drop the `awk` entirely.

PR #1323, opened 2026-03-20
jobs/kola-azure: use gallery-profile and new versioning for images

marmijo:
Define the new coreos gallery profile config item in config.yaml and
pass it to the Azure gallery image creation.

Also pass the full FCOS version through so ore can derive an
Azure-compatible gallery image version based on the gallery profile.
Azure gallery image versions must be in x.y.z form, while FCOS versions
include an additional "stream" component. The full FCOS version is now
provided and truncated before creating the image version since we are
using the "fedora-community" coreos gallery profile.

Needs: https://github.com/coreos/coreos-assembler/pull/4500

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -97,6 +97,7 @@ clouds:
     test_storage_account: fedoracoreostesting
     test_storage_container: fedora-coreos-testing-image-blobs
     test_gallery: fedoracoreostestinggallery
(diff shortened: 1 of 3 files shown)


aaradhak (approved):
LGTM

dustymabe (approved):
LGTM

PR #1325, opened 2026-03-30
build-node-image: support passing build-args file to builds

jlebon:
Add support for passing a `--build-arg-file` to `cosa remote-build-container` during node image builds. The build-args filename is derived from the `RELEASE` parameter (e.g. build-args-9.8-4.22.conf) and is only passed when the `build_args_file` knob is enabled in pipecfg.

Part of https://redhat.atlassian.net/browse/COS-4051 where we'll add build arg files for setting various security labels on the node image.

Assisted-by: OpenCode (Claude Opus 4.6)


dustymabe (approved):
LGTM

joelcapitao (approved):

PR #1326, opened 2026-03-31
Add configurable GCP zone for kola-gcp job

tlbueno:
Introduce a GCP_ZONE job parameter (default us-central1-b) and pass it to kola as --gcp-zone so tests run in the intended region as kola-gcp job fails due out of resources on the default zone, us-central1-a.


dustymabe (approved):
LGTM

PR #1332, opened 2026-04-11
Revert "strongly couple versionary use to build-with-buildah"

jbtrystram:
This reverts commit 7dc87c44ded20e27231b8dbeed501ac570502aa7. In COSA we recently made `build_qith_buildah` the default, dropping old the dead code[1].
This made `should_build_with_buildah` an undefined function. Since we can't use it anymore to decide whether to use versionnary, let's revert to the previous behaviour whith a pipecfg setting.

[1] https://github.com/coreos/coreos-assembler/pull/4527

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -112,6 +112,7 @@ clouds:
     test_architectures: [x86_64, aarch64]
 
 misc:
+  versionary: true
   generate_release_index: true
   run_extended_upgrade_test_fcos: true
   allow_missing_architectures_when_autotriggering_release_job: true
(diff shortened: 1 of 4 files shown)


dustymabe:
I was just looking at internal build discussion and didn't check my email notifications from GitHub to notice you opened this PR before I opened one too.

Since this has implications downstream (i.e. our `rhel-9.8` builds had the same problem) I think maybe it makes sense to change the coupling here instead of decoupling completely.

I opened https://github.com/coreos/fedora-coreos-pipeline/pull/1333 - see what you think of that?

PR #1334, opened 2026-04-13
tmt: add Testing Farm integration for kola tests

Roshan-R:
Add tmt/fmf metadata and shell scripts to enable running kola tests via Testing Farm. Three test plans are: kola-qemu, kola-test-upgrade, and kola-testiso.

Currently the disk image is built as part of the test run; in the future this will be replaced by importing the image directly from Konflux.


tmt/tests/build.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+set -eo pipefail
+set -x
+
+source "$HOME/utils.sh"
+
+mkdir -p "$COSA_DIR"
+cosa init --force "${TESTING_FARM_GIT_URL}" --branch "${TESTING_FARM_GIT_REF}"
```
joelcapitao: `TESTING_FARM_GIT_URL` and `TESTING_FARM_GIT_REF` refer now to the repo containing the TF/TMT plans i.e: the content of this patch.
For the `cosa init` operation we need to add an additional variable e.g: `CONFIG_REPO` and specifying the value e.g: https://github.com/coreos/fedora-coreos-config in the `IntegrationTest`.
For the branch, the value should be extracted/deducted from the meta.json file e.g: `com.coreos.stream`

joelcapitao (changes requested):

tmt/tests/build.sh
joelcapitao: Actually, we can even get the config URL from the `meta.json` file, with `org.opencontainers.image.source`, which is even better as the OCI is the source of truth.

Roshan-R: nice catch! looks like I got the urls mixed up, I'll sort it out

tmt/tests/init.sh
```diff
@@ -0,0 +1,41 @@
...
+EOF
+
+echo "cosa container: $COREOS_ASSEMBLER_CONTAINER_LATEST"
+echo "arch: $(arch)"
+echo "using image: $IMAGE_URL"
+skopeo inspect "docker://$IMAGE_URL" > meta.json
+GIT_URL=$(jq -r ".Labels.\"org.opencontainers.image.source\"" meta.json)
+GIT_REF=$(jq -r ".Labels.\"com.coreos.stream\"" meta.json)
+echo "export GIT_URL=${GIT_URL}" >> "$HOME/utils.sh"
+echo "export GIT_REF=${GIT_REF}" >> "$HOME/utils.sh"
```
joelcapitao: As you define the constant env var here, then let's be consistent and also add `CONFIG_COMMIT` here.
Also, could we use `CONFIG_GIT_URL` and `CONFIG_GIT_BRANCH_REF` to be more explicit.

tmt/tests/build.sh
```diff
@@ -0,0 +1,20 @@
...
+source "$HOME/utils.sh"
+
+mkdir -p "$COSA_DIR"
+cosa init --force "${GIT_URL}" --branch "${GIT_REF}"
+cosa import "docker://$IMAGE_URL"
+CONFIG_COMMIT=$(jq -r ".\"coreos-assembler.oci-imported-labels\".\"vcs-ref\"" "${COSA_DIR}/builds/latest/$(arch)/meta.json")
+pushd "${COSA_DIR}/src/config"
+git config --global --add safe.directory "${COSA_DIR}/src/config"
+git checkout "$CONFIG_COMMIT"
+popd
```
joelcapitao: I think this block should be moved to `init.sh` as it's pure initialization.
Also, we could use the`--commit` option in `cosa init` e.g: https://github.com/coreos/fedora-coreos-pipeline/blob/74326696c15a2e164516f1505511137a15ec6ed0/jobs/bump-lockfile.Jenkinsfile#L100 
We are also missing the `--variant` flag

joelcapitao (changes requested):

tmt/tests/init.sh
```diff
@@ -0,0 +1,41 @@
...
+    cd $COSA_DIR && tar -C "$OUTPUT_DIR" -c --xz "$KOLA_ID" > "$KOLA_ID-$TOKEN.tar.xz"
+    cd $COSA_DIR && mv "$KOLA_ID-$TOKEN.tar.xz" "$TMT_TEST_DATA/$KOLA_ID-$TOKEN.tar.xz"
+}
+run_kola(){
+    OUTPUT_DIR=$(cd $COSA_DIR && cosa shell -- mktemp -d tmp/kola-XXXX)
+    TOKEN=$(uuidgen | cut -f1 -d -)
+    KOLA_ID=${KOLA_ID:-kola}
+    cd $COSA_DIR && cosa kola "$KOLA_ACTION" --build=latest --arch=$(arch) --output-dir="$OUTPUT_DIR/$KOLA_ID" $KOLA_EXTRA_ARGS
+}
+EOF
```
joelcapitao: Why not defining statically the `utils.sh` file instead of the Heredoc usage ?

tmt/tests/build.sh
Roshan-R: in this case the `--variant` would be `CONFIG_GIT_BRANCH_REF` right?

joelcapitao: Yes, for fedora-coreos-config `branch = variant`, so we should not add it. That's why there is https://github.com/coreos/fedora-coreos-pipeline/blob/74326696c15a2e164516f1505511137a15ec6ed0/jobs/bump-lockfile.Jenkinsfile#L99 because we don't define variant: https://github.com/coreos/fedora-coreos-pipeline/blob/74326696c15a2e164516f1505511137a15ec6ed0/config.yaml#L1-L18

But for rhel-coreos-config, we have `variant` which differ from the branch pior to OCP 4.20. After 4.19-9.6, `branch = variant`.
That said, as we'll use Konflux only for the base OS (ART will take ownership of the node images), that means we don't require `--variant` at all :)

tmt/tests/init.sh
Roshan-R: I added it while the entire repo structure was different, looks like we don't need it now. I have removed the heredoc usage.

tmt/tests/init.sh
Roshan-R: renamed the variables

tmt/tests/init.sh
```diff
@@ -0,0 +1,32 @@
...
+
+echo "git version: $(git --version)"
+echo "git url: ${CONFIG_GIT_URL}"
+echo "git branch: ${CONFIG_GIT_REF}"
+# print commit message
+git clone "${CONFIG_GIT_URL}" /tmp/git-repo
+pushd /tmp/git-repo
+git checkout "${CONFIG_GIT_REF}"
+echo "git commit message: $(git log --format=%B -n 1 HEAD)"
+popd
```
joelcapitao: we can remove this block, because the `cosa init` takes cares of it

tmt/tests/init.sh
```diff
@@ -0,0 +1,32 @@
...
+
+mkdir -p "$COSA_DIR"
+cosa init --force "${CONFIG_GIT_URL}" --branch "${CONFIG_GIT_REF}" --commit "${CONFIG_COMMIT}"
+cosa import "docker://$IMAGE_URL"
+pushd "${COSA_DIR}/src/config"
+git config --global --add safe.directory "${COSA_DIR}/src/config"
+git checkout "$CONFIG_COMMIT"
+popd
```
joelcapitao: this block can be removed because `cosa init` with `--commit` option already checkouts to the right commit and fails if the commit does not exist.

joelcapitao (changes requested):
I just spotted the last changes needed and we are good to merge it

joelcapitao (approved):
Thank you for all the work Roshan, we have a solid base now

PR #1335, opened 2026-04-13
jobs/bump-lockfile: drop custom versionary logic

dustymabe:
When this was introduced I can't remember why it was needed and the git history doesn't help explain it much. Either way, now if you don't pass a version `cosa build` will automatically do the dev version thing so we won't have any risk of collisions especially since we delete the lockfiles prior to calling `cosa build`.


Rolv-Apneseth (approved):
LGTM

PR #1336, opened 2026-04-14
jobs/kola-gcp: Update gcp zone to us-central1-a

aaradhak:
The kola-gcp test seems to fail with gcp zone us-central1-b due to insufficient resources. Changing the zone to us-central1-a seems to address this issue.


dustymabe (approved):
LGTM

PR #1337, opened 2026-04-16
build-node-image: increase resources and timeout

dustymabe:
In f9e562e we switch the build-node-image job to start running actual tests, which means at least for the x86_64 architecture we're running kola (starting virtual machines) in this job.

Let's bump the cpu/memory requirements here and also bump the timeout.


tlbueno:
lgtm

tlbueno (approved):

PR #1339, opened 2026-04-30
jobs: adjust pipeline to build artifacts befora running kola tests

nikita-dubrovskii:
We are going to drop the `testiso` command and consolidate all ISO tests under `kola run iso.* ....`
This PR prepares the pipeline for that change and remains backward compatible with current and older `cosa` releases.

Related:
- https://github.com/coreos/coreos-assembler/issues/3989
- https://github.com/coreos/coreos-assembler/pull/4377

### DIFF
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -313,6 +313,11 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
             shwrap("cosa buildextend-qemu")
         }
 
+        // Build the remaining artifacts
+        stage("Build Artifacts") {
+            pipeutils.build_artifacts(pipecfg, params.STREAM, basearch, skip_untested_artifacts)
+        }
+
         // This is a temporary hack to help debug https://github.com/coreos/fedora-coreos-tracker/issues/1108.
         if (params.KOLA_RUN_SLEEP) {
             echo "Hit KOLA_RUN_SLEEP; going to sleep..."
@@ -328,11 +333,6 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
                  allowUpgradeFail: params.ALLOW_KOLA_UPGRADE_FAILURE,
                  skipSecureBoot: pipecfg.hotfix?.skip_secureboot_tests_hack,
                  skipKolaTags: stream_info.skip_kola_tags)
-        }
-
-        // Build the remaining artifacts
-        stage("Build Artifacts") {
-            pipeutils.build_artifacts(pipecfg, params.STREAM, basearch, skip_untested_artifacts)
         }
 
         // secex specific tests. 
--- a/jobs/build-arch.Jenkinsfile
+++ b/jobs/build-arch.Jenkinsfile
@@ -347,8 +347,10 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
             }
         }
 
-        // Run Kola TestISO tests for metal artifacts
-        if (shwrapCapture("cosa meta --get-value images.live-iso") != "None") {
+        // Run kola testiso tests if supported (legacy). On older releases testiso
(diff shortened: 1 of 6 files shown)


jobs/build.Jenkinsfile
```diff
@@ -387,6 +387,24 @@ lock(resource: "build-${params.STREAM}") {
             shwrap("cosa buildextend-qemu")
         }
```
dustymabe: For newer COSA (that supports OSBuild) it's better (more efficient) if we don't build this separate.

As part of this I think it would be good to just make `qemu` just like any other artifact.

We would need to 

1. add it to the list of artifacts in the upstream and downstream config.yaml
2. update the list in `get_artifacts_to_build()` (utils.groovy) to include qemu in the default list of testable artifacts
3. `update build_artifacts()` (utils.groovy) to (in the non-osbuild case) build the qemu first before it does the parallel runs (since a lot of other artifacts depend on the qemu)

jobs/build-arch.Jenkinsfile
```diff
@@ -348,10 +348,12 @@ lock(resource: "build-${params.STREAM}-${basearch}") {
         }
 
         // Run Kola TestISO tests for metal artifacts
```
dustymabe: Update comment to add more context:

```
// On older releases testiso tests were run separately from
// other kola tests. If we are on one of those releases run
// testiso tests now.
```

dustymabe: also add to the other instance of this.

jobs/build.Jenkinsfile
```diff
@@ -429,19 +429,21 @@ lock(resource: "build-${params.STREAM}") {
...
-                                skipSecureBoot: pipecfg.hotfix?.skip_secureboot_tests_hack)
-                }
-            } else {
-                stage("Kola:TestISO") {
-                    kolaTestIso(cosaDir: env.WORKSPACE, arch: basearch,
-                                skipSecureBoot: pipecfg.hotfix?.skip_secureboot_tests_hack)
+        if (pipeutils.kola_has_testiso()) {
+            if (shwrapCapture("cosa meta --get-value images.live-iso") != "None") {
+                if (pipecfg.hacks?.skip_uefi_tests_on_older_rhcos &&
+                    (params.STREAM in ['4.6', '4.7', '4.8', '4.9'])) {
```
dustymabe: maybe we add a separate commit here to drop this hack since we no longer build for any of those streams

dustymabe:
We probably need to update other consumers of this (like bump-lockfile).

jobs/build.Jenkinsfile
nikita-dubrovskii: will do in separate PR

jobs/bump-lockfile.Jenkinsfile
```diff
@@ -263,8 +263,13 @@ lock(resource: "bump-lockfile") {
                             // OOMkilled.
                             shwrap("cosa shell -- env XZ_DEFAULTS=--memlimit=4G cosa compress --artifact=metal")
                         }
```
dustymabe: move this compression to above `kola` step.

utils.groovy
```diff
@@ -1049,4 +1057,9 @@ def should_we_skip_untested_artifacts(pipecfg) {
     }
 }
 
+// Check if kola has testiso command available
+def kola_has_testiso() {
+    return shwrapRc("cosa kola help | grep -q testiso") == 0
+}
```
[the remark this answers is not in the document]
dustymabe: not worth it.

jobs/bump-lockfile.Jenkinsfile
nikita-dubrovskii: Done

dustymabe (approved):

PR #1340, opened 2026-04-30
docs: document build_args_file knob in config reference

jlebon:
Assisted-by: Pi (Claude Opus 4.6)


c4rt0 (approved):
I traced it down to [build_args_file](https://github.com/coreos/fedora-coreos-pipeline/blob/main/jobs/build-node-image.Jenkinsfile#L68) and at least learned something new about [stream_info](https://github.com/coreos/fedora-coreos-pipeline/blob/main/jobs/build-node-image.Jenkinsfile#L47).
/lgtm

PR #1342, opened 2026-05-01
config: disable next-devel 2026-05

dustymabe:
We've switched `testing-devel` to be based on Fedora 44 so we can disable `next-devel` now since there's no difference in `testing-devel` and `next-devel` now.


marmijo (approved):
LGTM

PR #1348, opened 2026-05-15
kola-upgrade: add --debug so we can get more insight on failures

dustymabe:
We're noticing a lot of the upgrades time out. Let's add debug for now so we can get more information.


c4rt0 (approved):
/lgtm

PR #1354, opened 2026-05-27
kola-upgrade: don't run uefi and uefi-secure tests

dustymabe:
uefi-secure tests uefi as well so no need really to run both tests on x86_64 for starting points that support it.

Let's drop running them both and just do one or the other based on the starting version.

Assisted-by: <anthropic/claude-opus-4.6>


aaradhak (approved):
LGTM

PR #1359, opened 2026-06-15
build-node-image: update build-args-file format

dustymabe:
Now it matches the stream name.
See https://github.com/openshift/os/pull/1941


ravanelli (approved):
/lgtm

PR #1364, opened 2026-07-07
tmt: add build-fcos stage and inline test commands into fmf

Roshan-R:
Add a new `build-fcos` test case to build.sh that runs `cosa build`, producing the OSTree commit and base image artifacts. This sits between the existing `init` and `build-qemu` stages in the pipeline.

The `build-fcos` stage is added because coreos-assembler (and other repos in the future) will reference this repo as the central source of tmt tests. Those repos need an explicit entry point that runs `cosa build` to produce the base artifacts. While this repository does not use the new `build-fcos` stage itself, it is provided for external consumers referencing these tests.

This commit is part of a broader effort to centralize tmt tests in this repo so other repos can reference it.

Also inlined the test command directly into each fmf test definition, removing the need for entrypoint.sh as a central dispatcher.

ref: https://redhat.atlassian.net/browse/COS-4414


Rolv-Apneseth (approved):

joelcapitao:
I don't follow the pattern of the centralization of the stages. If we merge this PR, then the next f-c-c PR will both import the Konflux build _and_ build the container images, right ?
How do we adjust the plans ?

Roshan-R:
@joelcapitao This PR only adds the build-fcos stage for other repos to make use, this stage is not used in the `fedora-coreos-pipeline`. My aim was to make this the repository where all the stages are hosted that includes stages which are not used by `fedora-coreos-pipeline`

joelcapitao:
@Roshan-R let's take the cosa example: how would we make use of the new test `/build-fcos` ? I guess we should create a dedicated FMF plan in `tmt/plans` and pass it as arg in the Konflux ITS, but currently we have `tmt` in `TMT_PLAN` c.f. `Edit parameters` in https://konflux-ui.apps.kflux-fedora-01.84db.p1.openshiftapps.com/ns/coreos-tenant/applications/fedora-coreos-testing-devel/integrationtests/fedora-coreos-testing-devel-kola-tests-x86-64
There is something I'm not following

Roshan-R:
@joelcapitao Since the test definitions will be living here we could reference it like the one in this PR: https://github.com/coreos/coreos-assembler/pull/4593

```/kola-test-upgrade:
  summary: run kola upgrade tests
  discover+:
    - how: fmf
      test:
        - init
        - build-fcos
    - how: fmf
      url: https://github.com/coreos/fedora-coreos-pipeline
      test:
        - build-qemu
        - test-kola-upgrade
```

right now in that PR the `build-fcos` stage is taken from the `coreos-assembler` repo itself. when this PR is merged, we can change the yaml to be 
 
 ```/kola-test-upgrade:
  summary: run kola upgrade tests
  discover+:
    - how: fmf
      test:
        - init
    - how: fmf
      url: https://github.com/coreos/fedora-coreos-pipeline
      test:
        - build-fcos
        - build-qemu
        - test-kola-upgrade
```

**Comment by joelcapitao** [comment]
@Roshan-R okk, the logic lives in the stage itself with the `url` keyword:

```
    - how: fmf
      url: https://github.com/coreos/fedora-coreos-pipeline
```
That's neat, thank you for the explanation.

joelcapitao (approved):

PR #1371, opened 2026-07-29
build.Jenkinsfile: bump pod memory from 10.5G to 12G

c4rt0:
Some tests get OOMKilled during kola testing (LUKS root-reprovision test). Bumping the value up for memory heavy tests.

See: https://github.com/coreos/fedora-coreos-tracker/issues/2200


c4rt0:
Closing this as by @dustymabe :
> the pipeline now does memory aware schedulling so hypothetically it will just schedule more jobs and exhaust memory again if the current accounting is wrong

PR #1372, opened 2026-07-29
builders: add a konflux user

jbtrystram:
Until Konflux is able to provide builders with nested virtualization support, allow konflux to run builds in our builders.

Builds data will live under the konflux $HOME dir and be cleaned up daily, if there are more than one day old (using the `mtime` filter in find).

### DIFF
--- a/multi-arch-builders/builder-common.bu
+++ b/multi-arch-builders/builder-common.bu
@@ -2,11 +2,11 @@
 #
 # - Allow designated users to log in as the core user
 # - Disable kernel mitigations (not a shared instance)
-# - Set up the podman socket for the builder user (podman remote)
+# - Set up the podman socket for the builder and konflux users (podman remote)
 # - Build coreos-assembler on the first boot and once a day
 # - Configure zincati to allow updates at a specific time (early Monday)
-# - Enable linger for builder (so processes can run even if user hasn't
-#   logged in)
+# - Enable linger for builder and konflux (so processes can run even if user
+#   hasn't logged in)
 # - Configure zram
 #
 variant: fcos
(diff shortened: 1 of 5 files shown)


multi-arch-builders/builder-common.bu
```diff
@@ -115,17 +148,93 @@ storage:
...
+        name: konflux
+      group:
+        name: konflux
+      contents:
+        inline: |
+          [Unit]
+          Description=Clean up old build directories
+          [Service]
+          Type=oneshot
+          ExecStart=/usr/bin/find /home/konflux/builds -mindepth 1 -maxdepth 1 -type d -mtime +0 -exec rm -rf {} +
```
dustymabe, line 193: I wonder if systemd-tmpfiles would be better to implement the cleanup here? there should already be a timer that runs periodically for it. `systemd-tmpfiles-clean.timer`

but maybe that timer doesn't run often enough? I guess we could adjust it through configuration

dustymabe:
A few more questions:

1. I don't see anything in here configuring the konflux user with heightened permissions (i.e. `sudo`). So we don't need root access?
2. Do we need to update anything around giving the provisioned machines more disk?

multi-arch-builders/coreos-x86_64-builder.bu
```diff
@@ -16,6 +16,9 @@ passwd:
       ssh_authorized_keys:
         - ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAINmMuXJ9BSNQo8Dh7KMINvRbos/OfXlUMXCStZw72m6p builder@fcos-pipeline-x86_64
         - ssh-ed25519 AAAAC3NzaC1lZDI1NTE5AAAAIFb2urqFBm023Po1axRR6Jphvi30mRupwBrz3WM30ShM builder@rhcos-pipeline-x86_64
+    - name: konflux
+      ssh_authorized_keys:
+        - <KONFLUX_SSH_PUBKEY_X86_64>
```
dustymabe, line 21:
```suggestion
        - ssh-ed25519...
```

Are we planning on putting a real public key here in the final version of this PR?

multi-arch-builders/builder-common.bu
jbtrystram: ah yes, good suggestion !

multi-arch-builders/coreos-x86_64-builder.bu
jbtrystram: > Are we planning on putting a real public key here in the final version of this PR?

yes

jbtrystram:
> 1. I don't see anything in here configuring the konflux user with heightened permissions (i.e. `sudo`). So we don't need root access?

We have a couple of remaining obstacles that I'd like to tackle, and will resort to `sudo` if unable too. But haven't lost hope yet :)

dustymabe:
> But haven't lost hope yet :)

at least there is hope!

PR #1373, opened 2026-07-29
jobs/build-node-image: add skip-build mode for pre-built images

c4rt0:
Accept NODE_IMAGE and EXTENSIONS_IMAGE pullspec parameters. When both are set, the job skips Init, Build Node Image, Build Extensions Image, and Release Manifests stages, running only Run Tests and Brew Upload against the provided images.

This supports the Konflux migration ([COS-4703](https://redhat.atlassian.net/browse/COS-4703)): Konflux builds the node/extensions container images, Jenkins tests and uploads them to Brew.

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/jobs/build-node-image.Jenkinsfile
+++ b/jobs/build-node-image.Jenkinsfile
@@ -24,6 +24,14 @@ properties([
              description: 'Override the coreos-assembler image to use',
              defaultValue: "quay.io/coreos-assembler/coreos-assembler:latest",
              trim: true),
+      string(name: 'NODE_IMAGE',
+             description: 'Pre-built node image pullspec (e.g. quay.io/repo@sha256:abc...). When set, skips build stages.',
+             defaultValue: '',
+             trim: true),
+      string(name: 'EXTENSIONS_IMAGE',
+             description: 'Pre-built extensions image pullspec (e.g. quay.io/repo@sha256:def...). Required when NODE_IMAGE is set.',
+             defaultValue: '',
+             trim: true),
     ]+ pipeutils.add_hotfix_parameters_if_supported()),
     buildDiscarder(logRotator(
         numToKeepStr: '100',
@@ -48,6 +56,23 @@ def stream_info = pipecfg.ocp_node_builds.release[params.RELEASE]
 def skip_brew_upload = stream_info.skip_brew_upload ?: false
 def src_config_ref = stream_info.source_config.ref
 def src_config_url = stream_info.source_config.url
+
+// Null-safe reads: on the very first run after these params are introduced,
+// Jenkins has not registered them yet and `params.*` would be null. In Groovy
+// `null != ''` is true, which would wrongly enable skip-build mode and then
+// NPE on .split(). Same defensive pattern as `params.FROM ?:` below.
+def node_image = params.NODE_IMAGE ?: ''
+def extensions_image = params.EXTENSIONS_IMAGE ?: ''
+def skip_build = false
+if (node_image != '' || extensions_image != '') {
+    skip_build = true
+    if (node_image == '' || extensions_image == '') {
+        error("NODE_IMAGE and EXTENSIONS_IMAGE must both be set if either is set.")
+    }
+    if (!node_image.contains('@sha256:') || !extensions_image.contains('@sha256:')) {
+        error("NODE_IMAGE and EXTENSIONS_IMAGE must be digest pullspecs (repo@sha256:...)")
+    }
+}
 
 def basearches = params.ARCHES.split() as Set
 def timeout_mins = 300
@@ -100,6 +125,7 @@ lock(resource: "build-node-image") {
         def registry_prod_tags
         def node_image_manifest_digest
         def extensions_image_manifest_digest
+        def image_repo
 
         try {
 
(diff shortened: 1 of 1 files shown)


jobs/build-node-image.Jenkinsfile
```diff
@@ -49,6 +57,25 @@ def skip_brew_upload = stream_info.skip_brew_upload ?: false
...
+}
+if (node_image == '' && extensions_image != '') {
+    error("EXTENSIONS_IMAGE is set but NODE_IMAGE is empty. Both are required.")
+}
+// brew_upload and the test stage need digests, not tags. Fail fast with a
+// clear message instead of an ArrayIndexOutOfBounds deep inside the run.
+if (skip_build && (!node_image.contains('@sha256:') || !extensions_image.contains('@sha256:'))) {
+    error("NODE_IMAGE and EXTENSIONS_IMAGE must be digest pullspecs (repo@sha256:...)")
+}
+
```
dustymabe, line 76: There's a lot of conditional logic here and maybe we can simplify..


```
def node_image = params.NODE_IMAGE ?: ''
def extensions_image = params.EXTENSIONS_IMAGE ?: ''
if (node_image != '' || extensions_image != '') {
    skip_build = true
    if (node_image == '' || extensions_image != '') {
            error("NODE_IMAGE and EXTENSIONS_IMAGE need to be set if either are set")
    }
    // brew_upload and the test stage need digests, not tags. Fail fast 
    if (!node_image.contains('@sha256:') || !extensions_image.contains('@sha256:')) {
         error("NODE_IMAGE and EXTENSIONS_IMAGE must be digest pullspecs (repo@sha256:...)")
    }   
}
```

jobs/build-node-image.Jenkinsfile
```diff
@@ -142,7 +169,7 @@ lock(resource: "build-node-image") {
         pipeutils.addOptionalRootCA()
 
         def yumrepos_file = ""
-        if (stream_info.yumrepo?.url) {
+        if (!skip_build && stream_info.yumrepo?.url) {
```
dustymabe: I wonder here if it would be better to move the tests and brew into functions that we all and then:
```
   if (skip_build) {
        do_tests
        do_brew
   } else {
       do it all 
       update existing brew and tests to call do_tests() and do_brew() too.
   } // end if (skip_build)

**Comment by c4rt0** [comment]
@dustymabe Thanks for the suggestions. All implemented. As most of it was rearrengment I decided to also create a refactoring guide to make it easier to digest.

### Validation (lines 64-74)

Nested all checks inside one `if (node_image != '' || extensions_image != '')` block.
Small note: the inner check in the review snippet had `extensions_image != ''`, I went with `== ''` so it catches the "only one is set" case.

```groovy
def skip_build = false
if (node_image != '' || extensions_image != '') {
    skip_build = true
    if (node_image == '' || extensions_image == '') {
        error("NODE_IMAGE and EXTENSIONS_IMAGE must both be set if either is set.")
    }
    if (!node_image.contains('@sha256:') || !extensions_image.contains('@sha256:')) {
        error("NODE_IMAGE and EXTENSIONS_IMAGE must be digest pullspecs (repo@sha256:...)")
    }
}
```

### Pipeline structure (lines 175-387)

Extracted tests and brew into `do_tests`/`do_brew` closures, single
`if/else` controls the whole flow. No scattered `!skip_build` guards.

| Block | Before | After |
|---|---|---|
| Run Tests | inline, unguarded | `do_tests` closure, called from both branches |
| Brew Upload | inline, unguarded | `do_brew` closure, called from both branches |
| Init stage | guarded by `!skip_build &&` | inside `else` block (guard removed) |
| Build Node Image | wrapped in `if (!skip_build)` | inside `else` block (guard removed) |
| Build Extensions | guarded by `!skip_build &&` | inside `else` block (guard removed) |
| Release Manifests | wrapped in `if (!skip_build)` | inside `else` block (guard removed) |
| Pullspec parsing | guarded by `if (skip_build)` | inside `if` block |
| `image_repo` | declared + set inline | declared at top, set inside each branch |

```
def do_tests = { ... }     ← test block, lifted verbatim
def do_brew  = { ... }     ← brew block, lifted verbatim

if (skip_build) {
    parse pullspecs → set image_repo, digests
    do_tests()
    do_brew()
} else {
    Init
    Build Node Image
    Build Extensions Image
    image_repo = registry_staging_repo
    do_tests()
    do_brew()
    Release Manifests
}
```

jobs/build-node-image.Jenkinsfile
c4rt0: Thank you, commented: https://github.com/coreos/fedora-coreos-pipeline/pull/1373#issuecomment-5142011193

jobs/build-node-image.Jenkinsfile
c4rt0: Implemented: https://github.com/coreos/fedora-coreos-pipeline/pull/1373#issuecomment-5142011193

dustymabe:
Thanks @c4rt0.

I think it would read better if we were able to move the do_tests() and do_brew() parts to being defined later in the file.

Look at build.Jenkinsfile where we do something similar:

https://github.com/coreos/fedora-coreos-pipeline/blob/e4dca956869b586bbd12a1e7c7cd2e85ad2107ae/jobs/build.Jenkinsfile#L551-L572

The main block ends and then we define some functions that are used in the code above.

I'm not sure if using a closure would make this more complicated than if we were using functions or not. It might.

WDYT?

c4rt0:
> Thanks @c4rt0.
> 
> I think it would read better if we were able to move the do_tests() and do_brew() parts to being defined later in the file.
...
> WDYT?

Honestly, I looked into this, and the do_tests was supposed to be something like: do_tests(image_repo, digest, basearches , cosa_img, pipecfg, release...). It needed several parameters since it uses variables from inside the try block. Following that I had to call it filing up all of these parameters, and thought that's just asking myself for errors and confusion. I'm happy to switch if you so prefer, personally I'd rather leave it as is.

dustymabe:
> do_tests(image_repo, digest, basearches , cosa_img, pipecfg, release...). It needed several parameters since it uses variables from inside the try block.

yep that's fine to leave them as closures

> I'm happy to switch if you so prefer, personally I'd rather leave it as is.

do you know if leaving them as a closures means we can't move them to the bottom of the file? For me that was the bigger issue (i.e. the flow of reading the code is interrupted)

If we can leave them as a closure and move them to the bottom, let's do that. If not then let's leave them as is. Does that seem reasonable?

c4rt0:
> do you know if leaving them as a closures means we can't move them to the bottom of the file? For me that was the bigger issue (i.e. the flow of reading the code is interrupted)
> 
> If we can leave them as a closure and move them to the bottom, let's do that. If not then let's leave them as is. Does that seem reasonable?

I am genuinely greatful for this type of questions... 
I had to do some extra digging to make sure I know what I'm saying and it turns out that the answer is no. Acc to the groovy docs, closures "may reference variables declared in its surrounding scope" (the block they're written inside). If we move them outside our "try" block and to the bottom of the file, they lose access to the mentioned parameters. So no, can't move them down and keep them as closures.

dustymabe:
> If we move them outside our "try" block and to the bottom of the file, they lose access to the mentioned parameters. So no, can't move them down and keep them as closures.

ahh. I suspected that was the case

jobs/build-node-image.Jenkinsfile
```diff
@@ -142,194 +168,223 @@ lock(resource: "build-node-image") {
...
-                // copy the extensions first as the node image existing is a signal
-                // that it's ready for release. So we want all the expected artifacts
-                // to be available when the ART tooling kicks in.
-
-                // Skopeo does not support pushing multiple tags at the same time
-                // So we just recopy the same image multiple times.
-                // https://github.com/containers/skopeo/issues/513
-                if (stream_info.extensions != false) {
+
+            image_repo = registry_staging_repo
```
dustymabe, line 363:
```suggestion
            // Use the staging repo since we already have the digests
            image_repo = registry_staging_repo
```

Let's carry this comment forward from the old code to here.

dustymabe (approved):
It would be good to give this a run or two in the devel pipeline just to confirm we don't have any issues prior to merging.

Otherwise this LGTM

Thank you!

jobs/build-node-image.Jenkinsfile
c4rt0: Addressed.

c4rt0:
@dustymabe 

Tested this on the RHCOS devel pipeline. Node image (NO451) built successfully for 4.19-9.6 on all 4 arches. 
Push failed on s390x/ppc64le with registry auth (devel pipeline credential issue). NODE_IMAGE and EXTENSIONS_IMAGE now appear in Build with Parameters 🎉

First 4.22-10.2 build failed due to THE COS-4741 (`build_args_file: true` missing from devel pipecfg). After updating config.yaml, build NO455 for 4.22-10.2 succeeded on all 4 arches.

dustymabe:
That's great. Merge on Monday?

c4rt0:
> That's great. Merge on Monday? 

Sure, just needed to make sure I'm not missing anything before the merge. Thanks for all the help!

PR #1376, opened 2026-08-05
jobs/build-mechanical: Increase cron frequency to every 12h

aaradhak:
Run the mechanical build every 12 hours (10:00 and 22:00 UTC) instead of once daily to catch issues sooner.

### DIFF
--- a/jobs/build-mechanical.Jenkinsfile
+++ b/jobs/build-mechanical.Jenkinsfile
@@ -7,8 +7,8 @@ node {
 
 properties([
     pipelineTriggers([
-        // run every 24h at 10:00 UTC
-        cron("0 10 * * *")
+        // run every 12h at 10:00 and 22:00 UTC
+        cron("0 10,22 * * *")
     ]),
     buildDiscarder(logRotator(
         numToKeepStr: '100',


angelcerveraroldan (approved):
LGTM, thanks!

dustymabe:
any context here?

if a build fails we can always start a new one. I'm not sure we should increase the rate of our default schedule here.

aaradhak:
@dustymabe This came up from slack [thread](https://redhat-internal.slack.com/archives/C04668B0XK2/p1785936681070629?thread_ts=1785898444.458279&cid=C04668B0XK2) about build-node-image failing for 4.23-10.2. NetworkManager rpm shipped mid-day and the lockfile still had the old version pinned which broke NetworkManager-ovs deps and blocked CI in openshift/os too.

Since relying on manual reruns has the possibility that we may not notice content has shipped until downstream builds start failing, running it every 12h could help to shrink that window.

dustymabe:
@aaradhak thanks for the context.

There are a few changes downstream to try to make the builds more reliable in cases where the base content is out of date with the repos:

- https://github.com/openshift/os/pull/1958
- https://github.com/coreos/rhel-coreos-config/pull/296

I think we should skip merging this for now and revisit later if we are still having trouble.

aaradhak:
sounds good. we ll skip making these changes for now - closing this PR without merging the changes.

PR #1380, opened 2026-08-19
jobs/kola-upgrade: increase build history retention to 200

aaradhak:
`kola-upgrade` runs per-stream per-architecture so builds roll off quickly. Increase retention so results are still available during release verification.


Nemric (approved):

marmijo (approved):
LGTM

PR #1384, opened 2026-08-25
config: enable branched stream 2026-08

dustymabe:
Fedora 45 branched from rawhide. Enable the branched stream to build Fedora 45 CoreOS.

See https://github.com/coreos/fedora-coreos-tracker/issues/2127


aaradhak (approved):
LGTM

PR #1385, opened 2026-08-27
utils: signImages(): stop specifying gpgkeypath

dustymabe:
Just let the default get used. The default was changed in [1] and let's just use that in case it ever changes again.

[1] https://github.com/coreos/coreos-assembler/commit/2a26517b5e53d82f8fe11b537bd497829e72306f


Rolv-Apneseth (approved):
LGTM

PR #1387, opened 2026-09-08
jobs/build: Support importing Konflux artifacts

jbtrystram:
Leverage the new `--download` flag in `cosa import` [1] to pull artifacts that may have been already built by the Konflux pipeline.

This will allow us to do a smooth transition as we onboard more artifacts in the Konflux pipeline.
Any already existing artifacts won't be rebuilt and the pipeline will jump straight to testing.

[1] https://github.com/coreos/coreos-assembler/commit/b2d4323668cb8d5f60b729745dbdba38092261cd.

Assisted-by: Opencode.ai <Opus 4.6>


jbtrystram:
Opening this as a draft until I do some manual testing

PR #1388, opened 2026-09-08
jobs/build-node-image: sync registry pull secret to remotes

dustymabe:
Since we're not using "podman remote" directly here, but using `skopeo` instead, we need to sync across the pull secret prior to running `skopeo` that accesses the private registry.


dustymabe:
moving to draft until this gets tested

ravanelli (approved):
/lgtm CI worked

PR #1390, opened 2026-09-14
Add priority to build node image lock queue

ravanelli:
- In some cases, we need to run a build sooner instead of waiting for other builds already queued Without a way to prioritize the queue, we may need to abort other builds or wait for them to finish just to run the build we need.

 - Add a PRIORITY parameter and pass it to the lockable resources plugin, allowing higher-priority builds to move ahead in the lock queue.

See more:
https://github.com/jenkinsci/lockable-resources-plugin#lock-queue-priority


jobs/build-node-image.Jenkinsfile
```diff
@@ -32,6 +32,10 @@ properties([
              description: 'Pre-built extensions image pullspec (e.g. quay.io/repo@sha256:def...). Required when NODE_IMAGE is set.',
              defaultValue: '',
              trim: true),
+      string(name: 'PRIORITY',
+             description: 'Lock queue priority see: https://github.com/jenkinsci/lockable-resources-plugin#lock-queue-priority',
+             defaultValue: '0',
+             trim: true),
```
dustymabe, line 38: I was thinking instead of this below we would just derive a value for priority based on if the `NODE_IMAGE` parameter was set.

i.e. any "import + testing" runs get higher priority.

That being said, explicitly setting the priority seems OK too, but may require someone (the caller) to know how the priority system works in jenkins (i.e. is a smaller number or bigger number higher priority?)

dustymabe (approved):
LGTM if your testing looks good.

Left one comment on strategy in case you want to pivot to that, otherwise this will work just fine.

jobs/build-node-image.Jenkinsfile
ravanelli: 0 is the default, so all builds will get normally in the queue with that. Yes, higher number higher priority. Note I added the link explaining it in the field. 

For ART I think we set it higher as default as well.

ravanelli: Add 100 as priority for ART runs

dustymabe (approved):

dustymabe:
LGTM

